PR #36, opened 2022-12-16
feat: add field my_issue props to store display values of my issues page

pablohashescobar:
feat: add field my_issue props to store display values of my issues page


pablohashescobar:
Closing this due to stale commits

PR #39, opened 2022-12-16
feat: create a dedicated endpoint to view to the logged in user's project attributes

pablohashescobar:
feat: create a dedicated endpoint to view to the logged in user's project attributes


pablohashescobar:
- Reraising it to master

PR #53, opened 2022-12-20
fix: my project member viewset

pablohashescobar:
fix: my project member viewset


pablohashescobar:
re-releasing it to stage-release

PR #76, opened 2022-12-30
build: single docker file to run all the processes

pablohashescobar:
build: single docker file to run all the processes


pablohashescobar:
# Closing and reraising it to stage-release

PR #159, opened 2023-01-11
feat: return total member count on workspace member list

pablohashescobar:
feat: return total member count on workspace member list


pablohashescobar:
Closing and reraising

PR #160, opened 2023-01-11
fix: set dockerized value on staging as magic links are failing

pablohashescobar:
fix: set dockerized value on staging as magic links are failing


pablohashescobar:
Re raising

PR #238, opened 2023-02-03
fix: github auth login

venplane:
- Parsed private emails by making request to github resource


pablohashescobar (approved):

sriramveeraghanta (approved):

apiserver/plane/api/views/oauth.py
```diff
@@ -108,6 +107,12 @@ def get_user_data(access_token: str) -> dict:
 
     userData = resp.json()
```
pablohashescobar, line 108: @venplane can you please rename the variable to `user_data`

apiserver/plane/api/views/oauth.py
```diff
@@ -108,6 +107,12 @@ def get_user_data(access_token: str) -> dict:
...
 
+    req = requests.get(url="https://api.github.com/user/emails", headers=headers)
+    resp = req.json()
```
pablohashescobar, line 111: please rename variable to `response` and add `.json()` after `)`  which will help omit line 111

apiserver/plane/api/views/oauth.py
```diff
@@ -108,6 +107,12 @@ def get_user_data(access_token: str) -> dict:
 
     userData = resp.json()
 
+    req = requests.get(url="https://api.github.com/user/emails", headers=headers)
```
pablohashescobar, line 110: also can we use `/user` endpoint for getting this info [Get the authenticated user
](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user)

PR #291, opened 2023-02-16
feat: resend login magic code 

dakshesh14:
feat:

- users can request a new magic login code if they need one
- handling limit exhausted error on magic link login

### DIFF
--- a/apps/app/components/account/email-code-form.tsx
+++ b/apps/app/components/account/email-code-form.tsx
@@ -21,6 +21,7 @@ export const EmailCodeForm = ({ onSuccess }: any) => {
   const [codeSent, setCodeSent] = useState(false);
   const [codeResent, setCodeResent] = useState(false);
   const [isCodeResending, setIsCodeResending] = useState(false);
+  const [errorResendingCode, setErrorResendingCode] = useState(false);
   const [resendCodeTimer, setResendCodeTimer] = useState(RESEND_CODE_TIMER);
 
   const { setToastAlert } = useToast();
@@ -42,6 +43,7 @@ export const EmailCodeForm = ({ onSuccess }: any) => {
   });
 
   const onSubmit = async ({ email }: EmailCodeFormValues) => {
+    setErrorResendingCode(false);
     await authenticationService
       .emailCode({ email })
       .then((res) => {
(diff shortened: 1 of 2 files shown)


apps/app/components/account/email-code-form.tsx
```diff
@@ -66,18 +80,36 @@ export const EmailCodeForm = ({ onSuccess }: any) => {
       });
   };
 
+  useEffect(() => {
```
sriramveeraghanta, line 83: Reason behind this UseEffect

apps/app/components/account/email-code-form.tsx
```diff
@@ -66,18 +80,36 @@ export const EmailCodeForm = ({ onSuccess }: any) => {
...
+      timer = setTimeout(() => {
+        setResendCodeTimer((prev) => prev - 1);
+      }, 1000);
+    }
+    return () => clearTimeout(timer);
+  }, [codeSent, resendCodeTimer]);
+
+  const emailOld = getValues("email");
+
+  useEffect(() => {
```
sriramveeraghanta, line 95: Explain this.

apps/app/components/account/email-code-form.tsx
```diff
@@ -114,15 +146,35 @@ export const EmailCodeForm = ({ onSuccess }: any) => {
...
-                handleSubmit(onSubmit);
+                setIsCodeResending(true);
+                onSubmit({ email: getValues("email") }).then(() => {
+                  setCodeResent(true);
+                  setIsCodeResending(false);
+                  setResendCodeTimer(RESEND_CODE_TIMER);
+                });
               }}
+              disabled={
+                resendCodeTimer > 0 || isCodeResending || isSubmitting || errorResendingCode
```
sriramveeraghanta, line 163: Create a new variable isDisabled and handle it in html

apps/app/components/account/email-code-form.tsx
dakshesh14: At line 83, the useEffect function is implemented to create a 30-second timer. This timer ensures that if the user requires a new login code, they must wait for 30 seconds before being able to request a new code.

apps/app/components/account/email-code-form.tsx
dakshesh14: The code block found at lines 93-97 is intended to allow users to change their email address in case they have exceeded the maximum number of attempts to receive a login code for a particular email. This feature enables users to obtain a code for a different email address.

apps/app/components/account/email-code-form.tsx
sriramveeraghanta: Wrap this logic inside a hook.

sriramveeraghanta (approved):

PR #333, opened 2023-02-23
fix: create issue modal close on escape click

dakshesh14:
fix: 

- create issue modal close on escape click

### DIFF
--- a/apps/app/components/issues/modal.tsx
+++ b/apps/app/components/issues/modal.tsx
@@ -75,6 +75,19 @@ export const CreateUpdateIssueModal: React.FC<IssuesModalProps> = ({
       setActiveProject(projects?.find((p) => p.id === projectId)?.id ?? projects?.[0].id ?? null);
   }, [projectId, projects]);
 
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


sriramveeraghanta (approved):

slasktrat:
If you have the create issue modal open and are using the "Select Parent Issue" search, clicking escape should only abort the parent issue search, but is currently also closing the create issue modal. I think it also should be considered to add a prompt (if any text is added) before canceling the issue on escape click

aaryan610:
Hi @slasktrat, thank you for reporting this issue. We will work on fixing this problem and aim to include the solution in the upcoming release.

In the meantime, if you could create a separate issue for the same matter, it would be very helpful for us to keep track of it.

slasktrat:
> Hi @slasktrat, thank you for reporting this issue. We will work on fixing this problem and aim to include the solution in the upcoming release.
> 
> In the meantime, if you could create a separate issue for the same matter, it would be very helpful for us to keep track of it.

Created separate issue: https://github.com/makeplane/plane/issues/1786

PR #340, opened 2023-02-24
fix: remove project slug

pablohashescobar:
fix: remove project slug


vamsi:
Why are we dropping the Slug field?

vamsi (approved):

PR #429, opened 2023-03-13
fix: modules sidebar lead select not working

aaryan610:
fix:

- lead select not working on the modules sidebar.


dakshesh14 (approved):
LGTM

PR #424, opened 2023-03-12
Github actions to push images

pinarruiz:
Add Github Actions that will build and publish Frontend and Backend images of plane to Github's registry.

Things to note:
1. This is my first time with Github Actions, and wanted to give it a shot, so any feedback/commit is welcomed.
2. ARM jobs take longer, specially Backend, specially because the [Gevent](https://github.com/gevent/gevent) dependency does not have any wheels for arm64, and as it builds it from source on the spot, increases build time notoriusly.
3. The auto generated labes/tags from the `docker/metadata-action` job might not be accurate or adequate.

PR #445, opened 2023-03-15
fix: Add tags for frontend build action

pinarruiz:
Add tags for frontend action build (https://github.com/makeplane/plane/issues/424#issuecomment-1465601263).

### DIFF
--- a/.github/workflows/push-image-frontend.yml
+++ b/.github/workflows/push-image-frontend.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - 'develop'
       - 'master'
+    tags:
+      - '*'
 
 jobs:
   build_push_frontend:


.github/workflows/push-image-frontend.yml
```diff
@@ -5,6 +5,8 @@ on:
     branches:
       - 'develop'
       - 'master'
+    tags:
```
satyanarayanav, line 8: @pinarruiz can you please use tags-ignore instead of a wildcard here?

satyanarayanav (changes requested):
.

.github/workflows/push-image-frontend.yml
narayanavadapalli, line 8: @pinarruiz can you please use tags-ignore here instead of a wildcard ?

narayanavadapalli (changes requested):

.github/workflows/push-image-frontend.yml
narayanavadapalli: The asterisk is a special character in YAML. It is used in conjunction with &, which defines an anchor, that can then be referenced it with an asterisk (*) .  Therefore we can enclose it in single quotes.  Also I was thinking that every commit to develop need not get built. So we can pass a tag-ignore that ignores the workflow build on a specific tag. like "build : ignore" or something around that lines

pinarruiz: I dont think i fully understand, should not that be accomplished by adding a certain string to the commit message? Like: 

- `[skip ci]`
- `[ci skip]`
- `[no ci]`
- `[skip actions]`
- `[actions skip]`

As stated by: [skipping-workflow-runs](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs)

pablohashescobar (approved):

narayanavadapalli (approved):

PR #478, opened 2023-03-18
style: auth screens 

anmolsinghbhatia:
style:
- design sign-in page
- design not a member and not authorize for project setting


apps/app/components/account/email-signin-form.tsx
dakshesh14, line 1: Please remove comments, if they don't add any value.

dakshesh14 (approved):
Looks good to me

aaryan610 (approved):

PR #489, opened 2023-03-22
chore: new modules and cycles response

aaryan610:
chore:

- updated modules and cycles issue stats calculation from frontend to backend.


dakshesh14 (approved):
Looks good to me

PR #501, opened 2023-03-23
chore: new global select filters component

aaryan610:
chore:

- created a new global select filters component for the create view modal and issue filters.


dakshesh14 (approved):
Looks good to me

PR #542, opened 2023-03-27
feat: jitsu tracker setup

dakshesh14:
feat: 

- jitsu tracker setup
- tracking create, update and delete project
- tracking create, update and delete workspace


apps/app/services/track-event.service.ts
aaryan610, line 1: Why do we need these many functions? Can't we create one global trackWorkspace(eventName, data) which will track all the workspace events and one trackProject(eventName, data) which will track all the project events?

aaryan610 (approved):

PR #571, opened 2023-03-29
style: added divider to the pages block

kunalv17:
style:

- added divider to the pages block

### DIFF
--- a/apps/app/components/pages/single-page-block.tsx
+++ b/apps/app/components/pages/single-page-block.tsx
@@ -36,7 +36,6 @@ import { PAGE_BLOCKS_LIST } from "constants/fetch-keys";
 
 type Props = {
   block: IPageBlock;
-  index: number;
   projectDetails: IProject | undefined;
 };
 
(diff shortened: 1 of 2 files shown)


apps/app/components/pages/single-page-block.tsx
```diff
@@ -36,6 +36,7 @@ import { PAGE_BLOCKS_LIST } from "constants/fetch-keys";
 
 type Props = {
   block: IPageBlock;
+  index: number;
```
aaryan610, line 39: I don't see the need for this index prop, can we please remove it?

kunalv17: sure

apps/app/pages/[workspaceSlug]/projects/[projectId]/pages/[pageId].tsx
```diff
@@ -406,13 +404,16 @@ const SinglePage: NextPage<UserAuth> = (props) => {
...
-                        key={block.id}
-                        block={block}
-                        projectDetails={projectDetails}
-                      />
+                  <div className="space-y-4 divide-y">
+                    {pageBlocks.map((block, index) => (
+                      <>
+                        <SinglePageBlock
+                          key={block.id}
+                          index={index}
```
Comment (aaryan610) [inline apps/app/pages/[workspaceSlug]/projects/[projectId]/pages/[pageId].tsx:412]: Why is index being passed here?

PR #577, opened 2023-03-29
feat: cycle list page

anmolsinghbhatia:
feat: 
- in cycle list page message  added when no current cycle is available .


apps/app/pages/[workspaceSlug]/projects/[projectId]/cycles/index.tsx
```diff
@@ -124,7 +126,7 @@ const ProjectCycles: NextPage<UserAuth> = (props) => {
           </div>
         </div>
         <div className="flex flex-col gap-5">
-          <h3 className="text-3xl font-semibold text-black">Others</h3>
+          <h3 className="text-3xl font-semibold text-black">Others Cycles</h3>
```
Comment (aaryan610) [inline apps/app/pages/[workspaceSlug]/projects/[projectId]/cycles/index.tsx:129]: The text should be Other Cycles

aaryan610 (approved):

PR #706, opened 2023-04-04
docs: update readme to include default email and password for self hosting setup

pablohashescobar:
docs: update readme to include default email and password for self hosting setup


ngetchell:
I just did the self-hosting route and these default credentials do not work.

PR #833, opened 2023-04-15
Add a few more options for self-hosting

kylewlacy:
This PR adds a few more settings to address some papercuts I ran into while I was setting up a self-hosted instance of Plane.

Just to go over the changes:

- Added `$EMAIL_USE_TLS` to optionally disable TLS when sending emails via SMTP. I added this because I have been using [MailHog](https://github.com/mailhog/MailHog) to test email delivery locally (which doesn't support TLS as far as I cant tell).
- Added `$EMAIL_PORT` change the port used when sending emails via SMTP. Port 587 still seems like the right default, but my understanding is that there's not even consensus that this is the "one true" SMTP port. Again, I added this because of MailHog (which uses port 1025 by default).
- Added `$EMAIL_FROM` to set the sender name/address. I needed this so I could switch over to sending emails through a real SMTP server (AWS SES in my case).
- ~~Added `$REDIS_BROKER_SSL` to optionally disable the extra TLS settings that are added to the Celery Redis connection string.~~ I added this in the first version of the PR, but removed it since a fix for this issue already landed on `develop`.
- Added `$AWS_S3_ENDPOINT_URL` to set the setting of the same name. I used this to use SeaweedFS for file uploads while self-hosting.
- Added `$NEXT_PUBLIC_EXTRA_IMAGE_DOMAINS`. This option is used  to add more domains (comma-separated) to the `next/images` config. Without this setting, I wasn't able to actually use the images that were being uploaded to S3 (avatar URLs, for example).


vihar:
Thanks for your contribution @kylewlacy, this is great. We will review this shortly and get back to you.

kylewlacy:
I pushed a few extra changes to resolve merge conflicts, to add support for setting `AWS_S3_ENDPOINT_URL` using an env var, and to remove the `$REDIS_BROKER_SSL` setting/env-var (since this is no longer necessary after some changes on `develop`).

kylewlacy:
Alright, just added one more change, which I think this is the final one I needed for my self-hosting setup. The new option is the `$NEXT_PUBLIC_EXTRA_IMAGE_DOMAINS` env var, which is a comma-separated list of domains to add for the `next/images` list of allowed domains.

pablohashescobar (approved):

vamsi (commented):
Looks good to me.

vamsi (approved):

PR #960, opened 2023-04-25
fix: sub issues list fix

anmolsinghbhatia:
fix: 
- sub issues list fix


apps/app/components/issues/sub-issues-list.tsx
```diff
@@ -86,13 +70,21 @@ export const SubIssuesList: FC<Props> = ({ parentIssue }) => {
           (prevData) => {
             if (!prevData) return prevData;
             let newSubIssues = prevData.sub_issues as IIssue[];
```
aaryan610, line 72: What is the use of this array?

aaryan610 (approved):

PR #965, opened 2023-04-28
chore: due date filter

kunalv17:
chore: 

- added due date filter.

fix:

- all issues view overflow when filters are applied.

refactor:

- core components folder structure.


apps/app/styles/globals.css
```diff
@@ -197,22 +215,105 @@ body {
...
-  );
+.react-datepicker-wrapper {
+  position: relative;
+  background-color: rgba(var(--color-bg-base)) !important;
 }
 
-.progress-bar {
-  fill: currentColor;
-  color: rgba(var(--color-bg-sidebar));
+.react-datepicker {
```
aaryan610, line 223: Create a new file for date picker styles,

apps/app/components/ui/multi-level-dropdown.tsx
```diff
@@ -1,10 +1,18 @@
 import { Fragment, useState } from "react";
```
aaryan610, line 1: Don't change the global dropdown component, if required, create a new component.

gurusainath (approved):

Issue #1019, opened 2023-05-06
[feature]: simplify the docker compose setup


Robin-Sch:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

I think the docker compose file could be improved/simplified.

### Why should this be worked on?

Right now the compose file is a mess, and to selfhost plane you need to clone the whole repo instead of just using one compose file.

### What could be changed/improved?

~~What bothers me the most is that I need to pull the whole repo instead of that everything already was included in the docker image.~~
I've taken a quick look and it seems like this can be easily fixed by removing the build from the api container (someone probably forgot to remove that). After removing that I only needed the docker compose file (so maybe change the readme to only download the compose file, instead of the whole project?)

Besides that, is it possible to use my already existing proxy, which is hosting other services? Instead of just running a nginx container and expecting nothing else to be hosted on the server (and using port 80/443).

Also, why not store all the environment variables in the .env file and use the [env file attribute](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute) instead of the `source .env`. 

The ports for redis and postgres don't have to be exposed (maybe you want this for the development environment, but for selfhosting it's better to remove it/disable it), as the containers can communicate freely in the docker network with each other. This way you can't connect to redis/postgres from the outside.

Web-eWorks:
> Besides that, is it possible to use my already existing proxy, which is hosting other services? Instead of just running a nginx container and expecting nothing else to be hosted on the server (and using port 80/443).

As part of self-hosting setup for v0.5, I've removed all external port bindings from the docker-compose file (as all services talk to each other via the docker network created for Plane), and exposed the nginx container on an "internal" port (e.g. 8000 with iptables rules to block all external traffic) for which all traffic is proxied to by the top-level nginx instance which handles the domain name in question. The only thing you have to do is change all public URLs to point to the specific domain or subdomain hosting the Plane instance so that all traffic between the frontend and backend is correctly routed according to your top-level nginx configuration.

This has changed *slightly* with v0.6-develop but shouldn't be significantly more difficult to implement.

The main advantage of this is that you don't have to "import" the apiservice proxy setup into your top-level nginx config and keep it up to date when Plane changes

gokuu:
+1 to this. 

I'm using traefik as the proxy container, and am still struggling to get this working.

Robin-Sch:
[...]

PR #1022, opened 2023-05-06
Clean up docker compose file for selfhosting

Robin-Sch:
### Removed nginx container (might want to put it back?)
I think the majority of people who will selfhost plane are selfhosting more than just plane, and this way the nginx container doesn't conflict with their current setup (port 80 is most likely already in use).
Maybe show/guide on how to setup a simple nginx server (in case they haven't done so)?

### Changed the order, first the important stuff (plane) and later the database/redis
For example, to upgrade to a different version or change port(s) you don't have to scroll all the way down anymore
Also I don't think anyone will ever change something with the redis/postgres instance?

### All containers are in the same format
First container_name, then image, then restart etc..
Just standardization, and looks better

### Removed links
Links are deprecated since compose version 2, all containers are in one docker network

### Removed build from plane-api
Now you don't have to download the whole repo anymore, just for the api image

### Removed ports from redis and postgresql
They can communicate freely with each other without exposing the port, increases the security since the postgres and redis instances aren't accessible from the outside

### Removed PGDATA environment variable
It's the default one and by removing it's less clutter in the file

### Renamed redis and db to plane-redis and plane-db
This way you know the redis and postgres containers belong to plane and aren't just named "redis" and "db"

### Could still be improved
The environment variables could be included in the `.env` file to remove them from the compose file and making the compose file more readable and smaller [see this](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute)
This way you can change all environment variables from the `.env` file and they aren't spread across two files.

https://github.com/makeplane/plane/issues/1019


Igorgro:
> Changed spacing to tabs for better readability

Use tabs instead of spaces is a bad idea for YAML files

vihar:
Thank you, folks, for your contributions. We are still testing different options, and this temporary unstable fix is just one of them. Our team is actively working on finding a permanent solution. We appreciate your contribution and will review it and merge it soon.

Web-eWorks:
> Removed ports from redis and postgresql
> They can communicate freely with each other without exposing the port, increases the security since the postgres and redis instances aren't accessible from the outside

I'd like to second this change - because the nginx instance and all of Plane's services are on the same Docker network, binding host ports is not required for communication between services. Exposing these services via bound host ports represents a security issue in single-server production environments; only the nginx proxy port needs to be accessible outside of the docker network for upstream traffic.

Robin-Sch:
> Use tabs instead of spaces is a bad idea for YAML files

My bad, I changed it to 4 spaces after I noticed tabs not working (that's why the multiple fix spacing commits)

> I'd like to second this change - because the nginx instance and all of Plane's services are on the same Docker network, binding host ports is not required for communication between services. Exposing these services via bound host ports represents a security issue in single-server production environments; only the nginx proxy port needs to be accessible outside of the docker network for upstream traffic.

Port 3000 (frontend) and 8000 (backend) are required if you want to run plane without the nginx container (or use your own reverse proxy).

pablohashescobar (approved):

pablohashescobar:
Merging this off, thank you for you contribution @Robin-Sch. Will update the documentation for users with and without nginx.

Robin-Sch:
https://github.com/makeplane/plane/issues/1019#issuecomment-1568079328

See @require-dev's suggestion too

Issue #1140, opened 2023-05-25
[feature]: Specify custom port other than 80


markskram:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

I would like to run Plane on a port other than 80. Port 80 is already in use by another server app. It looks like others would like the same thing, as commented here: https://github.com/makeplane/plane/issues/718

### Why should this be worked on?

A server often runs multiple services at once that each have their own web interfaces. They need to run on different ports so they can all be accessed.

pablohashescobar:
Hello, @markskram! Unfortunately, we don't have a direct configuration option to change the default port 80 at the moment. However, you can modify the host port on nginx in the docker-compose.yml and docker-compose-hub.yml files to achieve this.

require-dev:
Wouldn't changing the `80:80` portmapping in the docker-compose.yml & docker-compose-hub.yml to `${NGINX_PORT}:80` fix this issue in the current setup? 

This way you can add NGINX_PORT=80 to the `.env.example` which would allow easy changing of the port, without actually changing the original docker-compose or docker-compose-hub.yml?

pablohashescobar:
Thank you for pointing that out! We will make sure to include this in the .env.example.

pablohashescobar:
Created a PR for the above configuration #1144

markskram:
I cherry-picked the change and verified it works. I am running plane on port 8111 now.

I am experiencing a separate issue where the database connection fails on `django.db.utils.OperationalError: connection to server at "plane-db" (172.19.0.5), port 5432 failed: FATAL:  password authentication failed for user "plane"`. This might be unrelated.

pablohashescobar:
Hello @markskram, I wanted to inform you that we have made some changes to our Docker setup. In order to proceed, please provide the database username, password, and host. You can find the default values in [.env.example](https://github.com/makeplane/plane/blob/develop/.env.example) file. We appreciate your cooperation!

markskram:
What would be an appropriate value for `PGHOST`? I've tried all sorts of values 'localhost', '196.168.2.147' (the devices local IP address), `host.docker.internal`, or even leaving it as `plane-db`. They all give me "Connection refused" when planebackend attempts to connect on port 5432.

markskram:
I am attempting to use the postgres that comes with the docker setup. Should I be running my own?

markskram:
[...]

PR #1144, opened 2023-05-26
chore: configuration for nginx port

pablohashescobar:
chore: configuration for nginx port.

This is related to #1140


vihar (approved):

markskram (approved):
I am able to run on a different port by using this change in .env.example.

PR #1145, opened 2023-05-26
docs: update self hosting setup script

pablohashescobar:
docs: update self hosting setup script


markskram (approved):
This fixed it for me. I was able to access from another computer when I added the protocol onto the front.

vihar (approved):

PR #1176, opened 2023-06-02
fix: NextJS Api Route Nginx Proxy

### LINKED ISSUE #981: [bug]: Unsplash proxy API inaccessible

### Current behavior

When loading the project settings page or attempting to make a new project on the current `develop` branch (160b4a4), the image picker tries to load `<host>/api/unsplash` to proxy the actual Unsplash API call. This endpoint returns 404 because it's attempting to read from the Django `apiserver` service on port 8000 rather than the web service on port 3000 (where the `/api/unsplash` handler appears to be served from).

This prevents the project image from being able to be changed.

### Steps to reproduce

1. Self-host Plane `develop` branch on localhost
2. Create a new project in your workspace
3. Go to the workspace settings
4. Observe an XHR GET error in the console
5. Unable to change the workspace splash image

### Browser

Mozilla Firefox

### Version

Self-hosted

piyushgarg-dev:
Fixes #981

### Summary
The bug occurred because there was a nginx proxy setup on `/api/` which was forwarding the request to `http://planebackend:8000/api/` which is a Django server. The call to `/api/unspash` was intended to go to the API route of NextJS on port 3000 but it was being forwarded to PORT:8000. So, I have rewritten the NextJS Api routes to `api/next` and added an nginx rule on `/api/next/` to forward to `localhost:3000`.


sriramveeraghanta:
We wanted to take a different approach in solving this one using the nginx conf file. We will make those change in our next release. Please feel free to open this PR in case you need any more help.

PR #1181, opened 2023-06-03
feat: plane proxy setup

pablohashescobar:
feat: plane proxy setup

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -53,6 +53,8 @@ GITHUB_CLIENT_SECRET="" # For fetching release notes
 
 # Settings related to Docker
 DOCKERIZED=1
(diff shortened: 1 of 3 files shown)


Robin-Sch:
I think with this docker compose file, you only have to download the compose file and setup.sh file, so you could change the README too (it's not required to download the whole repo anymore, just those two files)

Robin-Sch:
Also I'm not exactly sure which services nginx proxies, but you can remove the ports from those containers (like the plane-web, plane-api and plane-minio)

pablohashescobar:
Hi @Robin-Sch, thank you for your valuable inputs. I appreciate your feedback, and I will make the necessary changes and provide an update accordingly.

PR #1270, opened 2023-06-10
chore: update docker setup to mount env file

pablohashescobar:
chore: update docker setup to mount env file


aaryan610 (approved):

groone:
It is better if each container's parameters are listed specifically as it was before.

PR #1358, opened 2023-06-22
fix: cannot change the state if it's the only state in group 

### LINKED ISSUE #1329: [bug]: able to remove all the states from a group

### Current behavior

According to the current workflow, I see that a project cannot have an empty state group. If I try to delete the last state of any group, it doesn't allow me to do so.

image

However, there's a loophole I found, the user cannot delete the only state of a group, but they can change the group of that state by using the dropdown menu, like this-

image

### Steps to reproduce

1. Go to project settings.
2. Go to the `States` menu.
3. Try changing the group of any state that is the only one in its group.

### Browser

Google Chrome

### Version

Cloud

Palanikannan1437:
* Closes #1329 

## Issue in brief
Though an user cannot delete the only state of a group, one could change the group of that state by using the dropdown menu.

## Suggested Fixes/Changes 

- Check if `groupLength` is 1 (to see it's the only state in the group
- Disable the Dropdown if groupLength is 1
- Add a tooltip to suggest why the DropDown is disabled and the action to perform when it's enabled

## Updated screenshots of the above Suggested Changes

- If only one state in the group
Screenshot 2023-06-22 at 1 10 29 PM

- If multiple states present in the group
Screenshot 2023-06-22 at 1 44 11 PM

### Before
https://github.com/makeplane/plane/assets/73993394/0a86f0f0-a5cf-415f-8180-eb0971e5cf55

### After
https://github.com/makeplane/plane/assets/73993394/8a939b45-d18c-4ce2-bdf1-98558cb35565

## Potential Further Steps
- Enforce this login on the backend as well

### DIFF
--- a/apps/app/components/states/create-update-state-inline.tsx
+++ b/apps/app/components/states/create-update-state-inline.tsx
@@ -15,7 +15,7 @@ import stateService from "services/state.service";
 // hooks
 import useToast from "hooks/use-toast";
 // ui
-import { CustomSelect, Input, PrimaryButton, SecondaryButton } from "components/ui";
+import { CustomSelect, Input, PrimaryButton, SecondaryButton, Tooltip } from "components/ui";
 // types
 import type { ICurrentUserResponse, IState, IStateResponse } from "types";
 // fetch-keys
@@ -28,6 +28,7 @@ type Props = {
   onClose: () => void;
   selectedGroup: StateGroup | null;
   user: ICurrentUserResponse | undefined;
+  groupLength: number;
 };
 
 export type StateGroup = "backlog" | "unstarted" | "started" | "completed" | "cancelled" | null;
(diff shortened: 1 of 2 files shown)


Palanikannan1437:
@aaryan610 Would love your feedback on this, thank you! 

PS: I'll squash the commits once all the necessary changes are made before the merge, thanks

apps/app/components/states/create-update-state-inline.tsx
```diff
@@ -228,22 +229,27 @@ export const CreateUpdateStateInline: React.FC<Props> = ({
...
-                </CustomSelect.Option>
-              ))}
-            </CustomSelect>
+            <Tooltip tooltipContent={groupLength === 1 ? "Cannot have an empty group." : "Choose State"}  >
+              <div>
+                <CustomSelect
+                  disabled={groupLength === 1}
+                  value={value}
+                  onChange={onChange}
+                  label={
```
aaryan610, line 238: Please make the text muted when the select dropdown is disabled. Use the class `text-brand-secondary`.

Palanikannan1437: Ohh sure, just to confirm, I made the the changes such that the text within that dropdown should have a muted appearance...and I've also updated the PR's description! Thanks

aaryan610 (approved):

Issue #1356, opened 2023-06-22
[feature]: Save label on press enter


ChandanJal:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Should be save the data from input box of label component on pressing enter from keyboard.

### Why should this be worked on?

This will make fast generation of creating a small task such as label.

aaryan610:
This issue will be fixed with PR #1527.

PR #1396, opened 2023-06-26
feat: submit label form when click enter on the input

ChandanJal:
feat: save on enter key press (https://github.com/makeplane/plane/issues/1356)


kblueberry (approved):

kblueberry (approved):

kblueberry (approved):

aaryan610:
Hi @ChandanJal, thanks for your contribution.

The approach used by you here would be an overkill as it's already a react-hook-form component and doesn't require the use of `useRef`.

This issue is fixed in PR #1527.

PR #1438, opened 2023-06-29
chore: Dry up docker compose environment variables

mdp:
* chore: Use docker compose extensions to dry up duplicate environment variables between worker and api services.
See: https://docs.docker.com/compose/compose-file/11-extension/


pablohashescobar (approved):

pablohashescobar:
Hi @mdp, thank you for contributing to Plane.

Issue #1301, opened 2023-06-15
[bug]: date picker is hidden in the board view


kirtan403:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

The date picker is hidden when the items are less

![image](https://github.com/makeplane/plane/assets/7900938/20c2b06e-20dc-4b00-ad22-3d55c00f0826)

### Steps to reproduce

1. Create 1 issue 
2. Click on the due date
3. You will see the date picker is hidden due to the height of the list

### Browser

Google Chrome

### Version

Self-hosted

Flavsditz:
I will give it a go on fixing this one.

Just to add to the information: 
- This bug only appears when displaying the issues on the "column" layout (see image below)
- When in other views the date view looks fine.

<img width="185" alt="image" src="https://github.com/makeplane/plane/assets/1968234/de00ad58-d86a-426e-b85d-2d2caa459081">
@Palanikannan1437 would you know (or can you point to who does) what the icons are meant to be? Having a mouseover text here wouldn't be a bad idea. Based on the symbols I thought this was about "grouping" and was expecting the columnar view to be the 4th button (before the Gannt/Roadmap one)

Palanikannan1437:
Hey @Flavsditz, that'd be so awesome!!

A tooltip on these definitely sounds like a cool idea! 

From the left, the icons are
1) list view
2) board view
3) calendar view
4) I'm not able to see it in the cloud version/the latest develop branch (could you please help me out with your environment/setup?)
5) gantt chart view

This is how it looks for me,

<img width="1163" alt="image" src="https://github.com/makeplane/plane/assets/73993394/b6fb6777-b9ad-4fdb-b082-9f9d36d2c478">

@aaryan610 and Plane crew folks would be better able to guide you here, thanks

Flavsditz:
@Palanikannan1437 I will put this in the PR.

I found the issue which is that we have `overflow: hidden` and `overflow-y: scroll` set to the column and that is an issue for the calendar when it gets poped-up.

I have been testing different approaches to conditionally change the rules only when the height requires or we have a minimum for the calendar.
- Container query: has a weird effect where the height doesn't get recognized and defaults to 0 rendering the height query unusable and thus not being an option, although it would be the most elegant solution
- JS Size event: checking if the column has a certain size and adapting the classes accordingly works great but during re-rendering the size of the column changes causing weird glitches. There are ways around it using a debounce and other tricks but this is just patchy code at best.
[...]

PR #1476, opened 2023-07-05
Fix calendar display hidden on "kanban" view

Flavsditz:
This PR actually tackles 2 elements as per discussion in the thread of ticket #1301 

1. It fixes the issue with the date-picker not properly showing. 
   1. The changes for this are all contained in the  `single-board.tsx` file
   2. It involved setting a minimum height for the columns and until then they are not having any kind of `overflow: hidden|scroll` set, which is the default. Once the min. height is achieved this gets turned on.
   3. This guarantees that there is enough space to render the calendar
2. It adds a tooltip on the view icons which are not really 100% understandable without it (from personal experience)
   1. This uses the `Tooltip` from `ComponentUI`
   2. I've noticed we also have `Popover2` as a concurrent option for tooltips which might be better for the size of the application if one only is used.
 3. Finally I also updated the props from the `Tooltip` since I used it, to remove an unused prop and add the control for the "show delay", i.e. the time in ms to wait to show the tooltip when the user hovers the icon

Happy to hear any feedback


Flavsditz:
I am not seeing the conflict that is being accused on the file `issues-view-filter.tsx` and I have it clean in my branches locally. Can anyone help this one out?

apps/app/components/ui/tooltip.tsx
```diff
@@ -5,7 +5,7 @@ import { Tooltip2 } from "@blueprintjs/popover2";
 type Props = {
   tooltipHeading?: string;
   tooltipContent: string | JSX.Element;
-  triangle?: boolean;
+  hoverOpenDelay?: number;
```
aaryan610, line 8: The prop name doesn't need to be this complex, you can rename it to just `delay`

Flavsditz: I used this name since it is the Tooltip2 prop name.
While I agree with you, this might pose an issue in the future if we decide to add the other time based controls that Tooltip2 uses. It has 3:
* `hoverOpenDelay`
* `hoverCloseDelay`
* `transitionDuration`

I would say I could change it to `openDelay` at least wdyt?

aaryan610: Yes, that'd work.

apps/app/components/core/filters/issues-view-filter.tsx
```diff
@@ -56,53 +56,63 @@ export const IssuesFilterView: React.FC = () => {
...
+            type="button"
+            className={`grid h-7 w-7 place-items-center rounded p-1 outline-none duration-300 hover:bg-brand-surface-2 ${
+              issueView === "spreadsheet" ? "bg-brand-surface-2" : ""
+            }`}
+            onClick={() => setIssueView("spreadsheet")}
+          >
+            <Icon iconName="table_chart" className="text-brand-secondary" />
+          </button>
+        </Tooltip>
+        <Tooltip tooltipContent="Gant Chart View" position="bottom" theme="dark" openDelay={500}>
```
aaryan610, line 103: Please update the spelling of `Gantt`.

aaryan610:
HI @Flavsditz, thanks for resolving this error but due to the change of code structure, we cannot merge this PR as it is now outdated. But this issue has now been resolved in #1526.

Looking for more contributions from you in the future.

PR #1505, opened 2023-07-12
fix: workspace upload image accepting any kind of file

chandanvc:
fixed: bug of workspace upload image accepting any king of file

### DIFF
--- a/apps/app/components/core/modals/image-upload-modal.tsx
+++ b/apps/app/components/core/modals/image-upload-modal.tsx
@@ -44,6 +44,9 @@ export const ImageUploadModal: React.FC<Props> = ({
   }, []);
 
   const { getRootProps, getInputProps, isDragActive } = useDropzone({
+    accept: {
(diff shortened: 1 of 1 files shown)


apps/app/components/core/modals/image-upload-modal.tsx
```diff
@@ -44,6 +44,9 @@ export const ImageUploadModal: React.FC<Props> = ({
   }, []);
 
   const { getRootProps, getInputProps, isDragActive } = useDropzone({
+    accept: {
+      "image/*": [".jpeg", ".png", ".jpg", ".webp"],
+    },
     onDrop,
```
aaryan610, line 50: Can you also add a small message informing the user about the type of files accepted?

chandanvc: Sure

aaryan610:
Hi @chandanvc, thanks for your contribution, but we have now fixed this issue in #1529.

PR #1645, opened 2023-07-24
fix: link validation fix

anmolsinghbhatia:
fix: 
- link validation fix

### DIFF
--- a/apps/app/components/core/sidebar/links-list.tsx
+++ b/apps/app/components/core/sidebar/links-list.tsx
@@ -12,7 +12,7 @@ type Props = {
     created_by: string;
     created_by_detail: IUserLite;
     metadata: any;
-    title: string;
+    title: string | null;
     url: string;
   }[];
   handleDeleteLink: (linkId: string) => void;
@@ -55,7 +55,7 @@ export const LinksList: React.FC<Props> = ({ links, handleDeleteLink, userAuth }
               <LinkIcon className="h-3.5 w-3.5" />
             </div>
             <div>
-              <h5 className="w-4/5 break-words">{link.title}</h5>
+              <h5 className="w-4/5 break-words">{link.title ?? link.url}</h5>
               <p className="mt-0.5 text-custom-text-200">
                 Added {timeAgo(link.created_at)}
                 <br />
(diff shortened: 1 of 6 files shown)


apps/app/components/core/sidebar/links-list.tsx
```diff
@@ -12,7 +12,7 @@ type Props = {
     created_by: string;
     created_by_detail: IUserLite;
     metadata: any;
-    title: string;
+    title: string | null;
```
aaryan610, line 15: The title cannot be null, it can be an empty string.

apps/app/services/issues.service.ts
```diff
@@ -437,7 +437,7 @@ class ProjectIssuesServices extends APIService {
     issueId: string,
     data: {
       metadata: any;
-      title: string;
+      title: string | null;
```
aaryan610, line 440: The title cannot be null, it can be an empty string.

apps/app/services/modules.service.ts
```diff
@@ -198,7 +198,7 @@ class ProjectIssuesServices extends APIService {
     moduleId: string,
     data: {
       metadata: any;
-      title: string;
+      title: string | null;
```
aaryan610, line 201: The title cannot be null, it can be an empty string.

apps/app/types/issues.d.ts
```diff
@@ -62,7 +62,7 @@ export interface IIssueParent {
 }
 
 export interface IIssueLink {
-  title: string;
+  title: string | null;
```
aaryan610, line 65: The title cannot be null, it can be an empty string.

apps/app/types/issues.d.ts
```diff
@@ -96,7 +96,7 @@ export interface IIssue {
     created_by_detail: IUserLite;
     id: string;
     metadata: any;
-    title: string;
+    title: string | null;
```
aaryan610, line 99: The title cannot be null, it can be an empty string.

apps/app/types/modules.d.ts
```diff
@@ -32,7 +32,7 @@ export interface IModule {
     created_by_detail: IUserLite;
     id: string;
     metadata: any;
-    title: string;
+    title: string | null;
```
aaryan610, line 35: The title cannot be null, it can be an empty string.

dakshesh14 (approved):

pablohashescobar (approved):

PR #1704, opened 2023-07-28
fix: project cover image upload

dakshesh14:
fix: 

- popover close on image upload
- overflow


apps/app/components/project/create-project-modal.tsx
```diff
@@ -227,7 +227,7 @@ export const CreateProjectModal: React.FC<Props> = ({ isOpen, setIsOpen, user })
                       <XMarkIcon className="h-5 w-5 text-white" />
                     </button>
                   </div>
-                  <div className="hidden group-hover:block absolute bottom-2 right-2">
+                  <div className="absolute bottom-2 right-2">
```
aaryan610, line 230: The `Change Cover` button should only be visible when hovered over the cover image.

aaryan610 (approved):

PR #1774, opened 2023-08-02
fix: remove custom theme on sign out

aaryan610:
fix:

- remove custom theme on sign out.


sriramveeraghanta:
This approach will remove the custom styling whenever the user signs out of the platform. When he signs back in, he needs to go to user settings again to add his custom styling. Also, it doesn't work when the user token expires and moves to the login page.

PR #1776, opened 2023-08-02
@thomasmazon: fixing some issues to run docker local

thomasmazon:


pablohashescobar:
Hi @thomasmazon, thanks for the contribution. We also have an another file `docker-compose-hub.yml` which is used to run images from the docker hub can you also make the necessary changes to that file as well.

pablohashescobar:
Hey @thomasmazon, we have updated the docker setup for all the environments including the [local setup](https://github.com/makeplane/plane/blob/develop/docker-compose-local.yml) which can be used for local development. Closing this PR. Thank you for your contribution.

PR #1807, opened 2023-08-08
chore: updated user permission for project and workspace settings

anmolsinghbhatia:
chore: 
- user permissions for project and workspace update and delete actions have been updated.

### DIFF
--- a/apps/app/components/workspace/delete-workspace-modal.tsx
+++ b/apps/app/components/workspace/delete-workspace-modal.tsx
@@ -51,11 +51,8 @@ export const DeleteWorkspaceModal: React.FC<Props> = ({ isOpen, data, onClose, u
 
   const handleClose = () => {
     setIsDeleteLoading(false);
-    const timer = setTimeout(() => {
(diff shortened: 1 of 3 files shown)


apps/app/components/workspace/delete-workspace-modal.tsx
```diff
@@ -50,8 +50,13 @@ export const DeleteWorkspaceModal: React.FC<Props> = ({ isOpen, data, onClose, u
   const canDelete = confirmWorkspaceName === data?.name && confirmDeleteMyWorkspace;
 
   const handleClose = () => {
-    onClose();
     setIsDeleteLoading(false);
+    const timer = setTimeout(() => {
```
sriramveeraghanta, line 54: Why do you need this?

sriramveeraghanta: You can simply clear the state and close the modal right.

sriramveeraghanta (changes requested):

PR #1817, opened 2023-08-09
style: displaying email in members page depending on user role

dakshesh14:
style: 

- displaying email on members page depending on user role

### DIFF
--- a/apps/app/pages/[workspaceSlug]/projects/[projectId]/settings/members.tsx
+++ b/apps/app/pages/[workspaceSlug]/projects/[projectId]/settings/members.tsx
@@ -10,8 +10,9 @@ import projectService from "services/project.service";
 import workspaceService from "services/workspace.service";
 // hooks
 import useToast from "hooks/use-toast";
-import useProjectDetails from "hooks/use-project-details";
 import useUser from "hooks/use-user";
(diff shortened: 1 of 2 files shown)


apps/app/pages/[workspaceSlug]/projects/[projectId]/settings/members.tsx
```diff
@@ -194,20 +202,27 @@ const MembersSettings: NextPage = () => {
                               className="absolute top-0 left-0 h-full w-full object-cover rounded-lg"
                             />
                           ) : (
-                            member.display_name.charAt(0)
+                            (member.display_name || member.email).charAt(0)
```
Comment (sriramveeraghanta) [inline apps/app/pages/[workspaceSlug]/projects/[projectId]/settings/members.tsx:205]: In Case of Null value returned as in an edge case, how are we handling it.

sriramveeraghanta (changes requested):

apps/app/pages/[workspaceSlug]/projects/[projectId]/settings/members.tsx
Comment (dakshesh14) [reply apps/app/pages/[workspaceSlug]/projects/[projectId]/settings/members.tsx]: I assumed the email won't be null, but I added a null check to avoid an Exception Error.

sriramveeraghanta (approved):

PR #1892, opened 2023-08-17
dev: implemented store for labels state management

dakshesh14:
dev: 

- implemented store for labels state management
- using the newly implemented label store across the site


dakshesh14:
Closing this as we settled on a different approach

PR #1896, opened 2023-08-17
fix: Image resize, Link selector in Modals, Delete/ sync images and so much more 

Palanikannan1437:
1. Added image resize functionality

Currently uses `react-moveable`, but I'll soon add a custom tiptap plugin to remove the dependancy. User can resize the editor by first selecting it followed by dragging the nodes on the right and left and shown in this demo. It's `state` is saved with the "Save" state of the editor.

https://github.com/makeplane/plane/assets/73993394/77a2e673-4f1c-4763-9b15-9052f5488710

2. Fixed link selector by preventing unnecessary form triggers on modals

Previously the link form on submit used to trigger the parent form (in case of Modals) on submit, now I've handled `Enter` key down events and form submit events via a onClick/onKeyDown functions.

Currently only `http` and `https` protocols are allowed and validated against.

3. Images are now uploaded and saved at 35% of the width, and then can be resized as per user's convenience

4. Added alerts behaviour if the user tries to reload while the editor's state isn't saved and synced with Image Uploading as well

https://github.com/makeplane/plane/assets/73993394/e3be25c3-ef39-4433-8be2-c20f73acb3ad

5. Removed shadow to maintain consistency across design in light mode

6. Added custom delete plugin that checks for various cases and deletes images ONLY if the image node is deleted!
    - Added various checks for detecting oldState and newState at every transaction instead of relying on React's state of the editor! 
    - Added checks to detect every set of Transaction and only delete if the node is not in the editor after a "remove"/"displacement" action. 
    - While comparing the oldState and newState, previously the check only took into consideration about the position and id, now it takes in the scenario if the "image" is still present in the entire editor and only then deletes it from the backend!


apps/app/components/pages/create-update-block-inline.tsx
```diff
@@ -311,13 +312,14 @@ export const CreateUpdateBlockInline: React.FC<Props> = ({
 
                 return (
                   <TiptapEditor
+                    workspaceSlug={workspaceSlug as string}
                     ref={editorRef}
                     value={
                       value && value !== "" && Object.keys(value).length > 0
```
aaryan610, line 318: Here, `value` and `watch("description_html")` hold the same value, double check is redundant. Can we remove `watch("description_html")`?

apps/app/components/tiptap/bubble-menu/link-selector.tsx
```diff
@@ -66,11 +75,15 @@ export const LinkSelector: FC<LinkSelectorProps> = ({ editor, isOpen, setIsOpen
               <Trash className="h-4 w-4" />
             </button>
           ) : (
-            <button className="flex items-center rounded-sm p-1 text-custom-text-300 transition-all hover:bg-custom-background-90">
+            <button className="flex items-center rounded-sm p-1 text-custom-text-300 transition-all hover:bg-custom-background-90" type="button"
```
aaryan610, line 78: Add a `type="button"` for here.

apps/app/components/pages/create-update-block-inline.tsx
Palanikannan1437: Ohh yeah sure! I've done it!

apps/app/components/tiptap/bubble-menu/link-selector.tsx
Palanikannan1437: Sure thing! Thanks @aaryan610 🙌

sriramveeraghanta (approved):

sriramveeraghanta (approved):

PR #1983, opened 2023-08-25
fix: my profile activity endpoint

aaryan610:
fix:

- added `workspaceSlug` to the my profile activity endpoint.
- activity loader position.

### DIFF
--- a/apps/app/services/user.service.ts
+++ b/apps/app/services/user.service.ts
@@ -101,8 +101,8 @@ class UserService extends APIService {
       });
   }
 
-  async getUserActivity(): Promise<IUserActivityResponse> {
-    return this.get("/api/users/activities/")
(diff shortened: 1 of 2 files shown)


apps/app/services/user.service.ts
```diff
@@ -101,8 +101,8 @@ class UserService extends APIService {
       });
   }
 
-  async getUserActivity(): Promise<IUserActivityResponse> {
-    return this.get("/api/users/activities/")
+  async getUserActivity(workspaceSlug: string): Promise<IUserActivityResponse> {
```
sriramveeraghanta, line 104: change the name to getUserWorkspaceActivity

sriramveeraghanta (changes requested):

sriramveeraghanta (approved):

PR #1996, opened 2023-08-28
fix: issue exports in self hosted instances

pablohashescobar:
fix: issue exports in self hosted instances


apiserver/plane/bgtasks/export_task.py
```diff
@@ -338,7 +375,7 @@ def issue_export_task(provider, workspace_id, project_ids, token_id, multiple, s
         exporter_instance.status = "failed"
         exporter_instance.reason = str(e)
         exporter_instance.save(update_fields=["status", "reason"])
-
+        print(e)
```
sriramveeraghanta, line 378: remove this

sriramveeraghanta (changes requested):

NarayanBavisetti (approved):

PR #2128, opened 2023-09-08
fix: husky was removed in commit #2086, but prepare still uses it

thomasync:
currently, _docker compose up -d_, triggers this error:
```
[4/4] Building fresh packages...
$ husky install
/bin/sh: husky: not found
error Command failed with exit code 127.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
     "packages/*"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


sriramveeraghanta (approved):

sriramveeraghanta:
Thanks for finding it out and fixing it.

PR #2135, opened 2023-09-10
[chore] Updated web and deploy backend configuration for reverse proxy & decoupled Plane Deploy URL generation for web

henit-chobisa:
## Tasks accomplished:
- [x] Changed the request URL for web and deploy containers from "localhost:8000" and arbitrary URLs to the same URL the application is running on.
- [x] Updated the docker-compose configuration to use reverse proxy for backend requests.
- [x] Decoupled the URL for Plane Deploy in the web environment variable.
- [x] Generated the Plane Deploy link using the host URL of the web project with a suffix, If the URL is explicitly specified, the web project will use that URL for Plane Deploy.

## Previous state:
- Containers in the docker-compose configuration did not use reverse proxy for backend requests.
- The web and deploy containers made requests to "localhost:8000" and arbitrary URLs.
- The URL for Plane Deploy had to be manually specified in the web environment variable.

## Current state:
- The web and plane deploy projects has been updated to use reverse proxy for backend requests, making request to `/api` on the same host .
- The web and deploy containers now make request on the same host.
- The URL for Plane Deploy is decoupled from the web project and is generated using the host URL of the web project with a suffix `spaces` in case if the URL is explicitly specified, the web project will use that URL for Plane Deploy.

## How I tested the changes
✅ Ran Individual Builds for all three projects
✅ Ran Individual Containers and tested them with environment variables provided and without giving any environment variables for self hosted deployments
✅ Built Individual Docker Compose Services for testing individually with the given environment variables
✅ Built complete docker compose with the given set of environment variables.

## Keep this in mind
- When working with self hosted part, the `NEXT_PUBLIC_API_BASE_URL` and `NEXT_PUBLIC_DEPLOY_URL` are given as empty strings at the time of the build, which means that the request for both `API` and `DEPLOY` should be made on the same host as web.
- In case of changes in the above two variables, or defining explicit endpoint for `NEXT_PUBLIC_API_BASE_URL` & `NEXT_PUBLIC_DEPLOY_URL`,  these envs can be passed as build args in docker-compose file.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -12,8 +12,13 @@ cp ./apiserver/.env.example ./apiserver/.env
 # Generate the SECRET_KEY that will be used by django
 echo -e "SECRET_KEY=\"$(tr -dc 'a-z0-9' < /dev/urandom | head -c50)\""  >> ./apiserver/.env
 
-echo -e "\nNEXT_PUBLIC_API_BASE_URL=$1\nWEB_URL=$1"  >> ./web/.env
-echo -e "\nNEXT_PUBLIC_API_BASE_URL=$1\nWEB_URL=$1"  >> ./space/.env
+if [ -n "$1" ]
+then
+    echo "hello"
+    echo -e "\nNEXT_PUBLIC_API_BASE_URL=$1\nWEB_URL=$1"  >> ./web/.env
+    echo -e "\nNEXT_PUBLIC_API_BASE_URL=$1\nWEB_URL=$1"  >> ./space/.env
+fi
+
 
 # Generate Prompt for taking tiptap auth key
 echo -e "\n\e[1;38m Instructions for generating TipTap Pro Extensions Auth Token \e[0m \n"
--- a/setup.sh
+++ b/setup.sh
@@ -14,7 +14,6 @@ echo -e "SECRET_KEY=\"$(tr -dc 'a-z0-9' < /dev/urandom | head -c50)\""  >> ./api
 
 if [ -n "$1" ]
 then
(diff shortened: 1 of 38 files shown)


setup.sh
```diff
@@ -12,8 +12,13 @@ cp ./apiserver/.env.example ./apiserver/.env
 # Generate the SECRET_KEY that will be used by django
 echo -e "SECRET_KEY=\"$(tr -dc 'a-z0-9' < /dev/urandom | head -c50)\""  >> ./apiserver/.env
 
-echo -e "\nNEXT_PUBLIC_API_BASE_URL=$1\nWEB_URL=$1"  >> ./web/.env
-echo -e "\nNEXT_PUBLIC_API_BASE_URL=$1\nWEB_URL=$1"  >> ./space/.env
+if [ -n "$1" ]
+then
+    echo "hello"
```
sriramveeraghanta, line 17: Hello????

sriramveeraghanta (approved):

leikoilja:
@henit-chobisa, what is the reason/benefit to set `NEXT_PUBLIC_API_BASE_URL` as empty strings during the build process? is that some sort of a hack? 
i m experiencing the same issue as #2434 describes, but looking at the PR changes, i can't quite understand how setting those build args in dockerfile would make container omit env variable.
I ssh'ed into my running `web` container on the latest `v0.13.2-dev` and can see that he container itself sees env variable just fine 
image

however the API calls the frontend does goes to the same host as `web` is hosted on
image

Issue #1273, opened 2023-06-11
[bug]: Change Cover button is missing on self-hosted version


mikhail-shevtsov-wiregate:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

On latest release v0.7.1 when creating new project or editing project `Change Cover` button is missing on self-hosted version.

### Steps to reproduce

1. Go to Projects;
2. Click `Add Project` button;
or
1. Go to Project Settings;

In both cases the button `Change Cover` is Missing

### Browser

Mozilla Firefox

### Version

Self-hosted

ChandanJal:
Yes, mine also not showing

rhea0110:
Hey @mikhail-shevtsov-wiregate, thank you for reporting this bug. Our engineers have already started working on fixing this issue, and it will be live in some time. We will keep you updated on the progress made.

Thank you!

vihar:
thanks for flagging this, @mikhail-shevtsov-wiregate. this issue is fixed on https://github.com/makeplane/plane/pull/2375.

PR #2170, opened 2023-09-13
fix: change cover button not showing

ix64:
fixes https://github.com/makeplane/plane/issues/1273


aaryan610:
Hi @ix64, thanks a lot for your contribution. However, we have now changed the workflow for displaying the Unsplash images with PR #2375 and it no longer requires the env variable. The latest changes are live on our `master` branch.

Looking forward to more contributions from you. 😄

PR #2209, opened 2023-09-19
[fix]: Removing dependency on tiptap pro extension

Palanikannan1437:
## Brief Description
The Tiptap token was used only for downloading an extension that previously "enabled" us to manage image deletions efficiently so that s3 storage is cleaned up properly if an image gets deleted in the editor, but now we're using the source from s3 to detect this instead since with the new approach I figured out a way to do that.

Here's a PR that got us one step forward toward that by removing the dependency https://github.com/makeplane/plane/pull/2092

### DIFF
--- a/.github/workflows/Build_Test_Pull_Request.yml
+++ b/.github/workflows/Build_Test_Pull_Request.yml
@@ -33,14 +33,9 @@ jobs:
             deploy:
               - space/**
 
-      - name: Setup .npmrc for repository
-        run: |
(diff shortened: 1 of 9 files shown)


Palanikannan1437:
@henit-chobisa please look into removing it from our DevOps and self hosting pipelines 🙌

sriramveeraghanta:
Updated setup script and github actions.

sriramveeraghanta (approved):

PR #2248, opened 2023-09-22
fix: hard-coded percentage value for completed cycle

dakshesh14:
fix: 

- hard-coded percentage value for completed cycles


dakshesh14:
Closing since fixed by this #2250

PR #2252, opened 2023-09-22
[fix] minio exposing current bucket contents publically

henit-chobisa:
## Tasks
- [x] Added Custom `mc` minio client by creating `Dockerfile.minio` 
- [x] Added Minio Configuration in `minio.sh`
- [x] Updated the existing minio configuration service to new custom docker build

## End Result
I am no longer able to access, content on `/uploads` route
image


sriramveeraghanta:
outdated changes

PR #2321, opened 2023-09-29
Feats: Add localization

thomasync:
## Add a global translation for Plane

- [X] A script to help find all the texts to be translated
- [X] A script to reload locales in Plane, so you don't have to run start.sh every time
- [X] Global function to import in every file (I chose to put the function in **useMobxStore**, as it was already imported in many files
- [X] Add FR Locale
- [X] Resolve the errors that occur when I touch files in web/layouts/*.
- [X] Review each file
- [X] Translate web files 
- [ ] Translate space files (not yet planned)

## Technical specifications:

The majority of commits are simply the replacement of strings by **localized(string)**, hence the large number of modified files.

### Files added with business logic are only : 

- **web/components/core/localization/***
- **web/helpers/localization.helper.ts**
- **web/store/locale.ts**

### There are also optional, but very useful, scripts for working on localization: 

- **scripts/localization.js**: Find all untranslated strings in **en_US.json** and add them. But also add it to all other translations with a _null_ to be edited.
- **scripts/reload-locales.sh**: Returns modified translations directly, without having to run **start.sh** again.


scripts/localization.js
thomasync, line 1: **Optional file**
Search for translation in all .ts, .tsx files

thomasync (commented):
Reported without bugs or problems.

Kritikkkaaa:
Hey @thomasync 
We have this in our pipeline and will be pushed in one of our upcoming releases.

thomasync:
> Hey @thomasync We have this in our pipeline and will be pushed in one of our upcoming releases.

Does "this" mean this PR or this feature?

PR #2332, opened 2023-10-02
refactor: remove unnecessary state.

david-sling:
- Removed an unnecessary state and an effect
- Renamed `handleTheme` to be `handleToggleTheme` as it seemed to be a more descriptive function name


dakshesh14:
Hello @david-sling, this PR will trigger a hydration error.

sriramveeraghanta:
Thanks for your contribution. But unfortunately, we cant merge your changes as it is going to trigger a hydration error.

PR #2352, opened 2023-10-03
[FEATURE] Enabled User `@mentions` in tiptap editor component

henit-chobisa:
## Description
The current pull request enables the usage of `user-mentions` in the tiptap editor, with features such as self highlighting, redirection and sending notifications from the backend.
image

### Sending Notification on being mentioned with context of being mentioned
Irrespective of a person being a subscriber of the issue or not, if he's mentioned in a particular issue, he will be sent a notification for being mentioned, along with that he will be added as a subscriber to the issue, to receive further updates on the issue. For testing the possible set of cases and scenarios please go through https://henit.notion.site/feat-mentions-c96dc6cfe53a4e2eafb0acb252cb80c4?pvs=4

image

### Redirecting on the user profile, when clicked on the mention
The mention acts as a button, hence if you have provided the `redirect_uri` in the underlying data structure `IMentionSuggestion[]`, you can see it redirecting to the user's profile. In future we can use this for redirecting to specific sets of views such as cycles or issues itself.

https://github.com/makeplane/plane/assets/72302948/63f6ad5b-5682-4706-92d3-32c305298caf

### Dynamic Highlights of entities in an issue
The mention feature also come with highlightin, which makes it extremely easy to spot a particular highlight, which can be anything from highlighting the current user from the content or highlighting a set of issues from a query of "Show me all the issues where a particular set of issues are mentioned". Currently the highlighting works only in the `issues/form.tsx`, and in `side-peak-view`.
image

**For Example:** : If the highlight group is for the current user, which means that in every issue, if the current user is mentioned he must be highlighted, in this case it's dynamic because for an issue where `@crew` and `@captain` both are mentioned, if `@crew` is signed in, he must be highlighted and so as captain, which makes us dynamically update the values. Below is the picture of the implementation, which show how dynamic highlights work for two different highlight groups.

**Issue View of the Captain**
image

**Issue View for the Crew** 
image


apiserver/plane/bgtasks/issue_activites_task.py
```diff
@@ -1428,12 +1526,58 @@ def issue_activity(
...
+
+            # Send those Bulk Notifications
+            Notification.objects.bulk_create(
+                mention_notifications, batch_size=100)
 
         return
     except Exception as e:
         # Print logs if in DEBUG mode
         if settings.DEBUG:
+            print("====================EXCEPTION OCCURED====================")
```
pablohashescobar, line 1580: @henit-chobisa please remove this line before merge

apiserver/plane/bgtasks/issue_activites_task.py
```diff
@@ -1428,12 +1526,58 @@ def issue_activity(
...
+            mention_notifications = []
+
+            # Send Notifications to the Mentioned and Add Mentioned as Subscribers
+            for mention_id in new_mentions:
+                for issue_activity in issue_activities_created:
+                    if (issue_activity.verb == "created" or issue_activity.verb == "updated"):
+                        mention_notifications.append(
+                            Notification(
+                                workspace=project.workspace,
+                                sender="in_app:issue_activities",
```
pablohashescobar, line 1544: Here can we add an addtional identifier `in_app:issue_activities:mention`

apiserver/plane/bgtasks/issue_activites_task.py
```diff
@@ -1428,12 +1526,58 @@ def issue_activity(
...
+                                        "actor": str(issue_activity.actor_id),
+                                        "new_value": str(issue_activity.new_value),
+                                        "old_value": str(issue_activity.old_value),
+                                    },
+                                },
+                            )
+                        )
+
+            # Send those Bulk Notifications
+            Notification.objects.bulk_create(
```
pablohashescobar, line 1573: Can we do a single bulk create both for mentions and other notifications also.

pablohashescobar (changes requested):

apiserver/plane/bgtasks/issue_activites_task.py
henit-chobisa: Yes, absolutely. I fixed it.

apiserver/plane/bgtasks/issue_activites_task.py
henit-chobisa: I've modified it ✌🏼

apiserver/plane/bgtasks/issue_activites_task.py
henit-chobisa: I have removed it. Actually that like was kinda intentional, because celery serves exceptions as warning only, so I thought, it can be better to have a symbol.

pablohashescobar:
Closing this PR as updated changes are already merged

PR #2391, opened 2023-10-05
fix: maintain scroll position when switching card views (closes #2374)

### LINKED ISSUE #2374: [bug]: when exiting an error card, the page returns to the first column on the board, not the one from which the error card was opened

### Current behavior

When I click on the back button, I go back to the first column on the board, not the one from which I opened the card
I expect that when closing the card the page will return to the column from which the defect card was opened

https://github.com/makeplane/plane/assets/83757237/22a128e6-036d-4183-86a3-252eff2f56c1

### Steps to reproduce

1. Opening the defect card from the last column on the right on the board
2. Click on the "Back" button

### Browser

Google Chrome

### Version

Self-hosted

ousszizou:


aaryan610:
Hi @ousszizou, thank you very much for your contribution. However, using sessionStorage might not be the ideal approach to achieve the solution, as sessionStorage persists until the browser is closed, which could lead to unexpected behavior.

We appreciate your efforts and look forward to more contributions from you. If you have any questions or would like further guidance, feel free to ask.

Issue #723, opened 2023-04-05
[feature]: Add Helm Chart for deploying on k8


PaleHazy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

A Helm chart for quick and painless deployment on Kubernetes

### Why should this be worked on?

- easy installation in a cloud native environment.

gcarrarom:
Big issue here is related to #884

I will share my work in progress tomorrow as an example of what could be used.

gcarrarom:
Here's where I'm at right now: https://github.com/fancy-whale/fancy-charts/tree/main/charts/plane

Chris-Greaves:
I have been trying to get this running with your Helm chart for a few weeks and I've got it deployed and working with my custom domain (although that did take some code changes). My problem at the moment is I get logged out seconds after logging in. I can't even get through the Onboarding steps before it gets 401s and kicks me back to the login. 

I've tried setting the default locale to London (where I am) to see if it was a timezone issue, but that hasn't worked.

I think I'm at a point of abandoning trying to get this self-hosted because there are just too many bugs.

shanduur:
Any updates on this?

Kelsinga:
There is also this one
https://artifacthub.io/packages/helm/makeplane/plane-ce

vihar:
Apologies for the delay, everyone. Plane now officially supports installation using Helm. You can find the documentation here: [link](https://docs.plane.so/self-hosting/kubernetes). Please don't hesitate to reach out if you encounter any issues.

Issue #1455, opened 2023-07-03
[feature]: add repository


TartanLeGrand:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Deploy each needs docker images to an repository for pull it from an chart helm.

### Why should this be worked on?

Self-hosted with k8s

mguptahub:
@TartanLeGrand

Helm Chart based deployment is available at artifacthub

[![Artifact Hub](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/makeplane)](https://artifacthub.io/packages/search?repo=makeplane)

Closing this issue.

TartanLeGrand:
@mguptahub where we can contribute to the chart ?

PR #2405, opened 2023-10-10
feat: add helm chart

jordibeen:
This PR contains a Helm Chart which enables a quick and painless deployment of Plane's `v0.12.2-dev` on Kubernetes.

People would like to self-host plane on Kubernetes, but because there's no Helm Chart for Plane yet, it's currently mandatory to create custom K8s configuration files for this. After building a custom helm chart for self-hosting Plane, I spent a few hours to be able to give it back to the community. It's loosely based on [Superset's Helm chart](https://github.com/apache/superset/tree/master/helm/superset), and contains an extensive [README.md](https://github.com/jordibeen/plane/tree/helm-chart/helm-chart), as well as a prepackaged redis as postgresql.

## Installation
Configure a [values file](https://helm.sh/docs/chart_template_guide/values_files/) containing overwrites of the environment variables such as credentials and secret keys, and then run:
`helm install plane plane/helm-chart -f [overwrites.yaml]`

## Missing Ingress
Because we use [traefik](https://traefik.io/) for our production cluster, I've added [traefik IngressRoutes](https://doc.traefik.io/traefik/routing/providers/kubernetes-crd/#kind-ingressroute) for routing `plane.example.com` and `plane-api.example.com` to their respective services. Using these requires Traefik to be installed. It'd be nicer if the chart would use the default [Kubernetes Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) CRD's instead. Contributions for this are highly appreciated.

Alternatively, you can verify the installation by port forwarding both the frontend as well as the API.
- `kubectl port-forward svc/plane-web 3000:3000`
- `kubectl port-forward svc/plane-api 8000:8000`

After this, Plane will now be running at http://localhost:3000/ 🎉 

## Missing Minio
I'm using AWS S3 storage. I have decided not to include minio for now.

---
This PR also fixes https://github.com/makeplane/plane/issues/723 and  https://github.com/makeplane/plane/issues/1455.


jordibeen:
Note that I'm currently unable to update the helm-chart to support the latest `v0.13.2-dev` because of [this issue](https://github.com/makeplane/plane/issues/2434). Will update the chart to support the latest version when this issue is fixed.

Kelsinga:
@jordibeen 
in the ingressroutes web.yaml and api.yaml you refer to 
.Values.web.ingressRoute.host and .Values.web.ingressRoute.tls.secretName
.Values.api.ingressRoute.host and .Values.api.ingressRoute.tls.secretName

But I think you meant to reference planeFrontend and planeApi

manishg3:
We have published the helm charts and are available at [Artifact Hub](https://artifacthub.io/packages/helm/makeplane/plane-ce)

rouja:
Hello,

@mguptahub where is the source  code of the helm chart that you published ? Is-it possible to contrib  to it ?

PR #2530, opened 2023-10-24
fix: Debounce title and Editor initialization

Palanikannan1437:
## Summary
This PR addresses two significant bug fixes for the new comment editor and the revamped peek views. The first fix ensures the comment editor's features work completely, including handling all edge cases for the enter key and 'shift-enter' key behaviors. The second fix improves the initialization of the editor for dynamically rendering issues using the same editor instance.

## Details
1. **Comment Editor's Features**
The enter key is now remapped to submit the comment, while the 'shift-enter' key is mapped to the default enter key behavior. The edge cases for the various data structures/scenarios present in the comment editor have been handled as follows:

	1. **Code Blocks**: When the cursor is inside a code block, "Shift-Enter" inserts a newline character at the current cursor position, creating a new line within the code block.
	2. **List Items**: When the cursor is inside a list item, "Shift-Enter" will split the list item at the current cursor position, creating a new list item below the current one.
End of Blocks: When the cursor is at the end of a block, "Shift-Enter" creates a new paragraph below the current block.
	3. **Empty Blocks**: When the cursor is inside an empty block that can be lifted, pressing "Shift-Enter" will lift the block out of its parent node.
	4. **General Case**: In all other cases, pressing "Shift-Enter" will split the block at the current cursor position, creating a new block below the current one.

	Here's a sleek demo video showcasing these changes (note I'm using Shift + Enter here as Enter key defaults to Submitting the comment itself)

https://github.com/makeplane/plane/assets/73993394/7df825de-f327-48ff-817a-f3b9dcd9bb71

2. **Editor Initialization for Dynamically Rendering Issues**

The editor initialisation for dynamically rendering issues was inconsistent due to the editor initializing with some data, then Tiptap taking over and updating the form state via the onUpdate callback. This caused issues when switching between issues in the peek view, as the editor didn't get re-initialized with the new content from the backend.

To fix this, a new hook has been introduced that initialises the editor completely dynamically. The hook, useInitializedContent, checks if the editor exists and if the value is not an empty string. If these conditions are met, it sets the content of the editor to the cleaned value and sets a ref, `hasInitializedContent`, to true. If the cleaned value is an empty string and hasInitializedContent is true, it sets hasInitializedContent to false.

This change ensures that the editor is correctly initialized with the new content when switching between issues in the peek view. A demo video of the new behaviour is added below. 

https://github.com/makeplane/plane/assets/73993394/5f005863-bc04-40a9-b333-7c5c740f9cf0

3. **Fixed Peek View Issue to use react-hook-form to Submit/render and control data along with `saving` states**

4. **Fixed debounce logic on saving Title of the Issue using a localState to avoid double binding**
The issue previously was that as soon as one starts typing, form save gets triggered after 1500s (debounce save) and then gets back propagated from the backend. Now the issue with this being, if the user starts typing after the form save gets triggered, the save will be overridden by the backend state being propagated back and user looses the typed changes.

To overcome this, we maintain a local state and bind that to the title input controller.

### Minor Changes

1. Changed button type of comment-card for the edit action to "button" to prevent page reloads.
2. Extracted debounce logic out of the core editor into the web project.


packages/editor/core/src/ui/hooks/useEditor.tsx
```diff
@@ -23,26 +27,44 @@ interface CustomEditorProps {
   forwardedRef?: any;
 }
 
-export const useEditor = ({ uploadFile, deleteFile, editorProps = {}, value, extensions = [], onChange, setIsSubmitting, debouncedUpdatesEnabled, forwardedRef, setShouldShowAlert, }: CustomEditorProps) => {
+export const useEditor = ({
+  uploadFile,
```
sriramveeraghanta, line 31: use props as your function argument and de-struct it inside the function for better readability.

packages/editor/core/src/ui/hooks/useReadOnlyEditor.tsx
```diff
@@ -11,17 +11,33 @@ interface CustomReadOnlyEditorProps {
...
       ...editorProps,
     },
     extensions: [...CoreReadOnlyEditorExtensions, ...extensions],
   });
 
+  const hasIntiliazedContent = useRef(false);
+
+  useEffect(() => {
+    if (editor && !value && !hasIntiliazedContent.current) {
+      console.log("setting content");
```
sriramveeraghanta, line 35: remove console logs.

web/components/issues/description-form.tsx
```diff
@@ -81,7 +90,7 @@ export const IssueDescriptionForm: FC<IssueDetailsProps> = (props) => {
     });
   }, [issue, reset]);
 
-  const debouncedTitleSave = useDebouncedCallback(async () => {
+  const debouncedFormSave = useDebouncedCallback(async () => {
```
sriramveeraghanta, line 93: for debounce don't use an external package to do it. You can create a hook that implements debounce for you.

sriramveeraghanta (changes requested):

web/components/issues/description-form.tsx
Palanikannan1437: I'll fix this in a later PR @sriramveeraghanta, don't want to make this PR longer 🙌

sriramveeraghanta (approved):

PR #2544, opened 2023-10-26
[FEATURE] Enabled User `@mentions` and `@mention-filters` in core editor package

henit-chobisa:
## Description
The current pull request enables the usage of `user-mentions` in the tiptap editor, with features such as self highlighting, redirection and sending notifications from the backend.
image

### Sending Notification on being mentioned with context of being mentioned
Irrespective of a person being a subscriber of the issue or not, if he's mentioned in a particular issue, he will be sent a notification for being mentioned, along with that he will be added as a subscriber to the issue, to receive further updates on the issue. For testing the possible set of cases and scenarios please go through https://henit.notion.site/feat-mentions-c96dc6cfe53a4e2eafb0acb252cb80c4?pvs=4

image

### Redirecting on the user profile, when clicked on the mention
The mention acts as a button, hence if you have provided the `redirect_uri` in the underlying data structure `IMentionSuggestion[]`, you can see it redirecting to the user's profile. In future we can use this for redirecting to specific sets of views such as cycles or issues itself.

https://github.com/makeplane/plane/assets/72302948/63f6ad5b-5682-4706-92d3-32c305298caf

### Dynamic Highlights of entities in an issue
The mention feature also come with highlightin, which makes it extremely easy to spot a particular highlight, which can be anything from highlighting the current user from the content or highlighting a set of issues from a query of "Show me all the issues where a particular set of issues are mentioned". Currently the highlighting works only in the `issues/form.tsx`, and in `side-peak-view`.
image

**For Example:** : If the highlight group is for the current user, which means that in every issue, if the current user is mentioned he must be highlighted, in this case it's dynamic because for an issue where `@crew` and `@captain` both are mentioned, if `@crew` is signed in, he must be highlighted and so as captain, which makes us dynamically update the values. Below is the picture of the implementation, which show how dynamic highlights work for two different highlight groups.

**Issue View of the Captain**
image

**Issue View for the Crew** 
image

## Issue filters with Embedded Mentions
Many of the times, customers tend to search for the issues where they are mentioned or maybe a particular issue or a person is mentioned, apps such as slack provide such functionality to quickly grab through text messages, which has been very convenient, hence along with the ability to tag people, I have also added an ability to filter issues according to the mentioned users.

https://github.com/makeplane/plane/assets/72302948/f11c83b7-e330-41ba-b1c0-6a1be1c9be00


apiserver/requirements/base.txt
```diff
@@ -33,4 +33,5 @@ django_celery_beat==2.5.0
 psycopg-binary==3.1.10
 psycopg-c==3.1.10
 scout-apm==2.26.1
-openpyxl==3.1.2
\ No newline at end of file
+openpyxl==3.1.2
+bs4=0.0.1
```
pablohashescobar, line 37: Shouldn't it be `beautifulsoup4==4.12.2`

pablohashescobar (changes requested):

apiserver/plane/bgtasks/notification_task.py
```diff
@@ -5,16 +5,99 @@
...
+
+def get_new_mentions(requested_instance, current_instance):
+    # requested_data is the newer instance of the current issue
+    # current_instance is the older instance of the current issue, saved in the database
+
+    # extract mentions from both the instance of data
+    mentions_older = extract_mentions(current_instance)
+    mentions_newer = extract_mentions(requested_instance)
+
+    # Getting Set Difference from mentions_newer
```
pablohashescobar, line 23: Can we use some list manipulation here instead of a for loop

apiserver/plane/bgtasks/notification_task.py
```diff
@@ -5,16 +5,99 @@
...
+def get_removed_mentions(requested_instance, current_instance):
+    # requested_data is the newer instance of the current issue
+    # current_instance is the older instance of the current issue, saved in the database
+
+    # extract mentions from both the instance of data
+    mentions_older = extract_mentions(current_instance)
+    mentions_newer = extract_mentions(requested_instance)
+
+    # Getting Set Difference from mentions_newer
+    removed_mentions = [
```
pablohashescobar, line 41: Can we use some list manipulation here instead of a for loop

apiserver/plane/db/models/issue.py
```diff
@@ -95,6 +96,13 @@ class Issue(ProjectBaseModel):
         through="IssueAssignee",
         through_fields=("issue", "assignee"),
     )
+    mentions = models.ManyToManyField(
```
pablohashescobar, line 99: Let's not keep a m2m here. Let's only keep the bridge table and user reverse_accessor for filtering

pablohashescobar (changes requested):

pablohashescobar (approved):

PR #2636, opened 2023-11-03
style: custom analytics bar graph label overlapping

1akhanBaheti:
fix: Bar graph avatar image rendering & tooltip

Before:
image

After:
image

image


aaryan610:
Can you add a screenshot or a screen recording of the same?

PR #2681, opened 2023-11-07
chore: remove active ids from the MobX stores if not present in the route

aaryan610:
If any active id is not present is not present in the route, mark it as `null` in the respective store. This will prevent the `autorun` from running unnecessarily.


web/lib/mobx/store-init.tsx
```diff
@@ -58,12 +58,24 @@ const MobxStoreInit = observer(() => {
    */
   useEffect(() => {
     if (workspaceSlug) workspaceStore.setWorkspaceSlug(workspaceSlug.toString());
+
     if (projectId) projectStore.setProjectId(projectId.toString());
+    else projectStore.setProjectId(null);
```
AbhiShake1, line 63:
```suggestion
    projectStore.setProjectId(projectId?.toString() ?? null);
```

web/lib/mobx/store-init.tsx
```diff
@@ -58,12 +58,24 @@ const MobxStoreInit = observer(() => {
...
+
     if (cycleId) cycleStore.setCycleId(cycleId.toString());
+    else cycleStore.setCycleId(null);
```
AbhiShake1, line 66:
```suggestion
    cycleStore.setCycleId(cycleId?.toString() ?? null);
```

AbhiShake1: and repeat for all please. this would reduce the repetition. `||` or `??` whatever the standard here is

AbhiShake1: also, if its possible to use undefined instead of null to do the same, it wouldnt even require `?? null` and make the code alot more elegant

AbhiShake1 (changes requested):

AbhiShake1 (approved):

PR #2756, opened 2023-11-13
fix: file uploads

NarayanBavisetti:


apiserver/plane/api/views/asset.py
```diff
@@ -73,3 +74,16 @@ def delete(self, request, asset_key):
             # Delete the file object
             file_asset.delete()
             return Response(status=status.HTTP_204_NO_CONTENT)
+
+class AssetsEndpoint(BaseAPIView):
+
+    permission_classes = [WorkspaceEntityPermission]
+    parser_classes = (MultiPartParser, FormParser)
+
+    def get(self, request, workspace_id, asset_key):
```
pablohashescobar, line 83: Can we do a `slug` here instead of `workspace_id` ? @NarayanBavisetti

pablohashescobar: Current permission layer works on `slug`

pablohashescobar (changes requested):

apiserver/plane/api/views/asset.py
NarayanBavisetti: ok sure, will change to slug

codevski:
Will this fix the issue how file uploads still direct to localhost?
image
image

PR #2784, opened 2023-11-16
feat: New Pages with Enhanced Document Editor Packages made over Editor Core 📝

henit-chobisa:
## _The New Pages_ with Enhanced Document Editor Package ✈️
image

## Description
Plane used to have a block based document editor, where blocks can be added, reordered and issues could be linked, but for better convenience and user experience we strived to make a page editor similar to notion, which provides a better and more natural user experience to our users. This Pull Request represents the first version of the document editor, and there is a lot coming up in the roadmap. Let me walk you through the awesome new features of plane pages...

### Summary View
We wanted to plan pages, for any of the use cases, for documents of any length, hence we we introduced summary view for your _super smooth_ navigation across the document, enhancing clarity and your productivity across the documents.
This is the first version of summary view, which includes navigation with only headings and subheadings. 

In the next version of plane pages, we strive to take summary view to a next level, including 
- Any type of headings
- Table Support for quick navigation across different types of tables scattered around your page
- Todo and checkmark support for summary view
and many more..............

Here is a glimpse of, how fun and _smooth_ is summary view

ezgif-3-a140bc41f3

### Markdown Support
Markdown is a crucial part of documentation, either it's a readme or creating an issue, it's everywhere, hence we focus a lot of markdown and now we have embedded markdown support for both copying from a page or pasting a markdown into the page, pages will auto format and structure everything according to the markdown format.

Copying Markdown from Pages, let's say to create a Github issue?

ezgif-3-cb056cf1b9

Ohh, you wanna paste some markdown text?? Say you have to paste add your Github change-logs / Contribution guidelines to pages?

ezgif-3-110c36a390

### Slash Commands / Customisation Menu

Similar to the comment editor and issue editor inside plane which we are familiar with, pages comes with slash commands too, which can be triggered by `/` along with that it also comes with a top menu from where you can select any types of text formatting, we tried our best to cover all of our user's needs, whether you're coming from notion or you're coming from confluent like editors.

ezgif-3-76170f9745

### Locking Pages
With pages, we had to ace collaboration, with this first version, we introduce you `page_locking`, say if you make a page and you want that page to be accessible by everyone but not editable, or you want that page to be editable and accessible you can you that. Let me take a below example where we have a user `crew` ( using dark theme ) and `captain` ( using light theme ), if captain makes a page and locks it,  crew looses it's privileges to edit the page.

ezgif-3-bc21658dad

### Shared / Private Pages
Now, there is locking the page, for restricting access to editing to users except the creator.  But what about restricting access to view the page, to solve this problem, we came up with the concept of public and private pages, if you're page is public, that can be viewed by the people who are the part of your project. In future versions, we can also publish the pages, hence anyone with the link can access the page which you have shared.

You can see all the pages which you shared in your shared tab, while all the private pages, which are only visible to you will be seen in the private tab.

ezgif-2-b1fdc706e8

### Favourite / Unfavourite Pages
You can favourite any page to mark it, with the proceeding versions of pages, we will have labels, which will provide a better experience filtering the pages, according to the labels.
 
ezgif-2-b89c42c089

### Duplication of Pages

Often times, we need to make copy of documents. Pages releases this hustle by usage of the duplicate command, which you can use to make a copy of the existing page that you're seeing. Say a page is locked and you have to make a copy of the page, such that you can add your changes, you can simply do it . 
ezgif-2-b063ab4abd

### Archiving / Unarchiving

We never know, which document comes handy at what time, right? Hence, instead of delete, we added archives, if you think you're done with a page, just archive it. If you need it back later you can safely restore the old version or you can have a glimpse at that without restoring it back.
ezgif-2-812d7ba9ef


web/components/pages/pages-list/all-pages-list.tsx
```diff
@@ -1,29 +1,31 @@
...
+export const AllPagesList: FC = observer(() => {
+  // store
+  const {
+    page: { projectPages },
+  } = useMobxStore();
 
-export const AllPagesList: React.FC<TPagesListProps> = ({ viewType }) => {
-  const router = useRouter();
-  const { workspaceSlug, projectId } = router.query;
+  console.log("projectPages", projectPages);
```
rahulramesha, line 16: NIT( Not important though): please remove un wanted console logs

web/components/pages/pages-list/list-item.tsx
```diff
@@ -0,0 +1,278 @@
...
+  };
+
+  const handleRestorePage = (e: any) => {
+    e.preventDefault();
+    e.stopPropagation();
+    restorePage(workspaceSlug, projectId, page.id);
+  };
+
+  const handleEditPageModal = (e: any) => {
+    e.preventDefault();
```
rahulramesha, line 127: This doesn't seem to do any logic for editing page, could you check why this is required?

rahulramesha (commented):
Most of the changes looks good to me... just couple of small comments.

web/components/pages/pages-list/list-item.tsx
sriramveeraghanta: This code is work in progress.

PR #2849, opened 2023-11-23
fix: user state after logout

1akhanBaheti:
### **This PR fixes :**

1. fetch user failed request after logout, conflicting with newly signed user first fetch request.


sriramveeraghanta:
This should be implemted using mobx store. Once the user is logged out you need to make the mobx current user to null.

sriramveeraghanta (approved):

PR #2879, opened 2023-11-25
fix: Sorting of cycles by date in cycle view

TaziSydney123:
This means the cycles will be in chronological order, making it easier to follow. Wasn't sure where to put date-sort, so it's in web/components/cycles.


sriramveeraghanta:
Hello @TaziSydney123 

Instead of sorting manually in JavaScript, we will handle all these sorting-related changes at the API level. So I am closing this pull request.

PR #2933, opened 2023-11-29
chore: api rate limiting

NarayanBavisetti:
- Removed the default rate limit class and added a custom rate limit class in external api's


sriramveeraghanta (approved):

YassineChe:
@NarayanBavisetti, There is anyway to lift the rate limit ?

PR #2952, opened 2023-12-01
Bug 2766 multiple form on same page

MayurMM:
Since email code form is only used in the user-signin form, remove the enter keypress listener should remove explicitly listening the the enter keystroke.

I believe the react-hook-form behavior is such that it tracks the enter event automatically - in order to address it, I created a custom wrapper that should listen to keypress event and ignore it based on flag.

### DIFF
--- a/space/components/accounts/email-code-form.tsx
+++ b/space/components/accounts/email-code-form.tsx
@@ -12,6 +12,7 @@ import useTimer from "hooks/use-timer";
 
 // ui
 import { Input, PrimaryButton } from "components/ui";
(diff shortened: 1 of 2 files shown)


aaryan610:
Hello @MayurMM, thank you for your contributions. However, we have recently updated our sign-in workflow on the web app, making this code outdated. We plan to implement the same workflow on our space app soon.

We appreciate your efforts and look forward to more contributions from you in the future. If you have any questions or if there's anything else you'd like to contribute, feel free to let us know.

PR #3109, opened 2023-12-13
Fix: update self-host docs link in README

ehteshamdev0:
The links to self-hosting docs in README were redirecting to an incomplete path, fixed it.

PR #3115, opened 2023-12-13
fix(README): use correct link for self-hosting documentation

Sayrus:
This link on the [landing page](https://plane.so/) is also returning 404 but I'm not sure where to change or notify about that.
image


Sayrus:
Oops, didn't see #3109 

Closing this one.

PR #3160, opened 2023-12-15
fix - file size limit not work on plane.settings.production

luandnh:
**Problem:**

- FILE_SIZE_LIMIT is limiting the amount of attachments allowed to be uploaded, however the backend is limited at DATA_UPLOAD_MAX_MEMORY_SIZE in django, default is 2621440 (i.e. 2.5 MB).

**Solution:**

- Add DATA_UPLOAD_MAX_MEMORY_SIZE in plane.settings.production with a value equal to FILE_SIZE_LIMIT.


pablohashescobar:
Hi @luandnh, can we move this setting to `common.py` instead of production.py as it is a common setting that can be used on all the three environments.

luanduongtel4vn:
> Hi @luandnh, can we move this setting to `common.py` instead of production.py as it is a common setting that can be used on all the three environments.

I think it's okay, i moved it to `common.py` in new commit. Thank you for the suggestion.

luandnh:
> Hi @luandnh, can we move this setting to `common.py` instead of production.py as it is a common setting that can be used on all the three environments.

I think it's okay, i moved it to `common.py` in new commit. Thank you for the suggestion.

pablohashescobar:
Hello @luandnh, I appreciate the update. It looks good to me.

pablohashescobar (approved):

Issue #3195, opened 2023-12-19
[bug]: Websocket connection to the proxy server failed while set up with docker compose file


abidkhan484:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

There is a 404 error while requesting to the `/_next/webpack-hmr` endpoint with a WebSocket connection. But it perfectly works while requesting an HTTP connection.

![websocket_connection_failed](https://github.com/makeplane/plane/assets/15053047/3c9bbabf-901c-450d-b010-ff72f8b4d5ce)

### Steps to reproduce

Following with the [Local setup steps](https://github.com/makeplane/plane?tab=readme-ov-file#steps), the error occurred. 

### Browser

Google Chrome

### Version

Local

### Possible Solution

Since the proxy server is running with Nginx, it is recommended to add the relevant proxy header to use Web Socket. The relevant documentation is shared below.

Reference: [Nginx WebSocket Proxying documentation](https://nginx.org/en/docs/http/websocket.html)

PR #3196, opened 2023-12-19
fix websocket connection issue in docker makeplane/plane#3195

abidkhan484:
Since the proxy server is running with Nginx, it is recommended to add the relevant proxy header to use Web Socket. The relevant documentation is shared below.

Reference: [Nginx WebSocket Proxying documentation](https://nginx.org/en/docs/http/websocket.html)

### DIFF
--- a/nginx/nginx.conf.template
+++ b/nginx/nginx.conf.template
@@ -18,6 +18,9 @@ http {
 
         location / {
             proxy_pass http://web:3000/;
+            proxy_http_version 1.1;
+            proxy_set_header Upgrade $http_upgrade;
+            proxy_set_header Connection "upgrade";
         }
 
         location /api/ {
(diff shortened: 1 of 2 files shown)


nginx/nginx.conf.template
```diff
@@ -18,6 +18,9 @@ http {
 
         location / {
             proxy_pass http://web:3000/;
+            proxy_http_version 1.1;
```
mguptahub, line 21: Move all 3 lines to nginx.conf.dev file which is used for local development

mguptahub (changes requested):
Please fix and revert.

nginx/nginx.conf.template
abidkhan484: The mentioned lines are removed from the main nginx template and added to the specified local template as per instruction.

mguptahub (approved):
good to go

abidkhan484:
> good to go

Thank you for your thorough review of my pull request. Your insights were incredibly helpful, and I appreciate your time and expertise.

sriramveeraghanta (approved):

Issue #3229, opened 2023-12-22
[bug]: clicking on a parent issue label removes it


olivierlambert:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

On an issue view, if there's a parent, it's visible with the right menu:

![image](https://github.com/makeplane/plane/assets/1241401/1eed28a1-55f6-4a97-9cf0-c46949c77beb)

If I click on the parent "label" (here `VATES-46`), it will be removed, even if I don't click on the small `x`. Which is surprising, since the UI is correct for the "Blocked by" just underneath.

So the parent behavior should be the same than "Blocked by".

### Steps to reproduce

1. Link a parent issue
2. Click on the label
3. It will remove the link to the parent

### Browser

Mozilla Firefox

### Version

Self-hosted

vihar:
@olivierlambert, thank you for flagging this issue. We might have been overlooked during our recent major updates to the issue detail page. We will shortly create a pull request to fix this. Apologies for any inconvenience.

cc: @sriramveeraghanta

olivierlambert:
No worries, happy to report bugs. Keep up the great work!

olivierlambert:
I confirm this is fixed in the latest release :+1:

PR #3237, opened 2023-12-22
Fix clicking on a parent issue label removes it 

Bahaatbb:
This change resolves issue #3229 by moving the closing behavior inside the x instead of the label


sriramveeraghanta:
Hello @Bahaatbb, 

We have already fixed this issue in the #3267 which will be available in the next release. Thank you for your contribution.

PR #3260, opened 2023-12-27
Chore/user role validation

prateekshourya29:
#### Problem
1. Guests and Viewers could access the 'Save View' button, causing errors due to restricted access. 
2. Guests and Viewers could access `Date` and `Label` properties even without access.

#### Solution
1. Removed `Save View` option from the header for Guest/ Viewers.
#### Before Fix
image

#### After Fix
image

2. Disabled `Date` and `Label` properties with proper styling for all views.
https://github.com/makeplane/plane/assets/33979846/461ab417-db26-4ef6-add3-63517e92d8d3

### DIFF
--- a/web/components/issues/issue-layouts/kanban/properties.tsx
+++ b/web/components/issues/issue-layouts/kanban/properties.tsx
@@ -166,7 +166,11 @@ export const KanBanProperties: React.FC<IKanBanProperties> = observer((props) =>
       {/* sub-issues */}
       {displayProperties && displayProperties?.sub_issue_count && !!issue?.sub_issues_count && (
         <Tooltip tooltipHeading="Sub-issues" tooltipContent={`${issue.sub_issues_count}`}>
-          <div className="flex h-5 flex-shrink-0 items-center justify-center gap-2 overflow-hidden rounded border-[0.5px] border-custom-border-300 px-2.5 py-1">
+          <div
+            className={`flex h-5 flex-shrink-0 items-center justify-center gap-2 overflow-hidden rounded border-[0.5px] border-custom-border-300 px-2.5 py-1 ${
+              isReadOnly ? "cursor-not-allowed" : "cursor-pointer"
+            }`}
+          >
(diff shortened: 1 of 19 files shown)


web/components/issues/issue-layouts/kanban/properties.tsx
```diff
@@ -166,7 +166,11 @@ export const KanBanProperties: React.FC<IKanBanProperties> = observer((props) =>
       {/* sub-issues */}
       {displayProperties && displayProperties?.sub_issue_count && !!issue?.sub_issues_count && (
         <Tooltip tooltipHeading="Sub-issues" tooltipContent={`${issue.sub_issues_count}`}>
-          <div className="flex h-5 flex-shrink-0 items-center justify-center gap-2 overflow-hidden rounded border-[0.5px] border-custom-border-300 px-2.5 py-1">
+          <div
+            className={`flex h-5 flex-shrink-0 items-center justify-center gap-2 overflow-hidden rounded border-[0.5px] border-custom-border-300 px-2.5 py-1 ${
+              isReadOnly ? "cursor-not-allowed" : "cursor-pointer"
```
aaryan610, line 171: Why to disable this? It's anyways not clickable

prateekshourya29: Oh okay. In that case we should not even show it as `cursor-pointer` cause then it may seem as if it is clickable?

prateekshourya29: Or we can keep it as we do have a tooltip. Not sure what's the best UX.

aaryan610: Remove the `cursor-pointer`, it should only be added to actionable fields. Do the same everywhere.

prateekshourya29: Okk

aaryan610 (approved):

sriramveeraghanta (approved):

PR #3278, opened 2023-12-29
Fix nginx template for custom s3 endpoints

dbtek:
Custom s3 endpoints can be used via setting environment variables. However proxy is using local minio always.


nginx/nginx.conf.template
```diff
@@ -30,7 +30,7 @@ http {
         }
 
         location /${BUCKET_NAME}/ {
-            proxy_pass http://plane-minio:9000/uploads/;
+            proxy_pass ${AWS_S3_ENDPOINT_URL}/${AWS_S3_BUCKET_NAME};
```
mguptahub, line 33: Please append "/" at the end as

`proxy_pass ${AWS_S3_ENDPOINT_URL}/${AWS_S3_BUCKET_NAME}/;`

mguptahub (changes requested):

nginx/nginx.conf.template
dbtek: Sorry, my bad. Fixed.

mguptahub: Thanks. 

I believe this is not the complete  fix for custom S3 end point. As today if the files were uploaded to S3 (and not minio), the uploaded file url gets into the issue content. That means every time you see that file, it doesn’t go through proxy but directly from client to S3 

if this has to go through proxy, application level code need to be modified as well.

 Let me reconfirm with team and get back to you. @sriramveeraghanta @pablohashescobar 

I will keep this PR on hold till we get clarity from team

dbtek: I confirm this works with external minio and proxies asset urls. Application level modification would be required to use direct s3 urls other than current with proxy.

pablohashescobar: @dbtek, this change requires a significant change in the application we will keep this PR open till then once we have made the application we will merge this. Thank you for your contributions.

poudelprakash: @pablohashescobar is there any way to make external minio url work with docker deployment currently?

dbtek: With this proposed change I perfectly use external minio without any issues.

poudelprakash: @dbtek Got it, but pr seems to be open for long time and this is blocker for anyone trying to use external minio

dbtek: @poudelprakash yeah I know. I built my own proxy image and advise anyone to the same as it seems that pr won't be merged anytime soon.

poudelprakash: @dbtek Thanks to your pr and comments, I got it working now :)

pablohashescobar:
@dbtek, with the new private bucket changes, this PR is no longer required. Thank you for your contributions.

PR #3296, opened 2024-01-01
Fix env substitute issue in websocket docker setup

abidkhan484:


nginx/nginx.conf.dev
abidkhan484, line 1: This line is occuring the below error for `evnsubst` [shared in this line](https://github.com/makeplane/plane/blob/preview/nginx/env.sh#L3).

`invalid number of arguments in "proxy_set_header" directive in /etc/nginx/nginx.conf`

04023401: Thanks, it worked.

mguptahub (approved):

PR #3352, opened 2024-01-11
[feat]: Build Docker Images for ARM

### LINKED ISSUE #3343: [feature]: ARM SUPPORT

### Summary

Hello Folks, 

I had tested in my POC Lab and i love it so far. 

I just like to have ARM support, so that i can self host in my Kubernetes. 
I tried to add ARM support but running in to issues like `replace-env-vars.sh` Not found. 

So, reaching the community and see if ARM support is included. 

Currently most of the Startups are planning to run their Infra on ARM based nodes, Because the cost is much less compared with X86. 

Also, Having multi Arch supported Docker can be helpful for the community so we can deploy in both X86 / Arm.

torbenraab:
This PR resolves #3343 for docker images.

Build the Docker Images for not only AMD64 but also for ARM64 and ARM (V7)

### DIFF
--- a/.github/workflows/build-branch.yml
+++ b/.github/workflows/build-branch.yml
@@ -93,7 +93,7 @@ jobs:
         with:
           context: .
           file: ./web/Dockerfile.web
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
           tags: ${{ env.FRONTEND_TAG }}
           push: true
         env:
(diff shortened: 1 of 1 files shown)


mguptahub:
@torbenraab

We have tried running this PR and run the branch build action. 

- PROXY : Successfully built the docker image
- FRONTEND / SPACE / BACKEND: It took a very long time and got stuck while installing libraries for ARM so  i had to cancel it

We even tried with self hosted ARM and AMD servers, and it was the same case. 

While we work on finding the right solution, considering the need for ARM self hosting, I have created a new PR #3358 as a workaround. 

As said, this is  just for time being till we get github action build sorted

sriramveeraghanta:
Closing this pull request because the provided solution is not functioning properly with GitHub Actions. Until we identify the correct resolution for this issue, we are implementing a workaround solution created by @mguptahub  in Pull Request #3358

PR #3359, opened 2024-01-12
[FIX] Pages Malfunctioning on Load and Recent Pages Computation

henit-chobisa:
- Updated Recent Pages Computation based on `updated_at` instead of `created_at`.
- Updated `usePage` hook, to return `IPageStore` instead of the whole context.

### DIFF
--- a/web/store/page.store.ts
+++ b/web/store/page.store.ts
@@ -196,7 +196,7 @@ export class PageStore implements IPageStore {
         return response;
       });
     } catch (error) {
-      return [];
+      throw error;
     }
   };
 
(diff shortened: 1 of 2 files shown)


web/store/page.store.ts
```diff
@@ -184,15 +184,21 @@ export class PageStore implements IPageStore {
...
+          response.forEach((page) => {
+            set(this.pages, [page.id], page);
+          });
         });
+        return response;
       });
-      return response;
-    });
+    } catch (error) {
+      return [];
```
sriramveeraghanta, line 199: catch block should always throw the error message, to handle it at the component level. By returning an empty array we can never identify if the server value is empty or it is an error case.

sriramveeraghanta (changes requested):

sriramveeraghanta (approved):

PR #3378, opened 2024-01-16
dev: workspace active cycles

anmolsinghbhatia:
This PR includes a feature that enables users to view all active cycles within their workspace conveniently. The feature is accessible through the app sidebar, where a new "Active Cycle" tab has been added. Additionally, search functionality implemented based on project name, project identifier, and cycle name, And Two call-to-action buttons are included for users to visit the active cycle detail page and the active cycle issues.

### Media
c77aecc2-72cd-4088-8ab4-d92e7885be67


sriramveeraghanta (approved):

ccrvlh:
Just noticed this PR - is this going to be added back into the master branch, or is the plan to maintain it under a paywall?

NeuroSan:
Can I ask if it will be possible to disable it if it requires purchase? On Self-Hosted the proposal to "Upgrade" looks a quite strange
screen

NeuroSan:
@sriramveeraghanta Thanks for Your reply. Of course, I appreciate Your explanation regarding the transition of the active cycles feature to the enterprise feature set. While I understand the need to enhance the robustness of the feature, I was hoping for a way to disable button itself on my self-hosted version, as the upgrade prompt seems out of place.

PR #3421, opened 2024-01-22
dev: email notifications 

pablohashescobar:
Email notifications for all issue-related activities  according to user preferences. The current preferences are configured at the user level, with the potential to extend them to workspace and project levels in the future.

The available preference fields include:
- `property_change`: Notifications for changes in issue properties (e.g., name, priority, start date, etc.).
- `state_change`: Notifications for any changes in the issue's state.
- `comment`: Notifications when a new comment is added to the issue.
- `mention`: Notifications when a user is mentioned in the issue.
- `issue_completed`: Notifications when an issue is moved to a completed state.

Email notifications will only be triggered if you have created, are assigned to, or have subscribed to the issue, and if you have set the corresponding condition to true.


apiserver/plane/bgtasks/email_notification_task.py
```diff
@@ -0,0 +1,209 @@
...
+                    ) else None
+
+    return data
+
+
+@shared_task
+def send_email_notification(
+    issue_id, notification_data, receiver_id, email_notification_ids
+):
+    base_api = "http://localhost:3000"
```
NarayanBavisetti, line 122: can you remove this hardcoded value?

sriramveeraghanta (approved):

PR #3474, opened 2024-01-25
feat: One Click Deployment

mguptahub:


sriramveeraghanta (approved):

Julian-Hackenberg:
@sriramveeraghanta will a migration need to be run to update from version 14 to 15?

mguptahub:
@Julian-Hackenberg 

This is a fresh setup and very different from regular docker compose setup. This setup does not need any migration. 

If you are looking at migrating from original docker-compose based setup to this one, it will be tricky and not straight forward. 

In coming days we do plan to come up with upgrade strategy from docker-compose to 1-click.

Hope this answers your query.

NandishRana:
Are there perhaps any potential **drawbacks** or **limitations** for installation using his method in the _long run_?
Referencing Tutor's Open Edx installation; its docker binary release solution, that carries restrictions on integrations with custom plugins.

PR #3489, opened 2024-01-29
chore: update time in real-time in dashboard and profile sidebar

aaryan610:
#### Problem:

Dashboard and profile sidebar time not updating in real-time.

#### Solution:

1. Created a new custom hook, `useCurrentTime`, to get current time and update it in real-time.
2. Replaced the old time logic with the new hook in the dashboard greeting and profile sidebar.

#### Other improvements:

1. Close cycle and module sidebar date-picker after selecting a date.

### DIFF
--- a/web/components/profile/sidebar.tsx
+++ b/web/components/profile/sidebar.tsx
@@ -5,9 +5,10 @@ import { Disclosure, Transition } from "@headlessui/react";
 import { observer } from "mobx-react-lite";
 // hooks
 import { useUser } from "hooks/store";
-import { useCurrentTime } from "hooks/use-current-time";
 // services
 import { UserService } from "services/user.service";
+// components
+import { ProfileSidebarTime } from "./time";
 // ui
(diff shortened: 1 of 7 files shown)


web/components/profile/sidebar.tsx
```diff
@@ -35,7 +38,6 @@ export const ProfileSidebar = observer(() => {
   );
 
   // Create a date object for the current time in the specified timezone
-  const currentTime = new Date();
```
rahulramesha, line 38: Since this is only being used here, 
`<span>
          {timeString} <span className="text-custom-text-200">{userProjectsData?.user_data.user_timezone}</span>
        </span>`

Can you please move this to a different component? or else the entire side bar will be re rendered every second

rahulramesha (changes requested):

sriramveeraghanta (approved):

rahulramesha (approved):

PR #3573, opened 2024-02-06
chore: mention notification and webhook faliure

NarayanBavisetti:
This PR addresses the following bugs:

- It corrects the payload structure for the mentioned user in the comment. Previously, Celery was encountering errors due to the incorrect structure generated for the mentioned user.

- It also resolves a bug where the issue no longer exists. If changes were made to an issue and then the issue was later deleted, Celery would throw an error stating that the issue does not exist. This occurred because email notifications were triggered every 5 minutes.

- When a webhook encounters a failure, it previously persisted in making requests using the retry mechanism. However, with this update, if the requests fail repeatedly for five times, the webhook's state will be automatically changed to 'disabled'. Additionally, an email notification will be sent to the webhook's creator, informing them that the webhook has been disabled.

Screenshot 2024-02-06 at 3 57 20 PM

### DIFF
--- a/apiserver/plane/api/views/base.py
+++ b/apiserver/plane/api/views/base.py
@@ -1,6 +1,8 @@
 # Python imports
 import zoneinfo
 import json
+from urllib.parse import urlparse
+
 
 # Django imports
 from django.conf import settings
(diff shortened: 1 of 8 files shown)


etaxi341:
Hello @NarayanBavisetti
I merged your change to my branch because i needed the mentions to work. With your change the mentions work.
But the e-mail does not contain the comment:

image

could you fix this too? Thank you! :)

NarayanBavisetti:
Thank you, @etaxi341, for bringing this to our attention. While the mention notification addresses the structural issue, it seems there are still some problems with the rendering of the details. We'll work on fixing this as soon as possible.

pablohashescobar (approved):

PR #3577, opened 2024-02-06
fix: show window closing alert only when page is not saved

jlucaso1:


Palanikannan1437:
Thanks so much @jlucaso1 for this PR, really appreciate it 🚀

I've pushed some changes along with your commit to remove the message parameter as modern browsers don't support it anyway and I've made some changes that now show and alert prevent route changes too if the content is unsaved.

cc: @sriramveeraghanta

sriramveeraghanta (approved):

PR #3578, opened 2024-02-07
feat: add and remove points in estimate

jlucaso1:
Estimate point fields are now dynamic.
It is still necessary to review the UI/UX and remove the points in the database that were removed in the edit (I will need help with this).
image

### DIFF
--- a/web/components/estimates/create-update-estimate-modal.tsx
+++ b/web/components/estimates/create-update-estimate-modal.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect } from "react";
 import { useRouter } from "next/router";
-import { Controller, useForm } from "react-hook-form";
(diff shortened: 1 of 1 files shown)


srinivaspendem:
@jlucaso1 we appreciate the contribution, but we have already started working on updating estimates to be dynamic and also subsequently will update the analytics, scope and demand along with cycle calculations. We believe that merging this PR right now is not fruitful as making estimate points dynamic is just a small part of a much larger problem. 

I would suggest you to wait for a few more weeks as what you desire will be available soon in the product with many more powers.

sriramveeraghanta:
@jlucaso1, thanks for your design suggestions. We have already shipped this feature, but unfortunately, we couldn't go with your design suggestions.

Try it out and let us know how you like it.

Issue #3587, opened 2024-02-07
[bug]: New Mobile UI removes Inbox


etaxi341:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Hello,

the PR #3580 by @rameshkumarchandra from 6hours ago made changes to the mobile layout on the issues page.
This looks great!
But you missed the inbox button. On Mobile View there is no inbox anymore.

Could you please have a look at this? Thanks

### Steps to reproduce

1. Go on Dev Branch
2. Enable "Inbox" Feature for Project
3. Open issues
4. Make window small for mobile size
5. You see the Inbox button disappears

### Browser

Google Chrome

### Version

Self-hosted

etaxi341:
Fixed in PR #3591

PR #3590, opened 2024-02-07
fix: module header hide on bigger screens

rameshkumarchandra:
This PR contains the fix for the module issues header persisting on bigger screens.


etaxi341:
Also the Bug #3587 is related to your last commit

pushya-plane (approved):

etaxi341:
I added Inbox back. You can check it in #3591

PR #3637, opened 2024-02-12
Get detailed error responses from apiserver

### LINKED ISSUE #3628: [feature]: Get detailed error responses from apiserver

### Summary

Get the `apiserver` to provide a meaningful error description to users and administrators, who are deploying self-hosted Plane in Docker and do not have development tools at hand. Instead of the useless `{"error": "Something went wrong please try again later"}` provide an exact exception type by default, and provide a `traceback` output in `DEBUG` mode.

This feature should help a lot to spend less time on debugging various issues, such as #3583 and #3618.

Currently, it's impossible to debug the failure in the `apiserver`, because all failures are falling into the same general form of error response:

```python
            if settings.DEBUG:
                print(e)
            capture_exception(e)
            return Response(
                {"error": "Something went wrong please try again later"},
                status=status.HTTP_500_INTERNAL_SERVER_ERROR,
            )
```

Moreover, the `print(e)` does not print anything into `docker-compose` with `DEBUG` present, with ` gunicorn ... --access-logfile - --error-logfile -`.

The entire gunicorn output is sunk somewhere beyond the container's stderr, which is a separate issue.

dmikushin:
@pablohashescobar , this should fix #3628.


pablohashescobar:
@dmikushin, thank you for your contribution. Returning errors to the frontend application even on `DEBUG` mode is a bad approach. We are working on enabling logging for all the containers which will be saved in a file that can be used for finding the errors.

dmikushin:
Yes, this is a poor man solution that allowed me to stop wasting time on fighting with meaningless errors. If you don't mind, will hold it here until there is a better solution.

I think a better solution is two-fold:

1. Figure out what necessary bits of Django logging config are missing. It should look like [this](https://stackoverflow.com/a/76262412/4063520):

```json
LOGGING = {
    'version': 1,
    'disable_existing_loggers': False,
    'handlers': {
        'console': {
            'class': 'logging.StreamHandler',
        },
    },
    'root': {
        'handlers': ['console'],
        'level': 'INFO',
    },
    'loggers': {
        'django': {
            'handlers': ['console'],
            'level': 'INFO',
            'propagate': True,
        },
        'django.request': {
            'handlers': ['console'],
            'level': 'INFO',
            'propagate': False,
        },
    },
}
```

2. Use proper logging functions provided by Django, if `print(e)` is not considered a part of the logging.

pablohashescobar:
@dmikushin, we will be implementing proper logging for the backend apis.

Issue #3369, opened 2024-01-14
[bug]: Docker build fails for v0.14.2-dev: "error next@14.0.3: The engine "node" is incompatible with this module"


dmikushin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Last time I've successfully built plane Docker container image for self-hosting for version `v0.13.1-dev`. Today I tried `v0.14.2-dev`, and it fails. I'm attaching the full log below.

```
((v0.14.2-dev))> docker-compose build
plane-db uses an image, skipping
plane-redis uses an image, skipping
plane-minio uses an image, skipping
Building with native build. Learn about native build in Compose here: https://docs.docker.com/go/compose-native-build/
Building api
Sending build context to Docker daemon  1.953MB

Step 1/24 : FROM python:3.11.1-alpine3.17 AS backend
 ---> ffc28cbd185e
Step 2/24 : ENV PYTHONDONTWRITEBYTECODE 1
 ---> Using cache
 ---> 54970d440cc7
Step 3/24 : ENV PYTHONUNBUFFERED 1
 ---> Using cache
 ---> ce28a6ecf7b9
Step 4/24 : ENV PIP_DISABLE_PIP_VERSION_CHECK=1
 ---> Using cache
 ---> 50d871da1f09
Step 5/24 : WORKDIR /code
 ---> Using cache
 ---> 490dcfa24025
Step 6/24 : RUN apk --no-cache add     "libpq~=15"     "libxslt~=1.1"     "nodejs-current~=19"     "xmlsec~=1.2"
 ---> Using cache
 ---> 378fa3d18e85
Step 7/24 : COPY requirements.txt ./
 ---> Using cache
 ---> 65eb46a0bb62
Step 8/24 : COPY requirements ./requirements
 ---> 907d733966c6
Step 9/24 : RUN apk add --no-cache libffi-dev
 ---> Running in 7f855273515c
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/community/x86_64/APKINDEX.tar.gz
(1/3) Installing linux-headers (5.19.5-r0)
(2/3) Installing pkgconf (1.9.4-r0)
(3/3) Installing libffi-dev (3.4.4-r0)
Executing busybox-1.35.0-r29.trigger
OK: 83 MiB in 56 packages
Removing intermediate container 7f855273515c
[...]

PR #3638, opened 2024-02-12
Please finally update the used version of Node in Docker containers

dmikushin:
`node:18-alpine` is no longer a good version for building Plane docker containers, already since `0.14-dev`. Many packages do require newer version. Therefore, this PR is bumping the used version of Node by changing the version signature of the base container:

```
find . -name "Dockerfile*" -exec sed -i "s/18-alpine/18.19-alpine/g" {} \;
```

Please share the detail explaining how the new prebuilt Plane containers are being successfully built, despite this issue. 

Please stop using any alternative ways of building the Docker containers (if they exist), because this breaks the chain of trust. If we are not able to reproduce the containers using the public Dockerfile, then one can assume the prebuilt images may contain extra features, such as malware.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18-alpine AS builder
+FROM node:18.19-alpine AS builder
 RUN apk add --no-cache libc6-compat
 # Set working directory
 WORKDIR /app
@@ -12,7 +12,7 @@ RUN turbo prune --scope=app --scope=plane-deploy --docker
 CMD tree -I node_modules/
 
 # Add lockfile and package.json's of isolated subworkspace
-FROM node:18-alpine AS installer
+FROM node:18.19-alpine AS installer
 
 RUN apk add --no-cache libc6-compat
 WORKDIR /app
--- a/space/Dockerfile.dev
+++ b/space/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM node:18-alpine
+FROM node:18.19-alpine
 RUN apk add --no-cache libc6-compat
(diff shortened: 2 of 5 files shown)


dmikushin:
This PR fixes #3369

mguptahub:
@dmikushin 

I do not see reason of changing `node:18-alpine` to `node:18.19-alpine` where both are same as on date - except that `node:18.19-alpine` will never upgrade to `node:18.20-alpine` where as `node:18-alpine` does. As of now, they both are aliases to each other till node:18.20 comesout 

image

For me, this PR will not bring any impact on library versions upgrades and vulnerabilities 

CC: @sriramveeraghanta

dmikushin:
@mguptahub , the main reason to change `18-alpine` to `18.19-alpine` is that the Plane Docker containers fail to build with `18-alpine` due to the error mentioned in #3369 :

```
error next@14.0.3: The engine "node" is incompatible with this module. Expected version ">=18.17.0". Got "18.16.0"
error Found incompatible module.
```

As you pointed out, `18-alpine` is a rolling alias to some more specific version. The problem is that we don't know which version `18-alpine` is currently aliased to, and whether this version is new enough for the current requirement of npm packages. Therefore, I believe relying on the alias which version is not known precisely is a wrong choice.

Currently, `18-alpine` is an alias of `18.16-alpine` for me. Is it different for you? If you have the same, could you please explain how you overcome the error above? I thank you for a specific answer!

mguptahub:
Closing this PR as there is no impact with the changes suggested.

PR #3732, opened 2024-02-21
Fix email configuration issue

rouja:
Hi,

We use an internal email gateway that support only ssl, not tls. So we need to be able to configure it. 

This PR add the capability to chose to use TLS or SSL for email configuration.


rouja:
Hi

Is-there any chance that you accept this PR ?

pablohashescobar (approved):

pablohashescobar:
@rouja, The PR looks good to me, we will merge it shortly.

PR #3847, opened 2024-02-29
[CHANG-6] Added Workflow for comparing version between Master & Head before Release

henit-chobisa:
## Description
The Pull Request adds a new workflow, that runs when a pull request is raised to `master` branch, which withstands the below two cases, 
- If the package.json version of the `master` branch and `head` branch is same, this means that the version has not been changed and the job will fail.
- If the package.json version of `master` and `head` are different, this implies that the version has been successfully changed, resulting in the job to pass.

### DIFF
--- a/.github/workflows/check-version.yml
+++ b/.github/workflows/check-version.yml
@@ -1,0 +1,45 @@
+name: Version Change Before Release
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


mguptahub (approved):
good to go

PR #3945, opened 2024-03-12
[WEB-720] chore: reaction tooltip added

anmolsinghbhatia:
#### Improvement:
This PR includes enhancements for issue reactions. Previously, tooltips were absent. I have now added tooltips to the reactions so that users can see who reacted.

#### Media:
| Before | After |
|--------|--------|
| b4d360e1-cd16-47f0-9f48-318f1b98bc58 | 2b9e7d16-d60b-44b0-8c5d-b531c49fa214 |

### DIFF
--- a/web/helpers/issue.helper.ts
+++ b/web/helpers/issue.helper.ts
@@ -185,20 +185,20 @@ export function getChangedIssuefields(formData: Partial<TIssue>, dirtyFields: { 
   return changedFields;
 }
 
-export const formatUserList = (users: string[]): string => {
-  const userCount = users.length;
-  switch (userCount) {
+export const formatTextList = (TextArray: string[]): string => {
+  const count = TextArray.length;
+  switch (count) {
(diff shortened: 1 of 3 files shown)


web/helpers/issue.helper.ts
```diff
@@ -184,3 +184,21 @@ export function getChangedIssuefields(formData: Partial<TIssue>, dirtyFields: {
 
   return changedFields;
 }
+
+export const formatUserList = (users: string[]): string => {
```
rahulramesha, line 188: Minor change can you use generic names like formatTextList and take an argument of TextArray. So that we can reuse this for something like labels, modules etc... in property dropdowns

anmolsinghbhatia: yeah sure

rahulramesha (approved):

sriramveeraghanta (approved):

Issue #3310, opened 2024-01-03
[bug]: Estimate numbers not respecting order


jordan-zimmitti-newmile:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Hi! Firstly thank you for the great work on this web-app, I love it so far and I am excited to use it through its continued development.

### __Summary Of Issue__

Since release `v0.14-dev` the estimate order is no longer respected. In the example video below I set the numbers in the specific order that I want them to show up in and after I hit save it reorders them alphabetically. I believe this is a bug and I am hoping it can be fixed in the next release!

https://github.com/makeplane/plane/assets/148405039/6dbc7a17-4636-4b13-aa32-2874ba6c4f67

### __Expected outcome__
I would like the numbers to stay in the order I put them in like it did in the prior release. So if I create estimates like this:
* 1:1, 2:2, 3:3, 4:5, 5:8, 6:13

After I hit save I'm hoping they stay in that order

### Steps to reproduce

1. Go to the estimates page
2. Create or edit an estimate
3. Save Estimates
4. Edit estimates and watch them get reordered alphabetically

### Browser

Google Chrome

### Version

Cloud

rahulramesha:
Closing this issues, as this was fixed in this PR https://github.com/makeplane/plane/pull/3326. It was fixed in v0.15-dev.

PR #3977, opened 2024-03-15
[bug]: Estimate numbers not respecting order #3310

ialloyd:
This PR fixes the issue of estimates being reordered alphabetically in Plane v0.14-dev. The solution involves storing estimates in an array of objects to preserve the input order. Please review and provide feedback.


rahulramesha:
@ialloyd Thank you for taking your time to try to fix this. But it looks like we have already fixed this in this PR, https://github.com/makeplane/plane/pull/3326 . Apologies as we haven't been able to update the github issue for this sooner.

PR #4052, opened 2024-03-25
[WEB-820] fix: state count in cycle and module

NarayanBavisetti:
Fix:

- This pull request corrects the total count of issues' states within the respective module and cycle.

Issue Link: [WEB-820](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/0df4f62d-f4d4-4dc0-8e3f-fe6b84376a70)

### DIFF
--- a/apiserver/plane/app/views/cycle/base.py
+++ b/apiserver/plane/app/views/cycle/base.py
@@ -2,6 +2,7 @@
 import json
 
 # Django imports
+from django.db import models
 from django.db.models import (
     Func,
     F,
@@ -14,6 +15,8 @@ from django.db.models import (
     When,
     Value,
     CharField,
+    Subquery,
+    IntegerField,
 )
 from django.utils import timezone
 from django.contrib.postgres.aggregates import ArrayAgg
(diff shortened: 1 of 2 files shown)


apiserver/plane/app/views/cycle/base.py
```diff
@@ -71,6 +74,72 @@ def get_queryset(self):
             project_id=self.kwargs.get("project_id"),
             workspace__slug=self.kwargs.get("slug"),
         )
+        cancelled_issues = (
+            CycleIssue.objects.filter(
```
pablohashescobar, line 78: Can we use `Issue` here instead of CycleIssue as Issue already has a manager ?

apiserver/plane/app/views/cycle/base.py
```diff
@@ -2,6 +2,7 @@
 import json
 
 # Django imports
+from django.db import models
```
pablohashescobar, line 5: Run the formatter once before creating a PR.

pablohashescobar (changes requested):

apiserver/plane/app/views/cycle/base.py
NarayanBavisetti: Yes. Changed to issue manager.

pablohashescobar (approved):

PR #4076, opened 2024-03-27
[WEB-838] fix: cycle dropdown fetch

anmolsinghbhatia:
#### Problem:
- When attempting to create a issue in the workspace level, the cycle dropdown is displaying no options.

#### Resolution:
- To resolve this, the "useWorkspaceIssueProperties" function was added in the workspace wrapper to fetch issue properties at the workspace level.

#### Media:
| Before | After |
|--------|--------|
| before | after |


web/layouts/auth-layout/workspace-wrapper.tsx
```diff
@@ -34,6 +35,8 @@ export const WorkspaceAuthWrapper: FC<IWorkspaceAuthWrapper> = observer((props)
   } = useMember();
   const { workspaces } = useWorkspace();
   const { isMobile } = usePlatformOS();
+  //swr hook for fetching issue properties
+  useWorkspaceIssueProperties(workspaceSlug);
```
rahulramesha, line 39: It will fetch all the data when not needed. please dont use this in workspace wrapper. this will make load times longer.

rahulramesha (changes requested):

web/layouts/auth-layout/workspace-wrapper.tsx
anmolsinghbhatia: I've implemented the modifications you recommended.

PR #4170, opened 2024-04-10
feat: api /workspaces/<str:slug>/issues/<str:project__identifier>-<str:issue__identifier>/

MichaelErmer:
This PR adds an API endpoint to find an issue by its sequenceId (XYZ-1234).

This is required to connect plane to any other services, where the tickets are only referenced by their sequenceId.

### DIFF
--- a/apiserver/plane/api/urls/issue.py
+++ b/apiserver/plane/api/urls/issue.py
@@ -6,9 +6,15 @@ from plane.api.views import (
     IssueLinkAPIEndpoint,
     IssueCommentAPIEndpoint,
     IssueActivityAPIEndpoint,
+    WorkspaceIssueAPIEndpoint,
 )
 
 urlpatterns = [
+    path(
+        "workspaces/<str:slug>/issues/<str:project__identifier>-<str:issue__identifier>/",
+        WorkspaceIssueAPIEndpoint.as_view(),
+        name="issue-by-identifier",
+    ),
     path(
         "workspaces/<str:slug>/projects/<uuid:project_id>/issues/",
         IssueAPIEndpoint.as_view(),
--- a/apiserver/plane/api/views/issue.py
+++ b/apiserver/plane/api/views/issue.py
@@ -32,6 +32,7 @@ from plane.api.serializers import (
     LabelSerializer,
 )
 from plane.app.permissions import (
+    WorkspaceEntityPermission,
     ProjectEntityPermission,
     ProjectLitePermission,
     ProjectMemberPermission,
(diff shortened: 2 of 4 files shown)


MichaelErmer:
Also fixed #4171

ryandevz:
Highly awaiting this feature.

MichaelErmer:
> Hi @MichaelErmer, thanks for contributing to Plane.
> 
> Since the url does not contain project_id the project level permissions can't be validated. The issue should not be available to a user who is not part of the project.

The Project is resolved from the Project SLUG which is part of the Ticket Identifier XYZ = project, 1234 = ticket sequence

MichaelErmer:
@pablohashescobar  is there anything you want changed / tested? Id happily add a test if there is a setup to do so.

pablohashescobar:
Hi @MichaelErmer 
> Project is resolved from the Project SLUG which is part of the Ticket Identifier XYZ = project, 1234 = ticket sequence

You are correct but the permission's are not validated that is if the user has access to that project or not. Can you add that layer as well.

MichaelErmer:
@pablohashescobar I modified the permission check to accept project__identifier instead of project_id, it works as expected, I tested it with 2 projects and access to only one.
image
image

apiserver/plane/api/urls/issue.py
```diff
@@ -6,9 +6,15 @@
     IssueLinkAPIEndpoint,
     IssueCommentAPIEndpoint,
     IssueActivityAPIEndpoint,
+    WorkspaceIssueAPIEndpoint,
 )
 
 urlpatterns = [
+    path(
+        "workspaces/<str:slug>/issues/<str:project__identifier>-<str:issue__identifier>",
```
pablohashescobar, line 14: @MichaelErmer can you please add a trailing `/` here.

pablohashescobar (changes requested):

apiserver/plane/api/urls/issue.py
MichaelErmer: done

apiserver/plane/api/views/issue.py
```diff
@@ -51,6 +52,65 @@
...
+            .select_related("workspace")
+            .select_related("state")
+            .select_related("parent")
+            .prefetch_related("assignees")
+            .prefetch_related("labels")
+            .order_by(self.kwargs.get("order_by", "-created_at"))
+        ).distinct()
+
+    def get(self, request, slug, project__identifier=None, issue__identifier=None):
+        if issue__identifier and project__identifier:
```
pablohashescobar, line 96: @MichaelErmer can you also add an `else` case here.

pablohashescobar (changes requested):

apiserver/plane/api/views/issue.py
pablohashescobar: Rest LGTM.

MichaelErmer: the else is covered by django, it will 404, what else should it do?

MichaelErmer:
@pablohashescobar can you have a look again? The else is handled by django, it 404's what else would you like to happen?

pablohashescobar:
> @pablohashescobar can you have a look again? The else is handled by django, it 404's what else would you like to happen?

@MichaelErmer, got it. The PR looks good. We will merge it. Thank you for contributing to Plane.

MichaelErmer:
> > @pablohashescobar can you have a look again? The else is handled by django, it 404's what else would you like to happen?
> 
> 
> 
> @MichaelErmer, got it.  We can explicitly handle that condition returning a JSON response. 
> 
> 
> 
> Also can you please make a small change can you change the URL to `workspaces/<str:slug>/projects/<str:project__identifier>/issues/<str:issue__identifier>/`
> 
> This will make it uniform to all the other APIs.
> 
> 
> 
> Sorry for the repeated pings—just aiming to align everything with our project standards. Really appreciate your patience and effort on this!

It is intentional to have a url that does resolve the project, the project is part of the identifiers XYZ-1234, XYZ resolves the project. 

This is to be used by an integration that only needs the identifier and can then use the API to perform updates/actions on plane. There can be multiple projects and using your approach would then require to make all projects known to the external part and do multiple lookups until one matches...

pablohashescobar:
@MichaelErmer, correct makes sense. Thank you for pointing out.

pablohashescobar (approved):

PR #4198, opened 2024-04-15
fix: code block paste from VSCode and indentation

Palanikannan1437:


packages/editor/core/src/ui/extensions/code/code-block.ts
```diff
@@ -0,0 +1,368 @@
...
+              // Ensure insertPos is within document bounds
+              if (insertPos < 0 || insertPos > tr.doc.content.size) {
+                console.error("Invalid insert position.");
+                return false;
+              }
+
+              // Create a new code block node with the pasted content
+              const textNode = view.state.schema.text(text.replace(/\r\n?/g, "\n"));
+              const codeBlock = this.type.create({ language }, textNode);
+              console.log(insertPos, tr.doc.content.size);
```
rahulramesha, line 312: Please remove unnecessary console logs

packages/editor/core/src/ui/extensions/code/code-block.ts
```diff
@@ -0,0 +1,368 @@
...
+              if (insertPos <= tr.doc.content.size) {
+                tr.insert(insertPos, codeBlock);
+                view.dispatch(tr);
+                return true;
+              }
+
+              return false;
+            } else {
+              // // complicated paste logic, to be handled later
```
rahulramesha, line 321: Please check and remove commented code if not required

packages/editor/core/src/lib/editor-commands.ts
```diff
@@ -34,32 +35,77 @@ export const toggleUnderline = (editor: Editor, range?: Range) => {
...
+      const docSize = editor.state.doc.content.size;
+      if (from < 0 || to > docSize || from > to) {
+        console.error("Invalid range for replacement.");
+        return;
+      }
+
+      const lines = textContent.split(/\r?\n/);
+      const parsedContent = lines.map((line) => `<p>${line.replace(/ /g, "&nbsp;")}</p>`).join("");
+      const div = document.createElement("div");
+      div.innerHTML = parsedContent;
```
rahulramesha, line 55: I think you'll need to Sanitize the pasted block for scripts before adding it to the innerHTML. Otherwise it will be a vulnerability

rahulramesha (changes requested):

packages/editor/core/src/ui/extensions/code/code-block.ts
Palanikannan1437: Done, thanks for the catch 🙌

packages/editor/core/src/ui/extensions/code/code-block.ts
Palanikannan1437: I'll soon get back to this part and build over the commented out logic, it was getting a bit complex to handle as i'll have to hook into the existing behaviours that comes with tiptap's default paste handler and the Markdown plugin's paste handler.

Could this stay here for now, so that I could build over it later or should I remove it? @rahulramesha

Palanikannan1437: Removed it for now 🙌

packages/editor/core/src/lib/editor-commands.ts
Palanikannan1437: Thanks to your suggestions, I've removed the innerHTML method, instead parse it line by line to create paragraph nodes via prosemirror apis using the parsed content seperated by \n

rahulramesha (approved):

PR #4252, opened 2024-04-22
[WEB-1015] fix: kanban layout cycle and module quick add

anmolsinghbhatia:
#### Problem:
- Quick add isn't functioning correctly in the kanban layout when grouped by cycle or module.

#### Resolution:
- Addressed this issue by implementing the required changes in the store.

#### Media:
| Before | After |
|--------|--------|
| 2a6f61d4-cd52-438a-be91-a67ebb34a111 | 39544a49-b542-409a-be0e-60d65dd18c7e |

### DIFF
--- a/web/store/issue/project/issue.store.ts
+++ b/web/store/issue/project/issue.store.ts
@@ -223,6 +223,12 @@ export class ProjectIssues extends IssueHelperStore implements IProjectIssues {
 
       const response = await this.createIssue(workspaceSlug, projectId, data);
 
+      if (data.cycle_id && data.cycle_id !== "")
+        await this.rootStore.cycleIssues.addIssueToCycle(workspaceSlug, projectId, data.cycle_id, [response.id]);
+
+      if (data.module_ids && data.module_ids.length > 0)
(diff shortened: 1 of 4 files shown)


web/store/issue/project/issue.store.ts
```diff
@@ -223,6 +223,12 @@ export class ProjectIssues extends IssueHelperStore implements IProjectIssues {
 
       const response = await this.createIssue(workspaceSlug, projectId, data);
 
+      if (data.cycle_id && data.cycle_id !== "")
```
rahulramesha, line 226: This seems to be only for project issues, but looks like it needs to be done in project views store, draft issues store, archived issue store etc, wherever we allow to be grouped by cycle and module

rahulramesha (changes requested):

web/store/issue/project/issue.store.ts
anmolsinghbhatia: Done! I've implemented the changes wherever necessary.

rahulramesha (approved):

rahulramesha (approved):

PR #4258, opened 2024-04-23
[WEB-1041] fix: preserve initial value on create more issues

anmolsinghbhatia:
#### Problem:
- The initial value is not retained when creating an issue with the "create more" toggle enabled.

#### Resolution:
- Resolve this by adjusting the form reset to include the provided data when the "create more" toggle is active.

#### Media:
| Before | After |
|--------|--------|
| 93911db0-636e-4a50-bb14-3be5a754e412 | d7fbb1fe-07c8-406d-b825-c0841f640522 |

### DIFF
--- a/web/components/issues/issue-modal/form.tsx
+++ b/web/components/issues/issue-modal/form.tsx
@@ -190,6 +190,7 @@ export const IssueFormRoot: FC<IssueFormProps> = observer((props) => {
 
     reset({
       ...defaultValues,
+      ...(isCreateMoreToggleEnabled ? { ...data } : {}),
       project_id: getValues("project_id"),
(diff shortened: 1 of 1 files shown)


web/components/issues/issue-modal/form.tsx
```diff
@@ -190,6 +190,7 @@ export const IssueFormRoot: FC<IssueFormProps> = observer((props) => {
 
     reset({
       ...defaultValues,
+      ...(isCreateMoreToggleEnabled ? { ...data } : {}),
```
rahulramesha, line 193: Isn't it better to not reset at all if isCreateMoreToggleEnabled is true? just what to understand.

rahulramesha (approved):

PR #4305, opened 2024-04-29
[WEB-] chore: added a helper function to check if issue is peeked

aaryan610:
#### Problem:

1. Accessing the `peekIssue` from the store directly caused re-render of each issue when the peek overview changes.

#### Solution:

1. Created a helper function to check whether an issue is peeked or not instead of directly comparing with the store value.

### DIFF
--- a/web/components/gantt-chart/blocks/block.tsx
+++ b/web/components/gantt-chart/blocks/block.tsx
@@ -36,7 +36,7 @@ export const GanttChartBlock: React.FC<Props> = observer((props) => {
   } = props;
   // store hooks
   const { updateActiveBlockId, isBlockActive } = useGanttChart();
-  const { peekIssue } = useIssueDetail();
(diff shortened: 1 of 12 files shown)


web/components/gantt-chart/blocks/block.tsx
```diff
@@ -36,7 +36,7 @@ export const GanttChartBlock: React.FC<Props> = observer((props) => {
   } = props;
   // store hooks
   const { updateActiveBlockId, isBlockActive } = useGanttChart();
-  const { peekIssue } = useIssueDetail();
+  const { isIssuePeeked } = useIssueDetail();
```
rahulramesha, line 39: just one comment on naming conventions, `isIssuePeeked` sounds like a boolean. `getIsIssuePeeked` sounds like a method that returns a boolean. please rename to that.

rahulramesha (approved):

PR #4334, opened 2024-05-02
[WEB-1142] chore: optimistically add issue to cycle/modules

aaryan610:
#### Improvements:

1. Optimistically add issue to cycle/modules.
2. Update the store actions to optimistically update the store data and revert the update on error.

#### Plane issue: [WEB-1142](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/1c5e0966-a140-4cfe-9245-d4698e1c8d4a)

### DIFF
--- a/web/store/issue/cycle/issue.store.ts
+++ b/web/store/issue/cycle/issue.store.ts
@@ -4,12 +4,12 @@ import set from "lodash/set";
 import uniq from "lodash/uniq";
 import update from "lodash/update";
 import { action, observable, makeObservable, computed, runInAction } from "mobx";
-// base class
+// types
+import { TIssue, TSubGroupedIssues, TGroupedIssues, TLoader, TUnGroupedIssues, ViewFlags } from "@plane/types";
 // services
 import { CycleService } from "@/services/cycle.service";
 import { IssueService } from "@/services/issue";
 // types
-import { TIssue, TSubGroupedIssues, TGroupedIssues, TLoader, TUnGroupedIssues, ViewFlags } from "@plane/types";
 import { IssueHelperStore } from "../helpers/issue-helper.store";
 import { IIssueRootStore } from "../root.store";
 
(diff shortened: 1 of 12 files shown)


web/store/issue/cycle/issue.store.ts
```diff
@@ -296,27 +296,49 @@ export class CycleIssues extends IssueHelperStore implements ICycleIssues {
     issueIds: string[],
     fetchAddedIssues = true
   ) => {
+    const issuesBeforeUpdate: Record<string, string | null | undefined> = {};
```
rahulramesha, line 299: can you please rename this to something like `issueCyclesBeforeUpdate`

web/store/issue/module/issue.store.ts
```diff
@@ -301,27 +300,39 @@ export class ModuleIssues extends IssueHelperStore implements IModuleIssues {
     fetchAddedIssues = true
   ) => {
     try {
-      await this.moduleService.addIssuesToModule(workspaceSlug, projectId, moduleId, {
-        issues: issueIds,
-      });
-
-      if (fetchAddedIssues) await this.rootIssueStore.issues.getIssues(workspaceSlug, projectId, issueIds);
-
+      // add the new issue ids to the module issues map
```
rahulramesha, line 303: I feel like addIssuesToModule and removeIssuesFromModule doesn't need optimistic updates. Since it is done with an action. I thing we can await for the API call. We'll only need optimistic updates for issue updates

web/store/issue/module/issue.store.ts
```diff
@@ -359,24 +370,37 @@ export class ModuleIssues extends IssueHelperStore implements IModuleIssues {
...
+      });
+
       return issueToModule;
     } catch (error) {
+      moduleIds.forEach((moduleId) => {
+        runInAction(() => {
+          // remove the new issue ids from the module issues map
+          update(this.issues, moduleId, (moduleIssueIds = []) => pull(moduleIssueIds, issueId));
+          // remove the new module ids from the root issue map
+          update(this.rootStore.issues.issuesMap, [issueId, "module_ids"], (issueModuleIds = []) =>
```
rahulramesha, line 398: just a suggestion, can we maybe move this out of for Each and just just use Set to update the "module_ids" of the issue from the array before the update?

rahulramesha (changes requested):
I think you also need to perform optimistic updates for `removeModulesFromIssue`. Can you check while removing the modules from the issue modules dropdown is optimistic?

web/store/issue/module/issue.store.ts
aaryan610: `removeIssuesFromModule` needs to be optimistic as it is done using quick actions and there is no loading state. Also, I do not see any harm in making `addIssuesToModule` optimistic.

rahulramesha (approved):

PR #4366, opened 2024-05-04
feat: Integrate Outline

SaadBazaz:
This is a sample Pull Request as I explore the possibility to allow users to use an external knowledgebase (in this case, Outline) instead of the default Pages editor in Plane.

## Why?
I've elaborated on this on [Discord](https://discord.com/channels/1031547764020084846/1053269018196320267/1209180125095141396), and have drawn this out on X ([here](https://twitter.com/saadahmedbazaz/status/1756343078376390765)). Here goes another:

I believe open-source benefits better by working together, instead of rebuilding the same thing over and over. At the bare minimum, they should be interoperable with one another.

Plane and Outline are a classic case: Plane focuses on project management, Agile practices, all that. And that's what they're best at. I'd love to see them improve their project management interfaces, make them flexible and basically replace Linear. Pages? Not their strong suite.

On the other hand, knowledgebasing is pretty much the fundamental concept of Outline. What Outline lacks is dynamicity; the ability to show dynamic data which is integrated with its own knowledgebase (elaborated [here](https://github.com/outline/outline/discussions/6507)). So while they focus on making their knowledgebase product strong, why not let Plane take the load of task management?

## How?
I think having a godmode option which allows admins to toggle this feature, and configure Outline properly would work.

Let's say a user turns **on** the Outline integration. What do we expect to happen?

### User stories
- When navigating to Plane -> Projects -> my-project -> Pages, it shows empty by default (of course)
- When the user creates their first page, it is saved in Outline under: Outline -> Plane Collection -> my-project (or a UID) -> <Page Title>
- An iframe is creating in Plane, in which the Outline editor is shown (navigated to the created document). The user can edit there.
- (To make it bidirectional) Whenever the document is updated on Outline, a webhook is sent to Plane. Possible events: page name changes, page deletion, page creation, etcetera.

I need to identify what exactly needs to be done to achieve the above, but my first question is:

Are we even on the right track? Is there a better way to do this?

### DIFF
--- a/web/components/instance/pages-config-form.tsx
+++ b/web/components/instance/pages-config-form.tsx
@@ -1,0 +1,122 @@
+import { FC, useState } from "react";
+import { Controller, useForm } from "react-hook-form";
+import { Eye, EyeOff } from "lucide-react";
+// ui
+import { Button, Input } from "@plane/ui";
+// types
+import { IFormattedInstanceConfiguration } from "@plane/types";
+// hooks
+import { useApplication } from "hooks/store";
+import useToast from "hooks/use-toast";
+
+export interface IInstancePagesConfigForm {
+  config: IFormattedInstanceConfiguration;
+}
+
+export interface PagesConfigFormValues {
+  OUTLINE_API_TOKEN: string;
+}
+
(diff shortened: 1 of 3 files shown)


SaadBazaz:
cc @aaryan610

PR #4373, opened 2024-05-06
[WEB-1134] fix: module link create and delete mutation

anmolsinghbhatia:
#### Changes:
This PR resolves the module sidebar link creation and deletion mutation.

### DIFF
--- a/web/store/module.store.ts
+++ b/web/store/module.store.ts
@@ -372,14 +372,22 @@ export class ModulesStore implements IModuleStore {
    * @param data
    * @returns ILinkDetails
    */
-  createModuleLink = async (workspaceSlug: string, projectId: string, moduleId: string, data: Partial<ILinkDetails>) =>
-    await this.moduleService.createModuleLink(workspaceSlug, projectId, moduleId, data).then((response) => {
+  createModuleLink = async (
+    workspaceSlug: string,
+    projectId: string,
+    moduleId: string,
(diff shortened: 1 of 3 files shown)


web/store/module.store.ts
```diff
@@ -429,6 +430,7 @@ export class ModulesStore implements IModuleStore {
       runInAction(() => {
         set(this.moduleMap, [moduleId, "link_module"], linkModules);
       });
+      this.fetchModuleDetails(workspaceSlug, projectId, moduleId);
```
SatishGandham, line 433: We don't have to refetch as this is just delete. We can remove the deleted record from local state.

Remove it from moduleMap after successful delete.

web/store/module.store.ts
```diff
@@ -377,6 +377,7 @@ export class ModulesStore implements IModuleStore {
       runInAction(() => {
         set(this.moduleMap, [moduleId, "link_module"], [response]);
       });
+      this.fetchModuleDetails(workspaceSlug, projectId, moduleId);
```
SatishGandham, line 380: Similar to delete. Just put the details from response in moduleMap.

SatishGandham (changes requested):

web/store/module.store.ts
anmolsinghbhatia: Done

SatishGandham (approved):

PR #4379, opened 2024-05-06
[WEB-1176] chore: filter dropdown indicator and code refactor

anmolsinghbhatia:
#### Changes:
This PR includes following changes:
- Applied filter indicator added to filter select dropdown.
- Code refactor.

#### Media:
| Before | After |
|--------|--------|
| before | after |

### DIFF
--- a/web/components/cycles/archived-cycles/header.tsx
+++ b/web/components/cycles/archived-cycles/header.tsx
@@ -1,4 +1,5 @@
 import { FC, useCallback, useRef, useState } from "react";
+import isEmpty from "lodash/isEmpty";
 import { observer } from "mobx-react-lite";
 import { useRouter } from "next/router";
 // icons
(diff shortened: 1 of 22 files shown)


web/components/cycles/archived-cycles/header.tsx
```diff
@@ -61,6 +62,13 @@ export const ArchivedCyclesHeader: FC = observer(() => {
     }
   };
 
+  const appliedFilters: TCycleFilters = {};
```
SatishGandham, line 65: Since we are not using applied filters for display purposes and only dependent on the boolean value, can you bailout early as soon as you come across a filter.

```
const hasFilters = objectHasValues(currentProjectArchivedFilters ?? {})

// Put this in a utilities file
objectHasValues = (obj)=>{
	// run the for loop here.
}
```

SatishGandham (changes requested):
Please follow the dry principle, and move the common code to a utility function.

web/helpers/filter.helper.ts
```diff
@@ -26,6 +27,24 @@ export const calculateTotalFilters = (filters: TFilters): number =>
         .reduce((curr, prev) => curr + prev, 0)
     : 0;
 
+/**
```
SatishGandham, line 31: We can use the existing function above using typescript generics. Please create an issue for that and we will track it as a separate change.

SatishGandham (approved):

SatishGandham (approved):

PR #4413, opened 2024-05-09
[WEB-1134] fix: module link mutation issue

rahulramesha:
[WEB-1134](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/8655410a-d3cb-4887-85ac-10ae8b892c77)

This PR fixes the module link mutation issue by referring the link_module property outside the disclosure. This could be the because of how <Disclosure> component is implemented.

### DIFF
--- a/web/components/modules/sidebar.tsx
+++ b/web/components/modules/sidebar.tsx
@@ -644,6 +644,8 @@ export const ModuleDetailsSidebar: React.FC<Props> = observer((props) => {
           </div>
 
           <div className="flex w-full flex-col items-center justify-start gap-2 border-t border-custom-border-200 px-1.5 py-5">
+            {/* Accessing link outside the disclosure as mobx is not  considering the children inside Disclosure as part of the component hence not observing their state change*/}
+            <div className="hidden">{moduleDetails?.link_module?.length}</div>
             <Disclosure>
               {({ open }) => (
                 <div className={`relative  flex  h-full w-full flex-col ${open ? "" : "flex-row"}`}>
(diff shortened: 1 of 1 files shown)


web/components/modules/sidebar.tsx
```diff
@@ -644,6 +644,8 @@ export const ModuleDetailsSidebar: React.FC<Props> = observer((props) => {
           </div>
 
           <div className="flex w-full flex-col items-center justify-start gap-2 border-t border-custom-border-200 px-1.5 py-5">
+            {/* Accessing link outside the disclosure as mobx is not  considering the children inside Disclosure as part of the component hence not observing their state change*/}
```
SatishGandham, line 647: Isn't there a better API in mobx to force observability?

SatishGandham: How about passing it to defaultOpen in Disclosure?
If that is not an option, at least move it to line 650 inside Disclosure.

SatishGandham (changes requested):

web/components/modules/sidebar.tsx
rahulramesha: we add observer to the class to make it observe all the references of the property. But in this situation, not aware of any at this time. I feel the mobx thinks the children inside Disclosure is not inside this component.

One more way to solve this could be that we separate the disclosure panel into a different component and access the store directly from there (not sure if this will work in this situation).

SatishGandham: I thin the problem is the Disclosure is closed to start with and its not rendering the component. So this is not added to the list of observables in the first render. 
See if it makes sense  to reference it in defaultOpen, else lets move that html inside Disclosure.

SatishGandham (approved):

web/components/modules/sidebar.tsx
rahulramesha: That is pretty good actually. Works well in this scenario. Made the changes.

PR #4462, opened 2024-05-15
[WEB-1324] chore: change password page improvement

anmolsinghbhatia:
#### Changes: 
This PR include following changes:
- Render "Password didn't match" only when the input loses focus (blur). Previously, the error message was displayed on each keystroke.
- Display password fields in a column layout. Previously, they were arranged in a row-wrap layout.

#### Media: 
| Before | After |
|--------|--------|
| before | after |


SatishGandham (approved):
@anmolsinghbhatia , we should have a password component and re-use that. Please create an issue for it.

PR #4463, opened 2024-05-15
[WEB-1322] dev: conflict free pages collaboration

aaryan610:
## What's new?

Pages now support a conflict-free data merging along with OFFLINE editing.

### Implementation

Added `yjs` to efficiently manage conflicts in pages. When on a page, new data gets fetched every 10 seconds and if there's anything new, the changes get merged without any conflicts with the local changes, removing the risk of data loss.

Apart from this, along with the auto-save logic, users can now hit `Ctrl/Cmd + S` or click on the `Save changes` button to save their data.

#### Custom provider

Created a custom provider- `CollaborationProvider` to handle updates on the page

```typescript
export interface CompleteCollaboratorProviderConfiguration {
  /**
   * The identifier/name of your document
   */
  name: string;
  /**
   * The actual Y.js document
   */
  document: Y.Doc;
  /**
   * onChange callback
   */
  onChange: (updates: Uint8Array) => void;
}
``` 

1. `name`- Uniquely identifier for each page.
2. `document`- Actual YDoc to read from and write on.
3. `onChange`- Callback to handle updates.

The main function of the provider is to handle updates on the document. All the updates performed in a time window of 2 seconds are merged using `mergeUpdates` from `yjs` and then the `onChange` callback is triggered using this merge updates.

#### IndexedDB for offline support

Added `y-indexeddb` to handle changes made when the client is offline, all of the updates made(online and offline) are stored in the indexedDB and on reconnecting, these changes are applied to the YDoc. These changes are then listened by the `CollaborationProvider` which triggers the `onChange` callback with these updates.

#### Plane issue: [WEB-1322](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/52a7b1be-95de-49f7-a3f0-56259b723005)


web/services/page.service.ts
```diff
@@ -119,4 +119,33 @@ export class PageService extends APIService {
...
+  async updateDescriptionYJS(
+    workspaceSlug: string,
+    projectId: string,
+    pageId: string,
+    data: {
+      description_binary: string;
+      description_html: string;
+    }
+  ): Promise<any> {
+    return this.patch(`/api/workspaces/${workspaceSlug}/projects/${projectId}/pages/${pageId}/description/`, data)
```
SatishGandham, line 145: You are sending binary data as a string, did you verify if this is working as expected and  data is not truncated.

web/services/api.service.ts
```diff
@@ -31,8 +31,11 @@ export abstract class APIService {
     );
   }
 
-  get(url: string, params = {}) {
-    return this.axiosInstance.get(url, params);
+  get(url: string, params = {}, config = {}) {
+    return this.axiosInstance.get(url, {
```
SatishGandham, line 35: Why the extra param here?

web/pages/[workspaceSlug]/projects/[projectId]/pages/[pageId].tsx
```diff
@@ -39,37 +38,23 @@ const PageDetailsPage: NextPageWithLayout = observer(() => {
   // store hooks
   const { createPage, getPageById } = useProjectPages(projectId?.toString() ?? "");
   const pageStore = usePage(pageId?.toString() ?? "");
```
Comment (SatishGandham) [inline web/pages/[workspaceSlug]/projects/[projectId]/pages/[pageId].tsx:40]: Can you rename this to page. This is an instance of page.

web/services/page.service.ts
SatishGandham: I see that you are encoding!!

web/services/api.service.ts
aaryan610: To fetch the binary data, we need to overwrite the default content type to-

```typescript
{
    "Content-Type": "application/octet-stream",
},
``` 

This is not accepted by the current `get` method.

web/pages/[workspaceSlug]/projects/[projectId]/pages/[pageId].tsx
Comment (aaryan610) [reply web/pages/[workspaceSlug]/projects/[projectId]/pages/[pageId].tsx:40]: Renamed `pageStore` to `page`.

web/hooks/use-page-description.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    return () => {
+      clearInterval(intervalId);
+    };
+  }, [handleSaveDescription]);
+
+  // handle ctrl/cmd + S to save the description
+  useEffect(() => {
+    const handleSave = (e: KeyboardEvent) => {
+      const { ctrlKey, metaKey, key } = e;
```
SatishGandham, line 120: Can you reset the 10second interval when the user manually saves.

web/hooks/use-page-description.ts
```diff
@@ -0,0 +1,146 @@
...
+        await mutateDescriptionYJS();
+        setIsDescriptionReady(true);
+      });
+    } else setIsDescriptionReady(true);
+  }, [mutateDescriptionYJS, pageDescription, pageDescriptionYJS, updateDescription]);
+
+  const handleSaveDescription = useCallback(async () => {
+    if (!isContentEditable) return;
+
+    const applyUpdatesAndSave = async (latestDescription: any, updates: Uint8Array) => {
```
SatishGandham, line 71: Can we move applyUpdatesAndSave outside.

web/hooks/use-page-description.ts
```diff
@@ -0,0 +1,146 @@
...
+      if (!workspaceSlug || !projectId || !pageId || !latestDescription) return;
+      // convert description to Uint8Array
+      const descriptionArray = new Uint8Array(latestDescription);
+      // apply the updates to the description
+      const combinedBinaryString = applyUpdates(descriptionArray, updates);
```
SatishGandham, line 76: Shouldn't we fetch the doc and then update?

SatishGandham (changes requested):

web/hooks/use-page-description.ts
```diff
@@ -0,0 +1,146 @@
...
+      // get the latest html content
+      const descriptionHTML = editorRef.current?.getHTML() ?? "<p></p>";
+      // make a request to update the descriptions
+      await updateDescription(combinedBinaryString, descriptionHTML).finally(() => setIsSubmitting("saved"));
+    };
+
+    setIsSubmitting("submitting");
+    // fetch the latest description
+    const latestDescription = await mutateDescriptionYJS();
```
SatishGandham, line 85: You should wrap this in a try catch block to handle the error.

SatishGandham (changes requested):

SatishGandham (approved):

PR #4516, opened 2024-05-19
[WEB-1177] fix: sub-issues count mutation.

prateekshourya29:
#### Problem
The `sub-issues-count` was not updating upon creating a new sub-issue.

#### Solution
The count wasn't updating because it used a static value (`issue.sub-issues-count`) from the issue-detail data without re-fetching it. To fix this, I'm now updating the `sub_issues_count` data after every create/ remove/ delete operation in sub issues store.

#### Media
* Before

[scrnli_5_20_2024_1-37-14 AM.webm](https://github.com/makeplane/plane/assets/33979846/426bf431-7ab7-4356-be0d-bd80839655c0)

* After

[scrnli_5_20_2024_1-36-13 AM.webm](https://github.com/makeplane/plane/assets/33979846/2ec947d3-06d2-4dad-b2e7-d448a28e859f)

This PR is linked to [WEB-1177](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/5a37e74a-4926-4188-9544-b04bbc6a6e84)


SatishGandham (changes requested):
Hi @prateekshourya29 , have you tried updating the parent issues sub_issues_count count property? Wouldn't that be a simpler solution?

prateekshourya29:
> Hi @prateekshourya29 , have you tried updating the parent issues sub_issues_count count property? Wouldn't that be a simpler solution?

Hey @SatishGandham, are you referring to updating it manually? Or we can also use the `sub_issues` store data to update `sub_issue_count` on every create/ remove/ delete action. 

We can also re-fetch the parent issue details but that would cost us an extra API call for just one property update.

SatishGandham (approved):

PR #4530, opened 2024-05-20
[WEB-1184] feat: issue bulk operations

aaryan610:
## Issue Bulk Operations

Users can now perform bulk operations on project level issues. Actions that can be performed-

1. Archive.
2. Delete.
3. Change properties(State, assignees, priority, start date, due date and labels).

#### Functionality:

Bulk operations are available in the List, Spreadsheet and Gantt layouts. There are various ways to select an issue-

1. By clicking on the checkbox which appears while hovering over an issue.
2. By using arrow keys to navigate between issues and using `Shift + ArrowDown/ArrowUp` to select issues.
3. By using `Shift + left mouse click` to select all the issues between the last selected issue and the clicked issue.

#### Media:

Various ways to select issues-

https://github.com/makeplane/plane/assets/65252264/49967e42-3f78-4e1e-8718-31ea2d258996

#### `MultipleSelectGroup` HOC

Created a HOC `<MultipleSelectGroup />` which accepts the following 2 props-

1. `groups`- Array of group ids.
2. `entities`- Object of the following structure-

```typescript
{
  [groupId: string]: string[] // entity ids
}
``` 

This HOC makes use of the `MultipleSelectStore` to manage selected entities state. This HOC can be reused to implement multiple select functionality anywhere.

#### `MultipleSelectGroupAction`

This component takes care of rendering the checkbox for each group.

#### `MultipleSelectEntityAction`

This component takes care of rendering the checkbox and adds extra attributes to the checkbox element to handle DOM manipulation for each entity(issue, in this case), if needed, and handle auto-scroll when using keyboard navigation.

#### `useMultipleSelect` hook

This custom hook contains all the logic to manage selected entities and returns the following two objects-

1. `helpers`- A list of all the helper functions.
2. `snapshot`- A list of current status of the select group.

#### Plane issue: [WEB-1184](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/401246d4-bc2b-493c-8fa7-5e712cbfe5d5)


web/components/core/multiple-select/entity-select-action.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const MultipleSelectEntityAction: React.FC<Props> = (props) => {
+  const { className, disabled = false, groupId, id, selectionHelpers } = props;
+  // derived values
+  const isSelected = selectionHelpers.isEntitySelected(id);
+
+  return (
+    <Checkbox
+      className={cn("!outline-none size-3.5", className)}
+      iconClassName="size-3"
```
SatishGandham, line 24: Why are we changing the icon size here, shouldn't it be constant across the app?

web/components/gantt-chart/blocks/block.tsx
```diff
@@ -80,10 +86,12 @@ export const GanttChartBlock: React.FC<Props> = observer((props) => {
     >
       <div
         className={cn("relative h-full", {
-          "bg-custom-background-80": isBlockActive(block.id),
+          "bg-custom-background-90": isBlockHoveredOn,
```
SatishGandham, line 89: why do we need isBlockHoveredOn here?

web/components/issues/issue-layouts/list/block.tsx
```diff
@@ -119,33 +129,76 @@ export const IssueBlock = observer((props: IssueBlockProps) => {
...
+                    Only issues within the current
+                    <br />
+                    project can be selected.
+                  </>
+                }
+                disabled={issue.project_id === projectId}
+              >
+                <div className="flex-shrink-0 grid place-items-center w-3.5">
+                  <MultipleSelectEntityAction
+                    className={cn(
```
SatishGandham, line 169: Are these classes different for different views? If not, can you move the classes to the component.

web/components/issues/issue-layouts/spreadsheet/spreadsheet-header.tsx
```diff
@@ -1,33 +1,69 @@
...
+              <Checkbox
+                className={cn(
+                  "size-3.5 opacity-0 pointer-events-none group-hover/list-header:opacity-100 group-hover/list-header:pointer-events-auto !outline-none",
+                  {
+                    "opacity-100 pointer-events-auto": !isGroupSelectionEmpty,
+                  }
+                )}
+                iconClassName="size-3"
+                onClick={() => selectionHelpers.handleGroupClick(SPREADSHEET_SELECT_GROUP)}
+                indeterminate={!isGroupSelectionEmpty}
```
SatishGandham, line 61: How do we know when its fully selected?

web/components/issues/issue-layouts/spreadsheet/spreadsheet-view.tsx
```diff
@@ -73,28 +74,46 @@ export const SpreadsheetView: React.FC<Props> = observer((props) => {
...
-        </div>
-      </div>
+      <MultipleSelectGroup
+        containerRef={containerRef}
+        groups={[SPREADSHEET_SELECT_GROUP]}
+        entities={{
+          [SPREADSHEET_SELECT_GROUP]: issueIds,
+        }}
+      >
+        {(helpers, snapshot) => (
```
SatishGandham, line 84: We should profile this. Snapshot changes when you select, right?

SatishGandham (changes requested):

web/components/issues/issue-layouts/spreadsheet/spreadsheet-header.tsx
aaryan610: The group checkbox state is same for partial and complete selections.

web/components/issues/issue-layouts/list/block.tsx
aaryan610: The class names depend on the group class, keeping it inside the component might cause confusion. Any better approach for this?

web/components/issues/issue-layouts/spreadsheet/spreadsheet-header.tsx
SatishGandham: Why? Because there could be more issues on the server?

aaryan610: Updated the indeterminate and checked states.

SatishGandham (approved):

PR #4542, opened 2024-05-21
[WEB-1394] fix: unassigned issues count in analytics page

rahulramesha:
[WEB-1394](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/d7237ad4-33ff-4960-a1b5-528a1f02e9d6)

This PR fixes the Unassigned count in Burndown chart for analytics

### DIFF
--- a/web/components/analytics/scope-and-demand/scope-and-demand.tsx
+++ b/web/components/analytics/scope-and-demand/scope-and-demand.tsx
@@ -45,7 +45,7 @@ export const ScopeAndDemand: React.FC<Props> = (props) => {
 
   // scope data
   const pendingIssues = defaultAnalytics?.pending_issue_user ?? [];
-  const pendingUnAssignedIssues = pendingIssues?.filter((issue) => issue.assignees__id === null);
+  const pendingUnAssignedIssuesUser = pendingIssues?.find((issue) => issue.assignees__id === null);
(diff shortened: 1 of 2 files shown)


web/components/analytics/scope-and-demand/scope-and-demand.tsx
```diff
@@ -45,7 +45,7 @@ export const ScopeAndDemand: React.FC<Props> = (props) => {
 
   // scope data
   const pendingIssues = defaultAnalytics?.pending_issue_user ?? [];
-  const pendingUnAssignedIssues = pendingIssues?.filter((issue) => issue.assignees__id === null);
+  const pendingUnAssignedIssuesUser = pendingIssues?.find((issue) => issue.assignees__id === null);
```
SatishGandham, line 48: Find gives the first element. I don't think that is the behaviour you want here?

SatishGandham (changes requested):

web/components/analytics/scope-and-demand/scope-and-demand.tsx
rahulramesha: As discussed offline, we need the first and only element.

SatishGandham (approved):

PR #4661, opened 2024-05-30
[WEB-1312] fix: trim file name before uploading

Palanikannan1437:
## Description

Currently if the file name is more than 100 words, then the image upload fails in the editor. So now we trim the filename while keeping the extension in place to make it pass.

## Minor Bug fix

The image insertion position and the cursor position after the upload completes is now dependent on node before the current cursor position, i.e. if it has any content or not.

### DIFF
--- a/apiserver/plane/db/models/asset.py
+++ b/apiserver/plane/db/models/asset.py
@@ -12,8 +12,8 @@ from .base import BaseModel
 
 
 def get_upload_path(instance, filename):
+    filename = filename[:60]
     if instance.workspace_id is not None:
-        filename = filename[:60]
         return f"{instance.workspace.id}/{uuid4().hex}-{filename}"
     return f"user-{uuid4().hex}-{filename}"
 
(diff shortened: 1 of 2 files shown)


apiserver/plane/db/models/asset.py
```diff
@@ -12,6 +12,7 @@
 
 
 def get_upload_path(instance, filename):
+    filename = filename[:50]
```
SatishGandham, line 15: Is this filename with extension or without? If with extension, I think we should preserve it.

Palanikannan1437: Yess!  @pablohashescobar, just as a note, I trim the image's filename while preserving the extension over [here](https://github.com/makeplane/plane/blob/c42bb5cca517805c5c55405aec61dad314d859e5/packages/editor/core/src/ui/plugins/image/image-upload-handler.ts#L132) while keeping the whole string length below 100.

SatishGandham (approved):

PR #4690, opened 2024-06-03
chore: Add components required for estimates

SatishGandham:
This PR adds sortable, radio and typography components. These are required for updated estimates.

### DIFF
--- a/packages/ui/src/radio-input/radio-input.tsx
+++ b/packages/ui/src/radio-input/radio-input.tsx
@@ -1,0 +1,72 @@
+import React from "react";
+import { Field, Label, Radio, RadioGroup } from "@headlessui/react";
+import { cn } from "../../helpers";
+// helpers
+
+type RadioInputProps = {
+  label: string | React.ReactNode | undefined;
+  wrapperClassName?: string;
+  fieldClassName?: string;
+  buttonClassName?: string;
+  labelClassName?: string;
+  ariaLabel?: string;
+  options: { label: string | React.ReactNode; value: string; disabled?: boolean }[];
+  vertical?: boolean;
+  selected: string;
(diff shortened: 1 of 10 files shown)


packages/ui/src/radio-input/radio-input.tsx
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  let aria = ariaLabel ? ariaLabel.toLowerCase().replace(" ", "-") : "";
+  if (!aria && typeof inputLabel === "string") {
+    aria = inputLabel.toLowerCase().replace(" ", "-");
+  } else {
+    aria = "radio-input";
+  }
+
+  // return <h1>Hello</h1>;
```
rahulramesha, line 46: we can remove this

packages/ui/src/sortable/draggable.tsx
```diff
@@ -0,0 +1,62 @@
...
+          element: el,
+          onDragStart: () => setDragging(true), // NEW
+          onDrop: () => setDragging(false), // NEW
+          getInitialData: () => data,
+        }),
+        dropTargetForElements({
+          element: el,
+          onDragEnter: (args) => {
+            setIsDraggedOver(true);
+            setClosestEdge(extractClosestEdge(args.self.data));
```
rahulramesha, line 35: This i think will be a conflict. Some item's bottom edge will conflict with other item's top. So wouldn't it be better to add bottom only to the last element?

SatishGandham: Imagine a case where the droptarget is tall say 100px, then it makes sense to handle where your current relative position is.

rahulramesha (approved):

PR #4694, opened 2024-06-04
chore: helm chart added

mguptahub:


Karitham:
Hi, I took the liberty to fork out a considerable amount of your chart. I would rather upstream it than maintain it myself, as it would have a bigger and better community, and more documentation.

I assume you have more constraints than I do (unsure what the `questions.yaml` file does for example). 

Other than that I've standardized the chart, used proper dependencies etc.

Most of the issues I had with your chart (when trying it out) were the ingress setup does not work well when you already have cert-manager/an issuer.

The postgres/minio/redis components weren't scalable at all and didn't have all required options for me. (eg I want password auth)

URLs were in configmap but they might have contained passwords for postgres/redis.

It was missing configuration option to disable signups.

Feel free to reuse any part of what I did.

https://github.com/Karitham/plane-helm

PR #4760, opened 2024-06-11
feat(configure_instance): check mandatory variables before starting

Royalsspirit:
Hello guys,

To add little contexte about it, i run a self hosted plane instance and i had the case when my SECRET_KEY has not been defined which occur a lot of weird behavior such as google auth configuration dropped after a container restart or even not retrieve after a uprade.

This is a PR to check mandatory keys such as SECRET_KEY in order to prevent starting it without them.

### DIFF
--- a/apiserver/plane/license/management/commands/configure_instance.py
+++ b/apiserver/plane/license/management/commands/configure_instance.py
@@ -19,7 +19,7 @@ class Command(BaseCommand):
         from plane.license.utils.instance_value import get_configuration_value
 
         mandatory_keys = [
-            {                           
+            {
                 "key": "SECRET_KEY",
                 "value": os.environ.get("SECRET_KEY", "1"),
                 "category": "CONFIGURATION",
--- a/apiserver/plane/license/management/commands/configure_instance.py
+++ b/apiserver/plane/license/management/commands/configure_instance.py
@@ -8,6 +8,9 @@ from django.core.management.base import BaseCommand
 from plane.license.models import InstanceConfiguration
 
 
+from django.core.exceptions import ObjectDoesNotExist
+
+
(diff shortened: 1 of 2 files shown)


pablohashescobar:
Hi @Royalsspirit, the check seems to be good addition. But we need not save the `SECRET_KEY` in the database. It should be configured through env only. Can you please update the PR to remove the save part.

Royalsspirit:
Hi Nikhil, thank you for your reply. I changed my if statement by trying to retrieve the keys from exporter var only

apiserver/plane/license/management/commands/configure_instance.py
```diff
@@ -15,6 +15,17 @@ def handle(self, *args, **options):
...
+        mandatory_keys = ["SECRET_KEY"]
+
+        for item in mandatory_keys:
+            if not os.environ.get(item):
+                self.stdout.write(
+                    self.style.ERROR(
+                        f"{item} env variable is required."
+                    )
+                )
+                exit(1)
```
pablohashescobar, line 27: Hi @Royalsspirit, rather than using `exit` you can use a `CommandError` something like `raise CommandError(f"{item} env variable is required.")`

pablohashescobar (changes requested):

apiserver/plane/license/management/commands/configure_instance.py
Royalsspirit: Indeed, way way better than what i did. Thank you

pablohashescobar (approved):

pablohashescobar:
@Royalsspirit, thanks for contributing to Plane. I have approved the PR. We will be merging it soon.

PR #4830, opened 2024-06-14
[WEB-1603] fix: load more issues when issues are deleted

rahulramesha:
This PR aims at fixing the scenario where, load more is not working when issues are deleted. This is solved by generating the cursor by using current number of issues rather than utilizing cursor from the server response.

### DIFF
--- a/web/core/store/issue/cycle/issue.store.ts
+++ b/web/core/store/issue/cycle/issue.store.ts
@@ -205,7 +205,7 @@ export class CycleIssues extends BaseIssuesStore implements ICycleIssues {
       // get params from stored pagination options
       const params = this.issueFilterStore?.getFilterParams(
         this.paginationOptions,
-        cursorObject?.nextCursor,
+        this.getNextCursor(groupId, subGroupId),
         groupId,
         subGroupId
       );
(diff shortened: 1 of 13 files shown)


web/core/store/issue/cycle/issue.store.ts
```diff
@@ -205,7 +205,7 @@ export class CycleIssues extends BaseIssuesStore implements ICycleIssues {
       // get params from stored pagination options
       const params = this.issueFilterStore?.getFilterParams(
         this.paginationOptions,
-        cursorObject?.nextCursor,
+        this.getNextCursor(groupId, subGroupId),
```
SatishGandham, line 208: @rahulramesha , shouldn't the check at line 200 also change?

web/core/store/issue/helpers/base-issues.store.ts
```diff
@@ -421,6 +422,24 @@ export abstract class BaseIssuesStore implements IBaseIssuesStore {
...
+   * Gets the next page cursor based on number of issues currently available
+   * @param groupId groupId for the cursor
+   * @param subGroupId subgroupId for cursor
+   * @returns next page cursor or undefined
+   */
+  getNextCursor = (groupId: string | undefined, subGroupId: string | undefined): string | undefined => {
+    const groupedIssues = this.getIssueIds(groupId, subGroupId) ?? [];
+    const currentIssueCount = groupedIssues.length;
+
+    if (!this.paginationOptions) return;
```
SatishGandham, line 435: NIT. Early return should be above.

SatishGandham (approved):

PR #4880, opened 2024-06-19
[WEB-1610] chore: re-implement reload confirmation hook.

prateekshourya29:
### Issue link: [WEB-1610](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/a9b08ddb-fd54-47f4-8c7c-3893259b3c92)

### DIFF
--- a/web/core/hooks/use-reload-confirmation.tsx
+++ b/web/core/hooks/use-reload-confirmation.tsx
@@ -41,15 +41,13 @@ const useReloadConfirmations = (isActive = true) => {
     // handle browser refresh
     window.addEventListener("beforeunload", handleBeforeUnload, true);
     // handle anchor tag click
-    const body = document.querySelector("body");
-    if (!body) return;
-    body.addEventListener("click", handleAnchorClick, true);
-  // TODO: handle back / forward button click
+    window.addEventListener("click", handleAnchorClick, true);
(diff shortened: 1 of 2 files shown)


web/core/hooks/use-reload-confirmation.tsx
```diff
@@ -1,41 +1,57 @@
...
 
-  //   return () => {
-  //     window.removeEventListener("beforeunload", handleBeforeUnload);
-  //     router.events.off("routeChangeStart", handleRouteChangeStart);
-  //   };
-  // }, [handleBeforeUnload, handleRouteChangeStart, router.events]);
+    return () => {
+      // cleanup
+      window.removeEventListener("beforeunload", handleBeforeUnload);
+      body.removeEventListener("click", handleAnchorClick, true);
```
SatishGandham, line 52: Can you try using window instead of body. It will not have difference but, window comes first and its recommended to attach to the closest node, and you can remove the check on line 45.

SatishGandham (changes requested):

web/core/hooks/use-reload-confirmation.tsx
prateekshourya29: Sure!

SatishGandham (approved):

PR #4901, opened 2024-06-21
[WEB-1690] fix: always show the search input if search query is present

aaryan610:
#### Improvements:

In the filters search input field, if the user types a search query, the search filed should not hide.

#### Pages fixed:

1. Projects search.
2. Cycles search.
3. Modules search.
4. Views search.
5. Pages search.

#### Media:

| Before | After |
|--------|--------|
| <video src="https://github.com/makeplane/plane/assets/65252264/8fb42895-1e7c-41c4-bd56-bf63fa034e20"></video> | <video src="https://github.com/makeplane/plane/assets/65252264/17cb0f13-e29e-4bce-aed7-76eacf5d43fb"></video> |

### DIFF
--- a/web/app/[workspaceSlug]/(projects)/projects/(list)/header.tsx
+++ b/web/app/[workspaceSlug]/(projects)/projects/(list)/header.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useCallback, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import { observer } from "mobx-react";
 import { useParams } from "next/navigation";
 import { Search, Briefcase, X, ListFilter } from "lucide-react";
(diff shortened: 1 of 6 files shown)


web/app/[workspaceSlug]/(projects)/projects/(list)/header.tsx
```diff
@@ -82,6 +82,10 @@ export const ProjectsListHeader = observer(() => {
     }
   };
 
+  useEffect(() => {
+    if (searchQuery.trim() !== "" && !isSearchOpen) setIsSearchOpen(true);
```
Comment (SatishGandham) [inline web/app/[workspaceSlug]/(projects)/projects/(list)/header.tsx:86]: We don't need the second check here.

SatishGandham (changes requested):

SatishGandham (approved):

PR #4980, opened 2024-07-01
[WEB-1791] fix: issue delete redirection

anmolsinghbhatia:
#### Changes:
This PR includes a fix for the issue with delete redirection. Previously, deleting an issue from the issue detail page did not redirect to the issue link page. Additionally, deleting an issue was triggering two toast alerts. I have made the necessary changes to ensure it works as intended.

#### Media:
| Before | After |
|--------|--------|
| before | after |

### DIFF
--- a/web/core/components/issues/issue-detail/issue-detail-quick-actions.tsx
+++ b/web/core/components/issues/issue-detail/issue-detail-quick-actions.tsx
@@ -80,6 +80,7 @@ export const IssueDetailQuickActions: FC<Props> = observer((props) => {
     try {
       if (issue?.archived_at) await removeArchivedIssue(workspaceSlug, projectId, issueId);
       else await removeIssue(workspaceSlug, projectId, issueId);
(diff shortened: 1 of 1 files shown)


SatishGandham (approved):

web/core/components/issues/issue-detail/issue-detail-quick-actions.tsx
```diff
@@ -80,11 +80,7 @@ export const IssueDetailQuickActions: FC<Props> = observer((props) => {
     try {
       if (issue?.archived_at) await removeArchivedIssue(workspaceSlug, projectId, issueId);
       else await removeIssue(workspaceSlug, projectId, issueId);
-      setToast({
-        title: "Success!",
-        type: TOAST_TYPE.SUCCESS,
-        message: "Issue deleted successfully",
-      });
+      router.push(`/${workspaceSlug}/projects/${projectId}/issues`);
```
SatishGandham, line 83: We should redirect to the module/cycle if he landed here from there.

PR #5001, opened 2024-07-02
[WEB-1819] dev: collapsible component

anmolsinghbhatia:
#### Changes:
This PR includes an collapsible component.

#### Issue link: [[WEB-1819]](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/ddb2e202-8fc7-4be0-b01b-84348e771c8f)

### DIFF
--- a/packages/ui/src/collapsible/collapsible.tsx
+++ b/packages/ui/src/collapsible/collapsible.tsx
@@ -1,7 +1,7 @@
 import React, { FC, useState, useEffect, useCallback } from "react";
 import { Disclosure, Transition } from "@headlessui/react";
 
-export type TAccordionProps = {
+export type TCollapsibleProps = {
   title: string | React.ReactNode;
   children: React.ReactNode;
   buttonClassName?: string;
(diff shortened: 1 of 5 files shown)


rahulramesha (approved):

SatishGandham (approved):

packages/ui/src/collapsible/collapsible.tsx
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }, [isOpen]);
+
+  // handlers
+  const handleOnClick = useCallback(() => {
+    if (isOpen !== undefined) {
+      if (onToggle) onToggle();
+    } else {
+      setLocalIsOpen((prev) => !prev);
+    }
```
rahulramesha, line 30:
```suggestion
    if (onToggle) onToggle();
    if (isOpen === undefined) {
      setLocalIsOpen((prev) => !prev);
    }
```
I think we can still trigger `onToggle` even when `isOpen` is `undefined`, so that we can have some side effect on Toggle without passing down `isOpen`

@SatishGandham what do you think?

PR #5088, opened 2024-07-09
[WEB-1759] fix: project dropdown action

gakshita:
### In projects dropdown action are not working as expected

Description: When clicking on the projects we get project details(issues, cycles, modules, views etc.) then if we click on the accordion dropdown closes, now if we click on the same project dropdown wont work, we have to click on the accordion only to open the project details.

[[WEB-1759]](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/0ac2736a-a496-41e3-82fa-4a40aa7d9954)


anmolsinghbhatia:
Could you also implement the same redirection behavior when the sidebar is collapsed?

SatishGandham (approved):

PR #5146, opened 2024-07-17
[WEB-1948] fix: priority icons

gakshita:
Summary
Broken priority icons displayed under custom analytics

[[WEB-1948]](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/5e6c7023-6641-4d90-a7e9-2b4b47ee94be)

### DIFF
--- a/web/core/components/analytics/custom-analytics/table.tsx
+++ b/web/core/components/analytics/custom-analytics/table.tsx
@@ -62,7 +62,7 @@ export const AnalyticsTable: React.FC<Props> = ({ analytics, barGraphData, param
           <tr key={`table-row-${index}`} className="divide-x divide-custom-border-200 text-xs text-custom-text-200">
             <td className="px-2.5 py-2">
               <div className="relative flex items-center gap-2 w-full overflow-hidden">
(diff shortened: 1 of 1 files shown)


web/core/components/analytics/custom-analytics/table.tsx
```diff
@@ -62,7 +62,7 @@ export const AnalyticsTable: React.FC<Props> = ({ analytics, barGraphData, param
           <tr key={`table-row-${index}`} className="divide-x divide-custom-border-200 text-xs text-custom-text-200">
             <td className="px-2.5 py-2">
               <div className="relative flex items-center gap-2 w-full overflow-hidden">
-                <div className="flex-shrink-0 h-3 w-3 rounded overflow-hidden">
+                <div className="flex-shrink-0 rounded overflow-hidden">
```
SatishGandham, line 65: This will increase the icon size. That may not be what's expected.
Check with who added the size, and see if we should reduce the icon size instead.

SatishGandham (approved):

PR #5266, opened 2024-07-30
[WEB-2121] fix: project issue creation

gakshita:
**Summary**
On adding a new issue to a project different from the current project, the issue was getting reflected in the wrong project. 

[[WEB-2121]](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/67c3f3b3-5ab4-472e-864d-c45f740223a5)

## Summary by CodeRabbit

- **New Features**
	- Enhanced issue creation functionality with additional parameters for improved flexibility.
	- Control over issue list updates based on the user's current project context.

- **Bug Fixes**
	- Improved logic to prevent unnecessary method calls, optimizing performance and state management.

- **Documentation**
	- Updated method signatures to clarify the functionality and usage of newly added parameters.

### DIFF
--- a/web/core/store/issue/helpers/base-issues.store.ts
+++ b/web/core/store/issue/helpers/base-issues.store.ts
@@ -520,7 +520,7 @@ export abstract class BaseIssuesStore implements IBaseIssuesStore {
       const response = await this.issueService.createIssue(workspaceSlug, projectId, data);
 
       // add Issue to Store
-      shouldUpdateList && this.addIssue(response, shouldUpdateList);
+      this.addIssue(response, shouldUpdateList);
 
       // If shouldUpdateList is true, call fetchParentStats
       shouldUpdateList && this.fetchParentStats(workspaceSlug, projectId);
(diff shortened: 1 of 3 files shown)


SatishGandham:
Isn't this issue coming from adding the issue to the current project store irrespective of project id?
Won't handling it where we add the created issue to the store solve it better?

web/core/store/issue/helpers/base-issues.store.ts
```diff
@@ -520,7 +520,7 @@ export abstract class BaseIssuesStore implements IBaseIssuesStore {
       const response = await this.issueService.createIssue(workspaceSlug, projectId, data);
 
       // add Issue to Store
-      this.addIssue(response, shouldUpdateList);
+      shouldUpdateList && this.addIssue(response, shouldUpdateList);
```
rahulramesha, line 523: This Change is not required.

rahulramesha (commented):
@gakshita, The logic of the changes look fine. But instead of modifying the `createIssue` in `use-issues-actions`, can you please consider writing it in [this](https://github.com/makeplane/plane/blob/6fa45d872392592e23f65ca198d0e37288783908/web/core/store/issue/cycle/issue.store.ts#L254) format inside `issue.store` for project. 
Only reason to recommend it that way is to maintain commonality between methods and classes.

SatishGandham (approved):

PR #5333, opened 2024-08-08
[WEB-2211] fix: input autoComplete

anmolsinghbhatia:
### Changes:
This PR addresses an issue with the input autoComplete. Previously, the password manager extension would trigger a suggestion popover on every input field, which was not the intended behavior. I've made the necessary adjustments to ensure it now functions as expected.

### Reference:
[[WEB-2211]](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/057bb76a-8093-41e7-9389-9550e58913c5)

## Summary by CodeRabbit

- **New Features**
	- Enhanced input field behavior by introducing an `autoComplete` attribute, allowing control over browser autofill suggestions for improved user experience, particularly in form completion scenarios such as sign-in, password reset, and account setup.


packages/ui/src/form-fields/input.tsx
```diff
@@ -31,6 +31,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>((props, ref) => {
         },
         className
       )}
+      autoComplete="off"
```
SatishGandham, line 34: It might be a desired feature in some place. We should probably take it as a prop.

sriramveeraghanta: Agreed. Which will default to `off` right?

SatishGandham (approved):

PR #5456, opened 2024-08-28
[WEB-2316] chore: Render Tooltips and Drop downs in certain places on hover hover to improve rendering performance

rahulramesha:
This PR contains changes to the Tooltips and Issue Property dropdown. The changes make sure that when all `Tooltips` or issue property `Dropdowns` in `Kanban` or `list` is rendered for the first time, it doesn't render it completely to save time on rendering large number of issues. Instead they are rendered once the user hovers over the component.

## Summary by CodeRabbit

- **New Features**
  - Introduced a customizable `ComboDropDown` component for enhanced dropdown functionality.
  - Added `renderByDefault` prop to various dropdown components for improved rendering control.

- **Bug Fixes**
  - Improved tooltip visibility logic for better user interaction.

- **Refactor**
  - Replaced `Combobox` with `ComboDropDown` in multiple dropdown components to enhance performance and maintainability.
  - Encapsulated button rendering logic within a new `comboButton` variable across components for clearer structure.

- **Documentation**
  - Updated component exports to improve accessibility within the UI package.

### DIFF
--- a/packages/ui/src/dropdowns/combo-box.tsx
+++ b/packages/ui/src/dropdowns/combo-box.tsx
@@ -44,7 +44,7 @@ const ComboDropDown = forwardRef((props: Props, ref) => {
 
     element.addEventListener("mouseenter", onHover);
 
-    () => {
+    return () => {
       element?.removeEventListener("mouseenter", onHover);
     };
   }, [dropDownButtonRef, shouldRender]);
--- a/packages/ui/src/dropdowns/combo-box.tsx
+++ b/packages/ui/src/dropdowns/combo-box.tsx
@@ -1,0 +1,73 @@
+import { Combobox } from "@headlessui/react";
+import React, {
+  ElementType,
+  Fragment,
+  KeyboardEventHandler,
+  ReactNode,
+  Ref,
+  forwardRef,
+  useEffect,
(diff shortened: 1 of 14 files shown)


packages/ui/src/dropdowns/combo-box.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const dropDownButtonRef = useRef<HTMLDivElement | null>(null);
+
+  const [shouldRender, setShouldRender] = useState(renderByDefault);
+
+  const onHover = () => {
+    setShouldRender(true);
+  };
+
+  useEffect(() => {
+    const element = dropDownButtonRef.current;
```
SatishGandham, line 41: If should render is true by default, we don't need to add the events.

SatishGandham (changes requested):

packages/ui/src/dropdowns/combo-box.tsx
rahulramesha: If should render is true by default the element will be undefined as `dropDownButtonRef`'s element would not be rendered and the events wont be added.

SatishGandham (approved):

PR #5468, opened 2024-08-29
[WEB-2216] fix: added validation check for white space for create issue modal

Jimmycutie:
### Problem 
When a issue title is a single space and there are properties added and try to create a issue the properties are being reset.

image
image

### Solution
Added a validation check for the modal form for rejecting whitespace character as input.

image
 
### References
[[WEB-2216]](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/5346b59e-6c2a-458c-a219-73c18921d341)

### DIFF
--- a/web/core/components/issues/issue-modal/components/title-input.tsx
+++ b/web/core/components/issues/issue-modal/components/title-input.tsx
@@ -19,13 +19,19 @@ type TIssueTitleInputProps = {
 
 export const IssueTitleInput: React.FC<TIssueTitleInputProps> = observer((props) => {
   const { control, issueTitleRef, errors, handleFormChange } = props;
-
+  const validateWhitespace = (value: string) => {
+    if (value.trim() === "") {
+      return "Title is required";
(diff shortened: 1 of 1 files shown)


SatishGandham:
@Jimmycutie , Why is the form getting reset when there is an error in creating issue?
You should clear data after the creation is successful.

Jimmycutie:
> @Jimmycutie , Why is the form getting reset when there is an error in creating issue? You should clear data after the creation is successful.

Right now when issue modal form is submitted it goes to onSubmit function(which is in issue-modal/base.tsx) shown in the ss below
image

Then in base `handleFormSubmit` function

image

the data is passed to `onCreateIssue` which outputs the error or success toast. after getting a result here the reset function is called inside `form.tsx` which resets the form values. So I implemented the validation so that user can't proceed without proper inputs.

rahulramesha (approved):

PR #5536, opened 2024-09-05
[WEB-2390] fix: Clickable Area for Issue List Layout Item

Jimmycutie:
### Problem Statement
The current issue list layout only allows the title of each item to be clickable.

### Solution
Moved Control function block to cover the whole issue block element instead of the inner title element.

### Screenshots and Media

https://github.com/user-attachments/assets/89c93642-8ada-4815-92ca-8e01786b5b31

 

### References
[[WEB-2390]](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/issues/4e2a70fd-0527-4ec0-837e-57ecfe5bc7e9)

### DIFF
--- a/web/core/components/issues/issue-layouts/list/block.tsx
+++ b/web/core/components/issues/issue-layouts/list/block.tsx
@@ -133,152 +133,125 @@ export const IssueBlock = observer((props: IssueBlockProps) => {
   const keyMinWidth = (projectIdentifier?.length ?? 0) * 7;
 
   return (
-    <Row
-      ref={issueRef}
-      className={cn(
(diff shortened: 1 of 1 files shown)


Jimmycutie:
> @Jimmycutie The code changes look fine. Since you have moved the ControlLink to the parent level, could you please confirm if the following edge cases work fine,
> 
> * The dropdowns open and close just fine in all three components
> * In case of the list layout the drag and drop works fine
> * In case of list layout Opening of peek overview in draft issues
> * In case of list layout selection of issues in EE work fine

Yes, I have checked these edge cases and It's functioning fine.

SatishGandham (approved):

PR #5583, opened 2024-09-11
Add preview image to board

kubilaysalih:
I have made the first image that comes in the issue detail as an image that we can see on the kanban layout. Various style and type checks need to be done.


sriramveeraghanta:
We plan to introduce a different layout, such as a Gallery View, to address this use case. However, this approach is not scalable because it extracts images from the description, which can negatively impact the performance of the list view API.

kubilaysalih:
Yes, I don't want to take images from detail_html either, but your situation at the time allowed for it. I honestly can't imagine a kanban without previews. If you can offer a better solution, I'd appreciate it.

PR #5609, opened 2024-09-13
[WEB-2453] fix: Render on hover only when enabled

rahulramesha:
This PR changes the code to enable render on Hover for both dropdown and tooltip only when enabled and treats `renderByDefault` as the single source of truth for issue property dropdowns


SatishGandham (commented):
What are the cases where we need to render it by default?

rahulramesha:
> What are the cases where we need to render it by default?

@SatishGandham for now the ones that are rendering on hover are the ones in issue layouts. everywhere it is rendering by default. All the cases needs to tested individually and fix indentation issues before enabling it in them

SatishGandham (approved):

PR #5669, opened 2024-09-20
[WEB-2513] fix: page container height

aaryan610:
#### Problem:

The pages container height doesn't extend to the complete screen height. Use has to click on a specific area to focus on the page.

#### Solution:

Added `h-full` class to Increase the page container height to match the screen height.

#### Media:

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/cd2dc30b-b046-4848-a585-97e578c70899"></video> | <video src="https://github.com/user-attachments/assets/50051e8e-a574-4329-8561-36c6f09a26f2"></video> | 

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved height styling for the page editor component, enhancing the overall layout and user experience.

### DIFF
--- a/web/core/components/pages/editor/editor-body.tsx
+++ b/web/core/components/pages/editor/editor-body.tsx
@@ -176,7 +176,7 @@ export const PageEditorBody: React.FC<Props> = observer((props) => {
               }}
               handleEditorReady={handleEditorReady}
               ref={editorRef}
-              containerClassName="p-0 pb-64"
+              containerClassName="h-full p-0 pb-64"
               displayConfig={displayConfig}
               editorClassName="pl-10"
               mentionHandler={{


web/core/components/pages/editor/editor-body.tsx
```diff
@@ -176,7 +176,7 @@ export const PageEditorBody: React.FC<Props> = observer((props) => {
               }}
               handleEditorReady={handleEditorReady}
               ref={editorRef}
-              containerClassName="p-0 pb-64"
+              containerClassName="h-full p-0 pb-64"
```
SatishGandham, line 179: It should be min-h-screen?

SatishGandham (changes requested):

web/core/components/pages/editor/editor-body.tsx
aaryan610: Not needed as it would add an unnecessary scroll to the container since the editor doesn't always occupy the complete screen height.

SatishGandham (approved):

PR #5683, opened 2024-09-23
[WEB-2001] chore: Code refactor for noload changes.

rahulramesha:
This PR separates out additional no load changes, these changes include:
1. Change the getIssues call of cycles and modules to use the same getIssues API of project with cycle and module as filters
2. Add a groupBy to server GroupBy values in constants
3. Change issue detail's logic on loaders while fetching issue detail
4. render 10 issues by default in Kanban per column
5. Fix Kanban height from group virtualization
6. Removed debounce while fetching next issues in kanban API

## Release Notes

- **New Features**
  - Introduced a new enumeration for mapping issue properties to server-side identifiers.
  - Added properties to manage loading states for issue details across various components.
  - Enhanced configurability in the Kanban layout with a new rendering property.
  - Added functions for retrieving and storing values in local storage.

- **Bug Fixes**
  - Improved handling of undefined filter parameters to prevent potential errors.
  
- **Style**
  - Updated styling to enforce a minimum height constraint in the Kanban layout.

### DIFF
--- a/web/core/hooks/use-local-storage.tsx
+++ b/web/core/hooks/use-local-storage.tsx
@@ -1,6 +1,6 @@
 import { useState, useEffect, useCallback } from "react";
 
-const getValueFromLocalStorage = (key: string, defaultValue: any) => {
+export const getValueFromLocalStorage = (key: string, defaultValue: any) => {
   if (typeof window === undefined || typeof window === "undefined") return defaultValue;
   try {
     const item = window.localStorage.getItem(key);
@@ -8,6 +8,16 @@ const getValueFromLocalStorage = (key: string, defaultValue: any) => {
   } catch (error) {
     window.localStorage.removeItem(key);
     return defaultValue;
+  }
+};
+
+export const setValueIntoLocalStorage = (key: string, value: any) => {
+  if (typeof window === undefined || typeof window === "undefined") return false;
+  try {
+    window.localStorage.setItem(key, JSON.stringify(value));
+    return true;
+  } catch (error) {
+    return false;
   }
 };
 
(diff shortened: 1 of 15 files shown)


SatishGandham (approved):

web/core/store/issue/cycle/issue.store.ts
```diff
@@ -185,7 +185,7 @@ export class CycleIssues extends BaseIssuesStore implements ICycleIssues {
       // get params from pagination options
       const params = this.issueFilterStore?.getFilterParams(options, cycleId, undefined, undefined, undefined);
       // call the fetch issues API with the params
-      const response = await this.cycleService.getCycleIssues(workspaceSlug, projectId, cycleId, params, {
+      const response = await this.issueService.getIssues(workspaceSlug, projectId, params, {
```
SatishGandham, line 188: You should add the cycle id to the params.

SatishGandham (changes requested):

web/core/store/issue/cycle/issue.store.ts
rahulramesha: That is being handled in the `const params = this.issueFilterStore?.getFilterParams` at line 186

web/core/hooks/use-local-storage.tsx
```diff
@@ -11,6 +11,16 @@ const getValueFromLocalStorage = (key: string, defaultValue: any) => {
   }
 };
 
+export const setValueIntoLocalStorage = (key: string, value: any) => {
+  if (typeof window === undefined || typeof window === "undefined") return false;
```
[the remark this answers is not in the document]
rahulramesha: It was just a check that was copied over from the previous method getValueFromLocalStorage. The second check should be good. Left the first one there as i was not sure, couldn't find references on the first check. Should i change it to only the second condition?

PR #5689, opened 2024-09-24
[WEB-2534] fix: Add requestIdleCallback polyfill to fix crash in Safari browser

rahulramesha:
This PR adds a polyfill for `requestIdleCallback` to fix the crash in Safari browser because of the non support of the function on the browser

## Summary by CodeRabbit

- **New Features**
	- Introduced polyfills for `requestIdleCallback` and `cancelIdleCallback` functions to enhance performance during idle periods.
	- Added a consolidated export for easier access to polyfills related to idle callback functionality.

- **Chores**
	- Updated import statements to include the new polyfills in the application.

### DIFF
--- a/web/app/provider.tsx
+++ b/web/app/provider.tsx
@@ -12,6 +12,8 @@ import { SWR_CONFIG } from "@/constants/swr-config";
 import { resolveGeneralTheme } from "@/helpers/theme.helper";
 // nprogress
 import { AppProgressBar } from "@/lib/n-progress";
+// polyfills
+import "@/lib/polyfills";
 // mobx store provider
 import { StoreProvider } from "@/lib/store-context";
 // wrappers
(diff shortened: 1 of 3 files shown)


SatishGandham (approved):

pushya22 (approved):

thomassaphir:
I still reproduced the Safari crash on self-hosted `makeplane/plane-frontend:v1.3.0`.

Console error was:

`TypeError: window.requestIdleCallback is not a function`

The failing stack pointed to `gantt-layout-loader-*.js`.

I inspected the built assets in the container and found that some bundles already include a fallback, but `gantt-layout-loader` still calls `window.requestIdleCallback(...)` directly.

As a self-hosted workaround, I fixed it by shipping a global `requestIdleCallback` / `cancelIdleCallback` polyfill and loading it in `index.html` before the app bundles.

So the PR exists in repo history, but the issue was still observable in the Docker frontend image I was serving (`v1.3.0`).

PR #5761, opened 2024-10-07
[WEB-2603] fix: remove validation of roles from the live server

Palanikannan1437:
## Description 

If a guest created a page, they aren't able to sync their page edits in realtime. Hence now, we've removed the check of membership from the live server and we're letting the app subscribing to our server handle it now.

## Summary by CodeRabbit

- **Bug Fixes**
	- Removed unnecessary logic related to fetching user project membership for project pages, streamlining the authentication process.
	- Adjusted error message formatting for incomplete query parameters.
- **Chores**
	- Removed methods for fetching user workspace and project membership, simplifying the user service.
	- Updated the authentication process by eliminating unnecessary parameters in the function signature.


rahulramesha (approved):

live/src/core/lib/authentication.ts
```diff
@@ -18,11 +14,7 @@ type Props = {
 };
 
 export const handleAuthentication = async (props: Props) => {
```
aaryan610, line 16: Update the props.

Palanikannan1437: Done, thanks!

aaryan610 (approved):

PR #5910, opened 2024-10-24
chore: core/editor updates to support mobile update

1akhanBaheti:


Palanikannan1437 (approved):

packages/editor/src/ce/extensions/document-extensions.tsx
```diff
@@ -7,14 +7,16 @@ import { TIssueEmbedConfig } from "@/plane-editor/types";
 import { TExtensions, TUserDetails } from "@/types";
 
 type Props = {
+  slashCommandEnabled?: boolean;
```
aaryan610, line 10: We already have a prop to handle disabled extensions, `disabledExtensions`, make use of that instead of creating a new prop.

packages/editor/src/core/types/editor.ts
```diff
@@ -78,20 +81,28 @@ export interface IEditorProps {
...
   onEnterKeyPress?: (e?: any) => void;
   placeholder?: string | ((isFocused: boolean, value: string) => string);
   tabIndex?: number;
   value?: string | null;
 }
-
-export type ILiteTextEditor = IEditorProps;
-
+export interface ILiteTextEditor extends IEditorProps {
+  isEnterExtensionEnabled?: boolean;
```
aaryan610, line 93: We already have a prop to handle disabled extensions, disabledExtensions, make use of that instead of creating a new prop.

PR #6158, opened 2024-12-05
fix: remove unwanted states fetching logic to avoid multiple API calls.

prateekshourya29:


SatishGandham (approved):
Other drop-downs also need similar fixes.

PR #6325, opened 2025-01-06
[WEB-2915] fix: Modules reordering through drag and drop

rahulramesha:
### Description

This PR makes change to use `getBlockById` from the module timeline store instead of the props being passed down to fix the re ordering of modules in timeline chart

### Type of Change
- [ ] Code refactoring
- [ ] Documentation update

## Summary by CodeRabbit

- **New Features**
	- Enhanced Gantt chart sidebar with improved state management
	- Integrated new timeline chart hook for more dynamic data retrieval

- **Refactor**
	- Updated component to use MobX observer for reactive state updates
	- Simplified component props and data fetching mechanism

### DIFF
--- a/web/core/components/gantt-chart/sidebar/modules/sidebar.tsx
+++ b/web/core/components/gantt-chart/sidebar/modules/sidebar.tsx
@@ -1,9 +1,14 @@
 "use client";
 
+import { observer } from "mobx-react";
 // ui
 import { Loader } from "@plane/ui";
 // components
-import { ChartDataType, IBlockUpdateData, IGanttBlock } from "@/components/gantt-chart";
+import { IBlockUpdateData } from "@/components/gantt-chart";
+// hooks
+import { useTimeLineChart } from "@/hooks/use-timeline-chart";
+//
+import { ETimeLineTypeType } from "../../contexts";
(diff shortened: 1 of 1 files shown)


web/core/components/gantt-chart/sidebar/modules/sidebar.tsx
```diff
@@ -12,13 +17,14 @@ import { ModulesSidebarBlock } from "./block";
 type Props = {
   title: string;
   blockUpdateHandler: (block: any, payload: IBlockUpdateData) => void;
-  getBlockById: (id: string, currentViewData?: ChartDataType | undefined) => IGanttBlock;
```
prateekshourya29, line 15: We removed a required prop here. Are we not using this component anywhere? @rahulramesha

rahulramesha: Nope, only used in one place but props for that spread through the parents. The build and linting passed on my local. So it shouldn't be an issue

prateekshourya29: Got it. Thanks for the clarification!

prateekshourya29 (approved):

PR #6415, opened 2025-01-17
chore: ln support for cycles constant file

gakshita:
Summary

This PR moved the Cycles constants file to package and resolved title based on language selected

### DIFF
--- a/packages/constants/src/cycle.ts
+++ b/packages/constants/src/cycle.ts
@@ -1,20 +1,4 @@
 // types
-import { TCycleLayoutOptions, TCycleTabOptions } from "@plane/types";
-
-export const CYCLE_TABS_LIST: {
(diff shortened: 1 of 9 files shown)


packages/constants/src/cycle.ts
```diff
@@ -24,33 +8,33 @@ export const CYCLE_STATUS: {
   bgColor: string;
 }[] = [
   {
-    label: "day left",
+    label: "project_cycles.status.days_left",
```
prateekshourya29, line 11: We will have to change it to `labelTranslationKey` and `titleTranslationKey` respectively.

prateekshourya29 (changes requested):

prateekshourya29 (approved):

PR #6480, opened 2025-01-28
refactor: page constants

mathalav55:
### Description

This update moves pages constants to packages.

### Type of Change

- [x] Code refactoring
- [ ] Documentation update


packages/constants/src/page.ts
```diff
@@ -0,0 +1,14 @@
+export enum EPageAccess {
```
aaryan610, line 1: This enum can be moved to the `types` package

prateekshourya29: We are keeping all the `enums` in constants package.

prateekshourya29 (approved):

PR #6778, opened 2025-03-19
chore: Live server restructuring and error handling

Palanikannan1437:
### Type of Change

- [ ] Code refactoring
- [ ] Documentation update


live/src/core/extensions/redis.ts
```diff
@@ -0,0 +1,41 @@
...
+      })
+    );
+    logger.info("HocusPocus Redis extension configured ✅");
+  } else {
+    logger.warn(
+      "Redis connection failed, continuing without Redis extension (you won't be able to sync data between multiple plane live servers)"
+    );
+  }
+
+  return extensions;
```
Prashant-Surya, line 31: Do we need to return an array here? Can we just return extension or null and handle this at caller end?

live/src/lib/controller.utils.ts
```diff
@@ -0,0 +1,32 @@
...
+import { registerControllers as registerRestControllers, registerWebSocketControllers } from "@plane/decorators";
+import "reflect-metadata";
+
+/**
+ * Register all controllers from the controllers array
+ * @param router Express router to register routes on
+ * @param controllers Array of controller classes to register
+ * @param dependencies Array of dependencies to pass to controllers
+ */
+export function registerControllers(router: Router, controllers: any[], dependencies: any[] = []): void {
```
Prashant-Surya, line 11: Can we use the existing function from decorators packages?

Palanikannan1437: Hey @Prashant-Surya, we're indeed using those only, but stitching the web socket and rest controllers exposed from there together there that's all

Prashant-Surya (approved):

PR #6804, opened 2025-03-24
fix: settings dynamic pages permissions

gakshita:
### Description
This PR fixes the permission for settings page dynamic routes

### DIFF
--- a/web/app/[workspaceSlug]/(projects)/settings/(with-sidebar)/layout.tsx
+++ b/web/app/[workspaceSlug]/(projects)/settings/(with-sidebar)/layout.tsx
@@ -24,7 +24,7 @@ const WorkspaceSettingLayout: FC<IWorkspaceSettingLayout> = observer((props) => 
 
   const { workspaceUserInfo } = useUserPermissions();
   const pathname = usePathname();
-  const { workspaceSlug } = useParams();
+  const [workspaceSlug, suffix, route, _] = pathname.slice(1, -1).split("/");
 
   // derived values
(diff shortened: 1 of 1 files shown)


web/app/[workspaceSlug]/(projects)/settings/(with-sidebar)/layout.tsx
```diff
@@ -24,15 +24,15 @@ const WorkspaceSettingLayout: FC<IWorkspaceSettingLayout> = observer((props) =>
...
+  const [workspaceSlug, suffix, route, _] = pathname.slice(1, -1).split("/");
 
   // derived values
   const userWorkspaceRole = workspaceUserInfo?.[workspaceSlug.toString()]?.role;
   const isAuthorized =
     pathname &&
     workspaceSlug &&
     userWorkspaceRole &&
-    WORKSPACE_SETTINGS_ACCESS[pathname.replace(`/${workspaceSlug}`, "").slice(0, -1)]?.includes(
+    WORKSPACE_SETTINGS_ACCESS[route ? `/${suffix}/${route}` : `/${suffix}`]?.includes(
```
Comment (prateekshourya29) [inline web/app/[workspaceSlug]/(projects)/settings/(with-sidebar)/layout.tsx:35]: This logic doesn’t seem scalable as it only considers the first two parts of the URL, excluding `workspaceSlug`. What if there are more segments? This approach will break in such cases.  

For example, will it work for a URL like `${workspaceSlug}/settings/api-token/create`?

prateekshourya29 (approved):

sriramveeraghanta (approved):

PR #6839, opened 2025-03-27
[WEB-3673] fix: password change form

gakshita:
### Description
- Current Password input field removed from the Change password screen if the password is not set.
- Added missing translations for "new_password_must_be_different_from_old_password" key

### References
[[WEB-3673]](https://app.plane.so/plane/browse/WEB-3673/)

## Summary by CodeRabbit

- **New Features**
	- Enhanced the password change flow to display the old password field only when necessary and provide tailored error feedback for missing or incorrect entries.
	- Updated internationalization with new localized messages across multiple languages, clearly informing users that their new password must differ from the old one.


web/app/profile/security/page.tsx
```diff
@@ -40,10 +42,14 @@ const defaultShowPassword = {
 };
 
 const SecurityPage = observer(() => {
+  // store
+  const { data: currentUser } = useUser();
   // states
   const [showPassword, setShowPassword] = useState(defaultShowPassword);
   const [isPasswordInputFocused, setIsPasswordInputFocused] = useState(false);
   const [isRetryPasswordInputFocused, setIsRetryPasswordInputFocused] = useState(false);
+  const [oldPasswordRequired, setOldPasswordRequired] = useState(!currentUser?.is_password_autoset);
```
prateekshourya29, line 51: Why do we need to use a state over here? Let's just directly update `currentUser?.is_password_autoset` once the password is added?

prateekshourya29 (changes requested):

PR #6887, opened 2025-04-08
[WEB-3711] fix: relations delete issue

gakshita:
### Description
- Updated the relations mapping on deleting a relation from issue

### References
[[WEB-3711]](https://app.plane.so/plane/browse/WEB-3711/)

## Summary by CodeRabbit

- **New Features**
  - Enhanced issue relation management now allows for more intuitive removal of relationships between issues, streamlining your workflow.
  - Updates ensure smoother and more responsive handling of issue dependencies during removal operations. 
  - Introduced functionality to update local state during relation removal without requiring an API call.

### DIFF
--- a/web/core/components/issues/issue-detail-widgets/relations/content.tsx
+++ b/web/core/components/issues/issue-detail-widgets/relations/content.tsx
@@ -167,17 +167,17 @@ export const RelationsCollapsibleContent: FC<Props> = observer((props) => {
           data={issueCrudState?.delete?.issue as TIssue}
           onSubmit={async () => {
             if (
-              issueCrudState.removeRelation.issue &&
-              issueCrudState.removeRelation.issue.project_id &&
+              issueCrudState.removeRelation.issueId &&
+              issueCrudState.removeRelation.issue?.project_id &&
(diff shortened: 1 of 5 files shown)


web/core/components/issues/issue-detail-widgets/relations/content.tsx
```diff
@@ -150,6 +166,21 @@ export const RelationsCollapsibleContent: FC<Props> = observer((props) => {
...
+            if (
+              issueCrudState.removeRelation.issue &&
+              issueCrudState.removeRelation.issue.project_id &&
+              issueCrudState.removeRelation.relationKey &&
+              issueCrudState.removeRelation.relationIssueId
+            ) {
+              await removeRelation(
+                workspaceSlug,
+                issueCrudState.removeRelation.issue.project_id,
+                issueCrudState.removeRelation.issueId as string,
```
prateekshourya29, line 178: Do we really need the `as string` here? We're already checking the type in the if condition, so the type should be narrowed automatically and compatible after that check, right? Am I missing something?

prateekshourya29 (approved):

PR #6941, opened 2025-04-17
[WEB-3918] fix: api tokens is_active

sangeethailango:
### Description
This PR will fix the issue where `is_active` always return true for api-tokens api.

## Summary by CodeRabbit

- **New Features**
  - API token information now displays whether a token is currently active.
- **Enhancements**
  - Expiration details for API tokens now show both the date and time, improving clarity across token creation, details, and list views.
- **Bug Fixes**
  - Improved handling and formatting of expiration dates and times for better accuracy and consistency in the user interface.


apiserver/plane/app/serializers/api.py
```diff
@@ -17,10 +19,15 @@ class Meta:
...
 class APITokenReadSerializer(BaseSerializer):
+    is_active = serializers.SerializerMethodField()
+
     class Meta:
         model = APIToken
         exclude = ("token",)
 
+    def get_is_active(self, obj):
+        return timezone.now() < obj.expired_at
+
```
NarayanBavisetti, line 30: `is_active` is already a field in the APIToken model, but it's currently not being used. Since the expiration timestamp is already available, the frontend can determine whether the token is expired or not, so we don't need to add this extra check on the backend. Additionally, `expired_at` is set to None when the token is configured to never expire, which would cause a TypeError if we try to compare it with the current time.

dheeru0198 (approved):

NarayanBavisetti (approved):

pablohashescobar (approved):

PR #6977, opened 2025-04-25
[WEB-3957] chore: IntakeIssues with iexact 'in-app' changed to 'IN_APP'

sangeethailango:
### Description
This PR will change all the intake issues with iexact in-app to IN_APP and will create a new field called start_of_the_week on the profile model. 

## Summary by CodeRabbit

- **New Features**
  - Added a setting for users to specify their preferred start day of the week in their profile.
- **Bug Fixes**
  - Standardized the value of the "source" field for issue records to ensure consistency.


apiserver/plane/db/models/user.py
```diff
@@ -164,6 +164,16 @@ def save(self, *args, **kwargs):
 
 
 class Profile(TimeAuditModel):
+    START_OF_THE_WEEK_CHOICES = (
```
dheeru0198, line 167: @sangeethailango Define separate variables for each day with the integer values

pablohashescobar (approved):

PR #6998, opened 2025-04-30
[WEB-3985] feat: common postcss config and local fonts across all plane applications

prateekshourya29:
### Description

This PR ensures that all Plane application uses common postcss configs and local fonts. 

### Type of Change

- [ ] Code refactoring
- [ ] Documentation update


admin/styles/globals.css
```diff
@@ -1,5 +1,4 @@
-@import url("https://fonts.googleapis.com/css2?family=Inter:wght@200;300;400;500;600;700;800&display=swap");
-@import url("https://fonts.googleapis.com/css2?family=Material+Symbols+Rounded:opsz,wght,FILL,GRAD@48,400,0,0&display=swap");
+@import "@plane/propel/styles";
```
sriramveeraghanta, line 1: This should be `@plane/propel/styles/fonts` Try to export fonts css as a separate one.

packages/propel/src/styles/index.css
```diff
@@ -0,0 +1 @@
+@import url("./fonts/main.css");
```
sriramveeraghanta, line 1: This step is not required. Let other apps import the fonts css directly.

space/styles/globals.css
```diff
@@ -1,5 +1,4 @@
-@import url("https://fonts.googleapis.com/css2?family=Inter:wght@200;300;400;500;600;700;800&display=swap");
-@import url("https://fonts.googleapis.com/css2?family=Material+Symbols+Rounded:opsz,wght,FILL,GRAD@48,400,0,0&display=swap");
+@import "@plane/propel/styles";
```
sriramveeraghanta, line 1: This should be @plane/propel/styles/fonts Try to export fonts css as a separate one.

packages/propel/package.json
```diff
@@ -9,7 +9,8 @@
   },
   "exports": {
     "./ui/*": "./src/ui/*.tsx",
-    "./charts/*": "./src/charts/*/index.ts"
+    "./charts/*": "./src/charts/*/index.ts",
+    "./styles": "./src/styles/index.css"
```
sriramveeraghanta, line 13: Add new export for exporting only fonts, instead of `index.css`

sriramveeraghanta (changes requested):

packages/propel/src/styles/index.css
prateekshourya29: I did it this way so we can move all shared CSS files used across our apps. Let me know if you still want me to change it.

prateekshourya29: Got it 👍🏻

sriramveeraghanta (approved):

PR #7007, opened 2025-05-01
[WEB-3996] fix: attachment icon rendering and added support for rar and zip icons

anmolsinghbhatia:
### Description
This PR addresses the following issues:
- Fixed an issue where the attachment icon was not rendering correctly. 
- Added support for archive icons, including ZIP and RAR formats.

### Type of Change
- [x] Bug fix 
- [x] Improvement

### Media
| Before | After |
|--------|--------|
| before | after |

### Test Scenarios 
- Uploaded .zip and .rar files to verify that the correct attachment icons are displayed.
- Confirmed that icons for previously supported file types are rendering correctly.

### References
[[WEB-3996]](https://app.plane.so/plane/projects/02c3e1d5-d7e2-401d-a773-45ecba45d745/inbox/?currentTab=open&inboxIssueId=4e7ed309-8017-4e85-a46f-6a2d60d1bea1)

## Summary by CodeRabbit

- **New Features**
  - Added support for displaying icons for ZIP and RAR file attachments.
- **Bug Fixes**
  - Improved file extension detection for attachments by using the file name instead of the asset URL.
- **Enhancements**
  - Expanded supported archive MIME types to include additional compressed archive formats, including RAR.


web/core/components/icons/attachment/zip-file-icon.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import Image from "next/image";
+// image
+import ZipFileIcon from "@/public/attachment/zip-icon.png";
```
gakshita, line 4: why not use SVGs?

gakshita (changes requested):

web/core/components/icons/attachment/zip-file-icon.tsx
anmolsinghbhatia: PNG asset was provided by the designer, so I followed the same for consistency. That said, if there's an existing SVG icon set you'd prefer us to use, I'm happy to replace the PNG with an SVG version.

gakshita (approved):

PR #7213, opened 2025-06-13
Analytics refactor

JayashTripathy:
### Type of Change

- [x] Code refactoring


packages/i18n/src/locales/ru/translations.json
JayashTripathy, line 1: Added this translation which got missed in last PR

PR #7239, opened 2025-06-19
[WEB-4354] fix: moved fetch-keys constants to packages

gakshita:
### Description
Moved fetch-keys constants to packages

## Summary by CodeRabbit

- **Chores**
	- Consolidated imports of various constants across the app to use a centralized shared module, improving maintainability and consistency.
	- Updated import paths for constants in multiple components and pages without affecting any user-facing functionality or app behavior.


sriramveeraghanta:
Delete the old fetch keys file and replace all the instances where we are using it.

PR #7256, opened 2025-06-23
Chore: live server restructuring using esbuild

Palanikannan1437:
### Type of Change

- [ ] Code refactoring
- [ ] Documentation update


live/src/core/shutdown-manager.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  /**
+   * Gracefully shuts down the registered HTTP server.
+   */
+  async shutdown(message?: string, exitCode = 0): Promise<void> {
+    logger.error(`Initiating graceful shutdown ${`${message ? `with message: ${message}` : ""}`}`);
+
+    if (this.httpServer) {
+      this.httpServer.closeAllConnections?.();
```
aheckmann, line 65:
```suggestion
```
`closeAllConnections()` will [forcefully close](https://nodejs.org/api/http.html#servercloseallconnections) open connections. Instead, rely on `server.close()` which stops the server from accepting new connections and closes all connections which are not sending a request or waiting for a response. 

We also need a shutdown timeout configuration (10 seconds or so) after which we can forcefully terminate the process. This gives open connections a chance to finish up.

live/src/core/shutdown-manager.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    process.on("uncaughtException", (error) => {
+      logger.error("Uncaught exception:", error);
+
+      handleError(error, {
+        errorType: "internal",
+        component: "process",
+        operation: "uncaughtException",
+        extraContext: { source: "uncaughtException" },
+      });
```
aheckmann, line 34:
```suggestion
      });
      
      gracefulTermination();
```

live/src/core/shutdown-manager.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    process.on("unhandledRejection", (reason) => {
+      logger.error("Unhandled rejection:", reason);
+      handleError(reason, {
+        errorType: "internal",
+        component: "process",
+        operation: "unhandledRejection",
+        extraContext: { source: "unhandledRejection" },
+      });
```
aheckmann, line 44:
```suggestion
      });
      
      gracefulTermination();
```

live/src/core/shutdown-manager.ts
```diff
@@ -0,0 +1,74 @@
...
+  async shutdown(message?: string, exitCode = 0): Promise<void> {
+    logger.error(`Initiating graceful shutdown ${`${message ? `with message: ${message}` : ""}`}`);
+
+    if (this.httpServer) {
+      this.httpServer.closeAllConnections?.();
+      await new Promise<void>((resolve) => {
+        this.httpServer?.close(() => resolve());
+      });
+    }
+    process.exit(exitCode);
```
aheckmann, line 70: Exit should be called after our shutdown timeout configuration is met.

live/src/core/shutdown-manager.ts
```diff
@@ -0,0 +1,74 @@
...
+    return async () => {
+      logger.info("Signal received, shutting down HTTP server");
+      await this.shutdown();
+    };
+  }
+
+  /**
+   * Gracefully shuts down the registered HTTP server.
+   */
+  async shutdown(message?: string, exitCode = 0): Promise<void> {
```
aheckmann, line 61: `shutdown` should only be permitted to execute once.

PR #7302, opened 2025-07-01
chore: member tracker events

gakshita:
### Type of Change

- [ ] Code refactoring
- [ ] Documentation update


packages/constants/src/event-tracker/core.ts
```diff
@@ -196,6 +196,17 @@ export const MEMBER_TRACKER_EVENTS = {
   },
 };
 
+export const MEMBER_TRACKER_ELEMENTS = {
+  HEADER_BUTTON: "header_add_member_button",
```
prateekshourya29: This should be HEADER_ADD_BUTTON, right?

packages/constants/src/event-tracker/core.ts
```diff
@@ -196,6 +196,17 @@ export const MEMBER_TRACKER_EVENTS = {
...
+  INVITATION_BUTTON: "accept_invitation_button",
```
prateekshourya29: ACCEPT_INVITATION_BUTTON

PR #7334, opened 2025-07-03
feat: optimize docker builds for live server

sriramveeraghanta:
### Description

- Optimized the docker build for live server.

### Type of Change

### DIFF
--- a/apps/live/Dockerfile.live
+++ b/apps/live/Dockerfile.live
@@ -2,6 +2,9 @@ FROM node:20-alpine AS base
 # The web Dockerfile is copy-pasted into our main docs at /docs/handbook/deploying-with-docker.
 # Make sure you update this Dockerfile, the Dockerfile in the web workspace and copy that over to Dockerfile in the docs.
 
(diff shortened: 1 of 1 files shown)


apps/live/Dockerfile.live
```diff
@@ -28,16 +34,22 @@ RUN yarn install
...
+# *****************************************************************************
+
 FROM base AS runner
 WORKDIR /app
 
-COPY --from=installer /app .
-# COPY --from=installer /app/live/node_modules ./node_modules
+COPY --from=installer /app/packages ./packages
+COPY --from=installer /app/apps/live/dist ./live
+COPY --from=installer /app/apps/live/package.json ./package.json
```
Prashant-Surya: Should it be like below?
```
COPY --from=installer /app/apps/live/package.json ./live/package.json
```

Prashant-Surya: Also once confirm if the start command in helm chart is`node live/start.cjs`

Prashant-Surya (approved):

PR #7356, opened 2025-07-08
fix(partial): get dev env working

aheckmann:
part of #7355

### Description
Dev environment doesn't start up when running `yarn dev`. This patch fixes that step.

Note: There is a ton of TS errors and other noise in the console after fixing this. Needs :eyes: from the rest of the team.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Feature (non-breaking change which adds functionality)
- [ ] Improvement (change that would cause existing functionality to not work as expected)
- [ ] Code refactoring
- [ ] Performance improvements
- [ ] Documentation update

## Summary by CodeRabbit

* **Chores**
  * Simplified development scripts across multiple apps by removing redundant "develop" scripts and updating the "dev" scripts to directly start the development server on the appropriate port. This streamlines the process for starting the development environment.

PR #7358, opened 2025-07-08
fix: dev server issues

Palanikannan1437:
### Description
Added back dependencies for dev and develop scripts in turbo.json

### Type of Change

## Summary by CodeRabbit

* **Chores**
  * Updated development tasks to ensure they run after the build step.
  * Added a new "develop" task with improved configuration for development workflows.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -25,8 +25,14 @@
       "outputs": [".next/**", "dist/**"]
     },
     "dev": {
+      "dependsOn": ["^build"],
       "cache": false,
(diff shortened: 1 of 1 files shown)


sriramveeraghanta:
These are no longer needed because of [this](https://github.com/makeplane/plane/pull/7356#issuecomment-3047803163).

sriramveeraghanta:
We might need the build step as part of the dev command. We don't need develop pipeline here. I am making those changes.

PR #7398, opened 2025-07-12
[WIKI-544] chore: smoother insert handles for tables

aaryan610:
### Description

This PR improves the table insert handles by making the creation/removal flow smoother.

### Type of Change

## Summary by CodeRabbit

* **Refactor**
  * Improved the drag-to-insert/remove columns and rows feature in the table editor for smoother and more intuitive resizing during drag actions. The table now adjusts directly to match your drag distance, simplifying the experience and ensuring more predictable results.


packages/editor/src/core/extensions/table/plugins/insert-handlers/utils.ts
Palanikannan1437, line 1: can  we also add touch event support so that this works in mobile as well? touchstart, touchmove,etc

PR #7411, opened 2025-07-15
[WIKI-547] fix: update find suggestion logic for emoji extension

VipinDevelops:
### Description
Previously the  logic looked at tiny pieces of text (one node at a time) which was breaking text on different decorator like bold and italic the updated logic looks at the whole text block as one piece now no matter how the text is decorated (bold, italic, etc.), the emoji suggestion will always find the colon and work correctly.

### Type of Change
- [x]  Bug fix (non-breaking change which fixes an issue)

https://github.com/user-attachments/assets/0dec7afc-240c-46c0-b1b4-ad5d0a3e8c60

https://github.com/user-attachments/assets/f3d0aef7-9519-4997-b3d2-e8ce6fa17b04

## Summary by CodeRabbit

* **New Features**
  * Emoji suggestions now use improved matching logic for more accurate trigger detection.
  * The emoji list in the suggestion dropdown only appears when there is a non-empty query.

* **Bug Fixes**
  * Prevented the emoji list from displaying when the query is empty.


packages/editor/src/core/extensions/emoji/find-suggestion-match.ts
```diff
@@ -0,0 +1,88 @@
+import { escapeForRegEx, Range } from "@tiptap/core";
+import { ResolvedPos } from "@tiptap/pm/model";
+
+export interface Trigger {
```
aaryan610: Import this type from the extension.

iam-vipin: sure thing

packages/editor/src/core/extensions/emoji/components/emojis-list.tsx
```diff
@@ -142,52 +143,54 @@ export const EmojiList = forwardRef<EmojiListRef, EmojiListProps>((props, ref) =
...
-                )}
-                onClick={() => selectItem(index)}
-                onMouseEnter={() => setSelectedIndex(index)}
-              >
-                <span className="size-5 grid place-items-center flex-shrink-0 text-base">
-                  {item.fallbackImage ? (
-                    <img src={item.fallbackImage} alt={item.name} className="size-4 object-contain" />
-                  ) : (
-                    item.emoji
+    query.length > 0 && (
```
aaryan610: This logic makes the JSX part complex. Simply return `null` when there's no query.

iam-vipin: sure

packages/editor/src/core/extensions/emoji/find-suggestion-match.ts
aaryan610, line 1: Accept `regex` as an argument and define this function in a common file.

aaryan610 (approved):

PR #7415, opened 2025-07-15
[WEB-2576] fix: inactive members as issue assignees

sangeethailango:
### Type of Change

- [ ] Code refactoring
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filtering of assignees to include only active members of the current project across issue, cycle, sub-issue, and relation views.
  * Project details now correctly reflect the active status of the project lead and default assignee.

* **Performance**
  * Optimized project detail retrieval by prefetching related project members for faster access.

### DIFF
--- a/apps/api/plane/app/views/project/base.py
+++ b/apps/api/plane/app/views/project/base.py
@@ -219,7 +219,20 @@ class ProjectViewSet(BaseViewSet):
             )
             .filter(archived_at__isnull=True)
             .filter(pk=pk)
+            .prefetch_related("project_projectmember")
         ).first()
+
+        project_member_ids = ProjectMember.objects.filter(
+            project_id=pk, is_active=True
+        ).values_list("member_id", flat=True)
+
+        members_ids = set(project_member_ids)
+
+        if project.project_lead_id not in members_ids:
+            project.project_lead = None
+
+        if project.default_assignee_id not in members_ids:
+            project.default_assignee = None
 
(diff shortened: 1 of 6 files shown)


apps/api/plane/app/views/project/base.py
```diff
@@ -219,8 +219,21 @@ def retrieve(self, request, slug, pk):
...
+
+        members_ids = set(project_member_ids)
+
+        if project.project_lead_id not in members_ids:
+            project.project_lead = None
+
+        if project.default_assignee_id not in members_ids:
+            project.default_assignee = None
+
         if project is None:
```
dheeru0198: @sangeethailango this condition should be checked first, as the above code will fail if the project doesn't exist

apps/api/plane/utils/grouper.py
```diff
@@ -42,7 +44,10 @@ def issue_queryset_grouper(
     annotations_map: Dict[str, Tuple[str, Q]] = {
         "assignee_ids": (
             "assignees__id",
-            ~Q(assignees__id__isnull=True) & Q(issue_assignee__deleted_at__isnull=True),
+            ~Q(assignees__id__isnull=True)
+            & Q(issue_assignee__deleted_at__isnull=True)
```
dheeru0198, line 48: @sangeethailango Why is this filter checking different field than the others?

PR #7455, opened 2025-07-22
chore: misc editor updates and utility functions

1akhanBaheti:
### Type of Change

- [x] Code refactoring


packages/editor/src/core/hooks/use-editor.ts
```diff
@@ -145,7 +180,15 @@ export const useEditor = (props: TEditorHookProps) => {
           console.warn(`No command found for item: ${itemKey}`);
         }
       },
+      focus: ({ position = "start", scrollIntoView = false }) => editor?.commands.focus(position, { scrollIntoView }),
+      getCordsFromPos: (pos?: number) => editor?.view.coordsAtPos(pos ?? editor.state.selection.from),
       getCurrentCursorPosition: () => editor?.state.selection.from,
+      getSelectedNodeAttributes: (attribute) => {
```
Palanikannan1437: please rename this according to your use case of extending mark in your selection and getting it's attributes

packages/editor/src/core/hooks/use-editor.ts
Palanikannan1437, line 1: move all these ref methods into getEditorRefHelpers

PR #7483, opened 2025-07-26
chore: refactor live server into a scalable structure

sriramveeraghanta:
### Type of Change

- [ ] Code refactoring
- [ ] Documentation update

## Summary by CodeRabbit

- New Features
  - Real-time collaboration via WebSocket at {BASE_PATH}/collaboration (default /live/collaboration).
  - Document conversion endpoint to transform HTML descriptions to multiple formats.
  - Health check endpoint exposing status, timestamp, and version at {BASE_PATH}/health.

- Improvements
  - More reliable startup/shutdown flow and enhanced logging.
  - Optional Redis integration to improve real-time performance and resilience.

- Chores
  - Updated container startup command to use the new runtime entry point.


apps/live/src/server.ts
```diff
@@ -23,17 +26,33 @@ export class Server {
...
     this.setupRoutes();
   }
 
+  public async initialize(): Promise<void> {
+    return redisManager
+      .initialize()
+      .then(() => {
+        logger.info("Redis setup completed");
+        const manager = HocusPocusServerManager.getInstance();
+        manager.initialize().catch((error) => {
```
aheckmann: I recommend using `await` everywhere in the `server.initialize()` method to make reasoning about and handling errors easier on us mortals. If you want to keep the Promise based approach, then we'll need a `return` to prevent an unhandled rejection.

```suggestion
        return manager.initialize().catch((error) => {
```

apps/live/src/server.ts
```diff
@@ -97,35 +109,23 @@ export class Server {
...
+
+    // Disconnect Redis
+    await redisManager.disconnect();
+    logger.info("Redis connection closed gracefully.");
+
     // Close the Express server
     this.serverInstance.close(() => {
-      manualLogger.info("Express server closed gracefully.");
+      logger.info("Express server closed gracefully.");
       process.exit(1);
```
aheckmann: I recommend leaving the decision to exit the process to the creator of the server (`start.ts`) b/c it has additional context (e.g., maybe the start script wanted to recreate the server)

```suggestion
```

apps/live/src/controllers/health.controller.ts
```diff
@@ -0,0 +1,14 @@
...
+
+@Controller("/health")
+export class HealthController {
+  @Get("/")
+  async healthCheck(_req: Request, res: Response) {
+    res.status(200).json({
+      status: "OK",
+      timestamp: new Date().toISOString(),
+      version: process.env.APP_VERSION || "1.0.0",
+    });
```
aheckmann, line 12: If this is health check is intended to be used by process managers to determine when to shutdown and provision replacement servers, then consider making this health check reflect the server's overall ability to perform it's intended functionality - e.g. it should check that it can reach outside dependencies like redis or anything else it depends on to serve requests. If connections to those services are down, this service is unhealthy and should not receive incoming traffic.

apps/live/src/hocuspocus.ts
```diff
@@ -0,0 +1,200 @@
...
+// lib
+import { handleAuthentication } from "@/core/lib/authentication";
+// extensions
+import { DocumentCollaborativeEvents, TDocumentEventsServer } from "@plane/editor/lib";
+import { fetchPageDescriptionBinary, updatePageDescription } from "@/core/lib/page";
+// plane live libraries
+import { fetchDocument } from "@/plane-live/lib/fetch-document.js";
+import { updateDocument } from "@/plane-live/lib/update-document.js";
+// editor types
+import { TUserDetails } from "@plane/editor";
```
aheckmann:
```suggestion
import type { TUserDetails } from "@plane/editor";
```

apps/live/src/hocuspocus.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  private onDatabaseStore = async ({ context, state, documentName: pageId, requestParameters }: any) => {
+    const cookie = (context as HocusPocusServerContext).cookie;
+    // query params
+    const params = requestParameters;
+    const documentType = params.get("documentType")?.toString() as TDocumentTypes | undefined;
+
+    // TODO: Fix this lint error.
+    // eslint-disable-next-line no-async-promise-executor
+    return new Promise(async () => {
```
aheckmann: Does this logic need to execute later or is it fine to `awaitPageDescription()` etc immediately in this tick?

apps/live/src/redis.ts
```diff
@@ -0,0 +1,210 @@
+import Redis from "ioredis";
```
aheckmann, line 1: The ioredis docs recommend using https://github.com/redis/node-redis, do you want to switch in this PR or leave for a separate PR?

apps/live/src/redis.ts
```diff
@@ -0,0 +1,210 @@
...
+  public getClient(): Redis | null {
+    if (!this.redisClient || !this.isConnected) {
+      logger.warn("Redis client not available or not connected");
+      return null;
+    }
+    return this.redisClient;
+  }
+
+  public isClientConnected(): boolean {
+    return this.isConnected && this.redisClient !== null;
```
aheckmann, line 120: Can we use `this.redisClient.status === 'ready'` here? Relying on the driver to determine if we are connected should provide a more accurate answer. https://github.com/redis/ioredis/blob/bedcfb5/lib/Redis.ts#L43

sriramveeraghanta:
@coderabbitai review

apps/live/src/hocuspocus.ts
```diff
@@ -0,0 +1,186 @@
...
+        new Logger({
+          onChange: false,
+          log: (message) => {
+            logger.info(message);
+          },
+        }),
+        new Database({
+          fetch: this.onDatabaseFetch,
+          store: this.onDatabaseStore,
+        }),
```
Prashant-Surya, line 159: should we create a new class to hold onDatabaseFetch and store methods instead of using current manager class? 
This helps removing tight dependency between extension and manager class

apps/live/src/hocuspocus.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    // Extract cookie (fallback to request headers) and userId from token (for scenarios where
+    // the cookies are not passed in the request headers)
+    try {
+      const parsedToken = JSON.parse(token) as TUserDetails;
+      userId = parsedToken.id;
+      cookie = parsedToken.cookie;
+    } catch (error) {
+      // If token parsing fails, fallback to request headers
+      console.error("Token parsing failed, using request headers:", error);
```
Prashant-Surya: should we use logger.error here?

apps/live/src/server.ts
```diff
@@ -33,99 +48,42 @@ export class Server {
...
   public listen() {
     this.serverInstance = this.app.listen(this.app.get("port"), () => {
-      manualLogger.info(`Plane Live server has started at port ${this.app.get("port")}`);
+      logger.info(`Plane Live server has started at port ${this.app.get("port")}`);
     });
   }
 
   public async destroy() {
     // Close the HocusPocus server WebSocket connections
     await this.hocuspocusServer.destroy();
```
Prashant-Surya: Looks like we're not assigning the instance to hocuspocusServer property anymore. Please verify if this will break

apps/live/src/hocuspocus.ts
sriramveeraghanta: Idea is to have tightly coupled methods to this class and create service at the later point to implement the business logic.

PR #7529, opened 2025-08-01
feat: implementing zod schema in services pacakges

sriramveeraghanta:
### Type of Change

- [ ] Code refactoring
- [ ] Documentation update


packages/services/src/auth/sites.ts
```diff
@@ -25,9 +21,9 @@ export class SitesAuthService extends APIService {
...
    * @throws {Error} Throws response data if the request fails
    */
-  async emailCheck(data: IEmailCheckData): Promise<IEmailCheckResponse> {
+  async emailCheck(data: TEmailCheckData): Promise<TEmailCheckResponse> {
     return this.post("/auth/spaces/email-check/", data, { headers: {} })
-      .then((response) => response?.data)
+      .then((response) => EmailCheckResponseSchema.parse(response?.data))
       .catch((error) => {
         throw error?.response?.data;
       });
```
lifeiscontent, line 29: @sriramveeraghanta just calling it out, but it might be good to introduce never throw its a shift in mindset where you treat errors as values, and because they're values we're can safely handle all the edge cases we know about, and in the scenario that there's an edge case that crops up that we otherwise didn't perceive in the future, then and only then will the application throw, and will be caught in the sentry logs so we can fix it.

if we decided to do full server actions, I'd highly suggest effect over zod and never throw as it comes with a whole ecosystem of great tools for handling edge cases safely, but since its a fairly big library, its not a great idea to introduce it to a client side application, although I've heard some people do it anyway because of how good the tooling of effect can be, it outweighs the costs.

PR #7560, opened 2025-08-07
fix: Allow Docker service names in webhook URLs 

### LINKED ISSUE #7555: [bug]: Cannot use service name as webhook URL

### Current behavior

When entering a webhook URL, I'd like to be able to use services I add to the docker compose file as part of the URL. For example, this is supported for the email SMTP server URL (it's currently set to emailproxy and port 2465 so I can use OAuth for emails)
I have a [container that will forward webhooks to Discord](https://github.com/sigmondkukla/plane-discord-webhook-bridge), and I would like to be able to use this URL:
`http://plane-discord-webhook-bridge:8000/webhook`

However, I get a 400 response in in the API logs as well as the browser console when entering this URL.

The workaround for this is exposing it to the internet or using a local IP address such as 192.168.1.123:8000 instead.

### Steps to reproduce

1. Create another service in the docker-compose file generated in the plane-selfhost/plane-app directory and restart the stack
2. Go into Workspace Settings
3. Go into Webhooks
4. Enter the URL referring to the service such as `http://plane-discord-webhook-bridge:8000/webhook`
5. Attempt to create the webhook and observe error in API logs (from setup.sh) and browser console.

### Environment

Production

### Browser

None

### Variant

Self-hosted

### Version

v0.27.1

nehaaprasad:
### Description

fix: #7555

- Fixed webhook URL validation to allow Docker service names while maintaining security. The validation was previously failing for URLs like........  because it attempted DNS resolution on Docker service names that are only resolvable within the Docker network. Added smart detection to differentiate between Docker service names and external domains, skipping DNS validation for Docker services while preserving all security checks for external URLs.

### Type of Change

[x] Bug fix (non-breaking change which fixes an issue)
[ ] Feature (non-breaking change which adds functionality)
[ ] Improvement (change that would cause existing functionality to not work as expected)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation to more accurately block local addresses (e.g., localhost, 127.0.0.1, IPv6 variants) in webhook URLs, including cases with ports.
  * Docker service names are now properly detected and exempted from certain DNS and domain validation checks, reducing false validation errors when using Docker-based hostnames.

### DIFF
--- a/apps/api/plane/app/serializers/webhook.py
+++ b/apps/api/plane/app/serializers/webhook.py
@@ -1,6 +1,7 @@
 # Python imports
 import socket
 import ipaddress
+import re
 from urllib.parse import urlparse
 
 # Third party imports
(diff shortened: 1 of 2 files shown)


sigmondkukla:
It seems like the TLD detection may not be robust enough, I'm wondering if it would be necessary to do at all?
As we detect the use of a TLD and then eventually verify if that is a blocked domain, maybe it would be better to just pass the hostname (which could be a service name) to the disallowed domains validation no matter whether internal or on the web.

nehaaprasad:
@sigmondkukla 
you're right that TLD detection could be more robust! 
the current approach handles `.local` and most common cases, but I could definitely improve it. Would you prefer I

 1. Add more pseudo-TLDs (like `.docker`, `.container`) to the blocklist, or
2. Use a more comprehensive TLD database** for detection?

the current fix works for the reported case (`plane-discord-webhook-bridge` vs `webhook.example.com`) but I'm happy to make it more improve if you think it's worth it for this PR.

sigmondkukla:
It may make sense to actually use the DNS query on the given name as that should give away both if it successfully resolves as well as if it's a FQDN or not. It may successfully resolve in a container either way though because of the docker DNS. I'll take a look into this later.

nehaaprasad:
DNS resolution would be much cleaner than my TLD pattern matching. The tricky part is that Docker DNS will resolve service names inside containers, so we'd need to figure out how to tell internal vs external resolution apart.

This current fix works for the bug, but I like your DNS approach better. Let me know what you find when you look into it

PR #7568, opened 2025-08-10
fix: move helpers file into utils

sriramveeraghanta:
### Description

- Move helpers file into utils in UI package package 

### Type of Change

- [ ] Code refactoring


prateekshourya29:
Let’s directly import `cn` from `@plane/utils` since it’s already included in our UI package dependency. @sriramveeraghanta

prateekshourya29 (approved):

PR #7578, opened 2025-08-13
[WEB-4693] fix: remove initial load flicker on auto-archive and auto-close automation page

prateekshourya29:
### Description

Updated the `ToggleSwitch` components in both `auto-archive-automation.tsx` and `auto-close-automation.tsx` to resolve the flickering issue that occurred during the initial page load. 

### Type of Change

## Summary by CodeRabbit

Here are the updated release notes:

- **Bug Fixes**
  - Auto-archive and auto-close toggles now default to off when settings are unset.
  - Prevents unintended display of duration/details sections when project data is missing.
  - Ensures UI doesn't show active automation states based on undefined values.

- **Refactor**
  - Consolidated status logic for auto-archive and auto-close to ensure consistent behavior and simpler maintenance.

### DIFF
--- a/apps/web/core/components/automation/auto-archive-automation.tsx
+++ b/apps/web/core/components/automation/auto-archive-automation.tsx
@@ -49,7 +49,7 @@ export const AutoArchiveAutomation: React.FC<Props> = observer((props) => {
   );
 
   const autoArchiveStatus = useMemo(() => {
-    if (currentProjectDetails === undefined || currentProjectDetails.archive_in === undefined) return false;
+    if (currentProjectDetails?.archive_in === undefined) return false;
     return currentProjectDetails.archive_in !== 0;
   }, [currentProjectDetails]);
 
--- a/apps/web/core/components/automation/auto-archive-automation.tsx
+++ b/apps/web/core/components/automation/auto-archive-automation.tsx
@@ -48,6 +48,11 @@ export const AutoArchiveAutomation: React.FC<Props> = observer((props) => {
     currentProjectDetails?.id
   );
 
+  const getAutoArchiveStatus = () => {
+    if (currentProjectDetails === undefined || currentProjectDetails.archive_in === undefined) return false;
+    return currentProjectDetails.archive_in !== 0;
+  };
+
   return (
     <>
       <SelectMonthModal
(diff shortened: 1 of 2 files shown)


apps/web/core/components/automation/auto-archive-automation.tsx
```diff
@@ -48,6 +48,11 @@ export const AutoArchiveAutomation: React.FC<Props> = observer((props) => {
     currentProjectDetails?.id
   );
 
+  const getAutoArchiveStatus = () => {
+    if (currentProjectDetails === undefined || currentProjectDetails.archive_in === undefined) return false;
```
aaryan610: There's no need of this check.

apps/web/core/components/automation/auto-archive-automation.tsx
```diff
@@ -71,7 +76,7 @@ export const AutoArchiveAutomation: React.FC<Props> = observer((props) => {
             </div>
           </div>
           <ToggleSwitch
-            value={currentProjectDetails?.archive_in !== 0}
+            value={getAutoArchiveStatus()}
```
aaryan610: Instead of calling the function everytime, store it in a variable.

apps/web/core/components/automation/auto-close-automation.tsx
```diff
@@ -75,6 +75,11 @@ export const AutoCloseAutomation: React.FC<Props> = observer((props) => {
     currentProjectDetails?.id
   );
 
+  const getAutoCloseStatus = () => {
+    if (currentProjectDetails === undefined || currentProjectDetails.close_in === undefined) return false;
```
aaryan610: Same for here.

apps/web/core/components/automation/auto-archive-automation.tsx
prateekshourya29: It’s required because this was the cause of the flicker. We’re directly comparing `currentProjectDetails?.archive_in !== 0` and rendering the details, which is `true` even if the archive_in is `undefined`.

aaryan610: What I mean is, there's no need to check for `currentProjectDetails === undefined`

prateekshourya29: It will remain the same. Without that, we’ll check `currentProjectDetails?.archive_in === undefined` with optional chaining, which essentially combines both these conditions.

PR #7597, opened 2025-08-19
chore: created new description model

NarayanBavisetti:
### Description

this pull request creates a new description and description version model.

### Type of Change


apps/api/plane/db/models/description.py
```diff
@@ -0,0 +1,56 @@
...
+
+class Description(WorkspaceBaseModel):
+
+
+    description_json = models.JSONField(default=dict, blank=True)
+    description_html = models.TextField(blank=True, default="<p></p>")
+    description_binary = models.BinaryField(null=True)
+    description_stripped = models.TextField(blank=True, null=True)
+
+    class Meta:
```
dheeru0198, line 14: @NarayanBavisetti Check if we can add any indexes that will help with query performance.

PR #7614, opened 2025-08-21
[WEB-4725] chore: storybook setup & tailwind config package improvements

anmolsinghbhatia:
### Description
This PR includes the following changes:
- Integrated Storybook into the Propel package for component development and documentation
- Added global CSS file to the common @plane/tailwind-config package
- Introduced cn utility in the Propel package

## Summary by CodeRabbit

- New Features
  - Themeable design system with global CSS variables, Tailwind layers, UI utilities (scrollbars, inputs, overlays, progress, tags) for consistent light/dark themes.
  - Storybook added for interactive component preview and documentation, with dev and build scripts.

- Chores
  - Linting updated to include Storybook rules.
  - Task runner updated to build Storybook outputs.
  - Repository ignore updated to exclude generated Storybook assets.

- Refactor
  - Added a small class-name utility and a utils barrel for cleaner imports.


packages/propel/.storybook/main.ts
```diff
@@ -0,0 +1,20 @@
+import type { StorybookConfig } from "@storybook/react-vite";
+
+import { join, dirname } from "path";
+
+/**
+ * This function is used to resolve the absolute path of a package.
+ * It is needed in projects that use Yarn PnP or are set up within a monorepo.
+ */
+function getAbsolutePath(value: string): any {
```
lifeiscontent:
```suggestion
function getAbsolutePath(value: string) {
```

is `any` necessary? I feel like in recent versions of storybook, they've done a massive cleanup of types.

lifeiscontent: I think instead of using require.resolve, we might be able to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta/resolve if this project is ESM.

packages/propel/.storybook/preview.ts
```diff
@@ -0,0 +1,14 @@
+import type { Preview } from "@storybook/react-vite";
+import "@plane/tailwind-config/global.css";
+
+const preview: Preview = {
+  parameters: {
+    controls: {
+      matchers: {
+        color: /(background|color)$/i,
+        date: /Date$/i,
```
lifeiscontent: these are example matchers, don't merge them 😅

packages/propel/src/utils/classname.tsx
```diff
@@ -0,0 +1,4 @@
+import { clsx, type ClassValue } from "clsx";
+import { twMerge } from "tailwind-merge";
```
lifeiscontent: I would highly suggest avoiding tailwind-merge if possible, it treats a symptom, not the cause. https://www.psychologytoday.com/us/blog/the-resilient-brain/201712/are-you-simply-treating-your-symptoms#:~:text=People%20normally%20treat%20the%20symptom,the%20symptom%20will%20eventually%20stop.

e.g. instead of doing this, you could create components that use props as a contract for what you want visually, and then use something like clsx to compose styles together based on the requirements of the props.

from my point of view, tailwind-merge causes down stream head aches because if you need to refactor a component, and the component exposes className, then the design team asks you to increase the padding of the `lg` button, well, guess what? you potentially break every page that uses a button with `className` now. storybook is also a great use case to make sure that every aspect of a visual representation matches a spec in Figma, and then we could have visual regression tests that run in CI to be absolutely sure that the styles have not diverged.

packages/propel/package.json
```diff
@@ -9,7 +9,9 @@
     "check:format": "prettier --check \"**/*.{ts,tsx,md,json,css,scss}\"",
     "fix:lint": "eslint . --fix",
     "fix:format": "prettier --write \"**/*.{ts,tsx,md,json,css,scss}\"",
-    "clean": "rm -rf .turbo && rm -rf .next && rm -rf node_modules && rm -rf dist"
+    "clean": "rm -rf .turbo && rm -rf .next && rm -rf node_modules && rm -rf dist",
+    "storybook": "storybook dev -p 6006",
+    "build-storybook": "storybook build"
```
lifeiscontent, line 14: please add outputs to `turbo.json` so it can cache the builds ✨

packages/propel/package.json
```diff
@@ -29,17 +31,23 @@
     "@plane/utils": "workspace:*",
     "@tanstack/react-table": "^8.21.3",
     "class-variance-authority": "^0.7.1",
+    "clsx": "^2.1.1",
     "lucide-react": "^0.469.0",
+    "postcss": "^8.5.6",
```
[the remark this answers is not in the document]
lifeiscontent: @anmolsinghbhatia this is a solid ask, postcss has nothing to do with the runtime of our application, by the time the app runs, CSS is already compiled, unless you plan on sharing `postcss` with downstream consumers, in which case I would probably do it in a `tailwind-config` package like we do with marketing repo.

packages/tailwind-config/tailwind.config.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
       "./pages/**/*.tsx",
       "./app/**/*.tsx",
       "./ui/**/*.tsx",
+      "./propel/**/*.{js,ts,jsx,tsx}",
```
lifeiscontent: fwiw, not suggesting to do it now unless you want to, but all downstream consumers should extend this config rather than specifying all paths of everything that use it, that way the code has one responsibility.

packages/propel/.storybook/main.ts
anmolsinghbhatia: I tried import.meta.resolve first, but it caused a Storybook error because the .storybook/main.ts config runs in CommonJS mode, not ESM. 

The error was: `"import.meta" is not available with the "cjs" output format`

So require.resolve is the correct choice here for Storybook compatibility, even though import.meta.resolve would be preferred in a pure ESM environment.

packages/propel/package.json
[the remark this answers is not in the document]
anmolsinghbhatia: Removed postcss dependency from Propel package since it’s already included in tailwind-config

packages/propel/.storybook/preview.ts
anmolsinghbhatia: Done, I’ve removed them.

packages/propel/src/utils/classname.tsx
anmolsinghbhatia: Thanks for the detailed context, that makes sense regarding tailwind-merge and downstream issues. I’ve already removed it, and we can shift towards props + clsx with Storybook for consistency.

packages/tailwind-config/tailwind.config.js
anmolsinghbhatia: Resolved

packages/propel/package.json
anmolsinghbhatia: Done

packages/tailwind-config/global.css
```diff
@@ -0,0 +1,696 @@
+@tailwind base;
```
lifeiscontent, line 1: @anmolsinghbhatia why are we using all this CSS instead of the tailwind.config.ts?

anmolsinghbhatia: The .css files are for our global variables that feed into the Tailwind config (tokens). We still need them since the tokens get manipulated for custom theming. In this PR, I just moved the global CSS into the Tailwind config package so we don’t have to redefine them across apps.

PR #7617, opened 2025-08-21
[WEB-4434]fix: user name validation

sangeethailango:
### Type of Change

- [ ] Code refactoring
- [ ] Documentation update

### DIFF
--- a/apps/api/plane/db/models/user.py
+++ b/apps/api/plane/db/models/user.py
@@ -11,6 +11,8 @@ from django.db import models
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 from django.utils import timezone
+import re
+from django.core.exceptions import ValidationError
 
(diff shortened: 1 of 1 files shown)


dheeru0198 (dismissed):

sriramveeraghanta:
@dheeru0198 What happens in case of people use different language than english? Does this Validation fail?

sangeethailango:
@sriramveeraghanta They cannot use other languages because we only accept English letters, and if they try, an error will be shown.

dheeru0198:
Let's wait until we figure out if non english languages should be allowed or not

sangeethailango:
@dheeru0198  Instead of accepting only alphanumeric letters, we can validate if there are any special characters in the payload, and if so, we can throw an error. That will also work right?

PR #7666, opened 2025-08-28
[WEB-4559] fix: workspace menu item active state

anmolsinghbhatia:
### Description
This PR fixes an issue where the workspace menu item was not properly highlighting the current tab.  

### Type of Change
- [x] Bug fix

### DIFF
--- a/apps/web/core/components/workspace/sidebar/sidebar-item.tsx
+++ b/apps/web/core/components/workspace/sidebar/sidebar-item.tsx
@@ -48,7 +48,7 @@ export const SidebarItemBase: FC<Props> = observer(({ item, additionalRender, ad
   if (!isPinned && !staticItems.includes(item.key)) return null;
 
   const itemHref = item.key === "your_work" ? `/${slug}${item.href}${data?.id}/` : `/${slug}${item.href}`;
-  const isActive = itemHref === pathname;
+  const isActive = item.key === "home" ? pathname === itemHref : pathname.includes(itemHref);
   const icon = getSidebarNavigationItemIcon(item.key);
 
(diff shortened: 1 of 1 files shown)


apps/web/core/components/workspace/sidebar/sidebar-item.tsx
```diff
@@ -48,7 +48,7 @@ export const SidebarItemBase: FC<Props> = observer(({ item, additionalRender, ad
   if (!isPinned && !staticItems.includes(item.key)) return null;
 
   const itemHref = item.key === "your_work" ? `/${slug}${item.href}${data?.id}/` : `/${slug}${item.href}`;
-  const isActive = itemHref === pathname;
+  const isActive = item.key === "home" ? pathname === itemHref : pathname.includes(itemHref);
```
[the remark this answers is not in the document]
vamsikrishnamathala: @anmolsinghbhatia  Check this once.

prateekshourya29: Instead of this, could we create a `highlight` function, similar to other constants, to eliminate all these conditions and simplify and clean up the code?

prateekshourya29 (changes requested):

apps/web/core/components/workspace/sidebar/sidebar-item.tsx
anmolsinghbhatia: I think this logic is specific to this file, so a helper might not add much value here, happy to adjust if you feel otherwise.

prateekshourya29 (approved):

PR #7764, opened 2025-09-10
[WEB-4861] fix: update redirection path in MagicSignInEndpoint

JayashTripathy:
### Description
Add email search param while redirecting to /set-password to populate the page form

### Type of Change


dheeru0198:
This needs a different fix. Will be opening another PR for that.

PR #7784, opened 2025-09-12
[WIKI-659] chore: added issue relation and page sort order

NarayanBavisetti:
### Description

this pull request does a data migration in the sort_order field in which we randomly assign the number in the sort_order. Also added a new choice in the issue relation.

### Type of Change

## Summary by CodeRabbit

- New Features
  - Project-level webhooks: configure and manage webhooks per project with safeguards against duplicate active mappings.
  - New issue relation: “Implemented By” (with inverse “Implements”) now available for linking issues.

- Bug Fixes
  - More consistent default ordering for pages to improve list stability and predictability.

- Chores
  - Internal migrations and model updates to support new webhook associations and relation types.


apps/api/plane/db/migrations/0106_auto_20250912_0845.py
```diff
@@ -0,0 +1,52 @@
...
+
+    operations = [
+        migrations.AlterField(
+            model_name="issuerelation",
+            name="relation_type",
+            field=models.CharField(
+                default="blocked_by", max_length=20, verbose_name="Issue Relation Type"
+            ),
+        ),
+        migrations.RunPython(set_page_sort_order),
```
dheeru0198: @NarayanBavisetti Add a reverse function. But just don't add noop, if reverting, revert back to the old value.

PR #7793, opened 2025-09-13
chore(i18n): added scripts and utilities to check missing translations across locales

JayashTripathy:
### Description
Added locale scripts and respective tools for checking missing translations in i18n package. to maintain consistency across all locales

```
> @plane/i18n@1.0.0 check:translations 
> tsx scripts/check-missing-translations.ts

[34m[39m
[34mChecking for missing translations...[39m
[34m[39m
[33m[39m
[33mChecking 📄 translations:[39m
[37m[39m
[37m🔑 Key: preferences[39m
[31m  - Missing in 📂 locales/cs/translations.json[39m
[31m  - Missing in 📂 locales/de/translations.json[39m
[31m  - Missing in 📂 locales/es/translations.json[39m
```

### Type of Change

- [x] Code refactoring

image


packages/i18n/scripts/check-missing-translations.ts
Prashant-Surya, line 1: Can we make this into a unit test that can run with jest? Easier to plug this into build process that way

packages/i18n/scripts/locale/json-service.ts
```diff
@@ -0,0 +1,52 @@
...
+          Object.assign(acc, this.flatten(obj[key] as NestedTranslations, newKey));
+        } else {
+          // Handle empty strings explicitly
+          acc[newKey] = obj[key] === "" ? "" : (obj[key] as string);
+        }
+        return acc;
+      },
+      {} as Record<string, string>
+    );
+  }
```
Prashant-Surya: Can we make this more readable like below?
```suggestion
flatten(obj: NestedTranslations, prefix = ""): Record<string, string> {
  const result: Record<string, string> = {};
  
  for (const [key, value] of Object.entries(obj)) {
    const newKey = prefix ? `${prefix}.${key}` : key;
    
    if (typeof value === 'object' && value !== null) {
      Object.assign(result, this.flatten(value as NestedTranslations, newKey));
    } else {
      result[newKey] = value as string;
    }
  }
  
  return result;
}
```

packages/i18n/scripts/locale/manager.ts
```diff
@@ -0,0 +1,186 @@
...
+  public readonly translationFiles: TranslationFile[];
+
+  /**
+   * Initializes the TranslationManager with default paths and files.
+   * Automatically updates generated translations for each file.
+   */
+  constructor() {
+    this.rootPath = TRANSLATION_ROOT_PATH;
+    this.translationFiles = TRANSLATION_FILES;
+    this.translationFiles.forEach((f) => this.updateGeneratedTranslations(f));
```
Prashant-Surya: let's not run this in constructor. Call this function separately

packages/i18n/scripts/locale/manager.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  /**
+   * Updates a single translation key for a specific locale
+   * @param locale The locale to update
+   * @param key The translation key (dot-notation path)
+   * @param value The new translation value
+   * @param file The translation file category (e.g., "translations", "accessibility", "editor", "core")
+   * @throws If the file cannot be written
+   */
+  async updateTranslation(locale: TranslationLocale, key: string, value: string, file: TranslationFile): Promise<void> {
```
Prashant-Surya, line 84: Let's check if this is changing the order of keys in the json file. If it's changing the key order, we can try sorting the keys before adding in the json

packages/i18n/scripts/locale/json-service.ts
JayashTripathy: sure ill update

packages/i18n/scripts/locale/manager.ts
JayashTripathy: alright

packages/i18n/scripts/locale/manager.ts
JayashTripathy: Yes, it's sorted. It's doing DFS

PR #7841, opened 2025-09-23
[WIKI-638] fix: peek overview closing while dropdowns are open

aaryan610:
### Description

This PR fixes the bug where peek overview is getting closed on outside click/pressing the `Esc` key while editor dropdowns are still open.

### Solution:

1. Table dropdowns are now tracked using the utility extension storage.
2. Editor ref now exposes another callback to find if any dropdowns are open or not.
3. Instead of checking for DOM elements before closing the peek overview, we now use the editor ref helper.
4. Created new commands to manage active dropbars instead of manually checking at every instance.

### Type of Change

## Summary by CodeRabbit

* **New Features**
  * App now tracks editor dropdown/menu state and shares that state with issue peek views.

* **Bug Fixes**
  * Prevents accidental closing or route changes when editor menus/dropbars are open.
  * Improved outside-click and keydown handling to respect active editor menus.

* **Documentation**
  * Editor API expanded with a method to check whether any editor dropdown/menu is currently open.

### DIFF
--- a/packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.tsx
+++ b/packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.tsx
@@ -205,7 +205,7 @@ export const ColumnDragHandle: React.FC<ColumnDragHandleProps> = (props) => {
               zIndex: 100,
             }}
           >
-            <ColumnOptionsDropdown editor={editor} onClose={() => setIsDropdownOpen(false)} />
+            <ColumnOptionsDropdown editor={editor} onClose={() => context.onOpenChange(false)} />
           </div>
         </FloatingPortal>
       )}
--- a/packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.tsx
+++ b/packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.tsx
@@ -12,9 +12,11 @@ import {
 } from "@floating-ui/react";
 import type { Editor } from "@tiptap/core";
 import { Ellipsis } from "lucide-react";
-import { useCallback, useState } from "react";
+import { useCallback, useEffect, useState } from "react";
 // plane imports
 import { cn } from "@plane/utils";
+// constants
+import { CORE_EXTENSIONS } from "@/constants/extension";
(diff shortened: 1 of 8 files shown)


packages/editor/src/core/extensions/table/plugins/drag-handles/row/drag-handle.tsx
```diff
@@ -156,6 +160,21 @@ export const RowDragHandle: React.FC<RowDragHandleProps> = (props) => {
     [editor, row]
   );
 
+  // toggle utility storage
```
Palanikannan1437: can we remove this useEffect and move this logic to a function in regards of [this](https://react.dev/learn/you-might-not-need-an-effect) and do something like...  

```tsx
const handleOpenChange = useCallback((open: boolean) => {
  setIsDropdownOpen(open);
  
  // Update utility storage
  const utilityStorage = getExtensionStorage(editor, CORE_EXTENSIONS.UTILITY);
  const index = utilityStorage.activeDropbarExtensions.indexOf(CORE_EXTENSIONS.TABLE);
  
  if (open) {
    if (index === -1) {
      utilityStorage.activeDropbarExtensions.push(CORE_EXTENSIONS.TABLE);
    }
  } else {
    if (index !== -1) {
      utilityStorage.activeDropbarExtensions.splice(index, 1);
    }
  }
}, [editor]);

const handleClose = useCallback(() => {
  setIsDropdownOpen(false);
  // The utility storage cleanup will be handled by onOpenChange when it gets called with false
}, []);

const { refs, floatingStyles, context } = useFloating({
  // ... other props
  open: isDropdownOpen,
  onOpenChange: handleOpenChange,
  // ... rest
});
```

packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.tsx
```diff
@@ -157,6 +161,21 @@ export const ColumnDragHandle: React.FC<ColumnDragHandleProps> = (props) => {
     [col, editor]
   );
 
+  // toggle utility storage
```
Palanikannan1437: same suggestion as below

PR #7865, opened 2025-09-29
[WEB-4969] feat: add toggle button for work item filters row visibility

prateekshourya29:
### Description

This PR introduces a toggle button for filters across all work item layouts, allowing users to hide the filters bar.

### Type of Change

### DIFF
--- a/packages/shared-state/src/store/rich-filters/filter-helpers.ts
+++ b/packages/shared-state/src/store/rich-filters/filter-helpers.ts
@@ -1,7 +1,7 @@
 import { cloneDeep } from "lodash-es";
-import { toJS } from "mobx";
+import { action, makeObservable, observable, toJS } from "mobx";
 // plane imports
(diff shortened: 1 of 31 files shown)


packages/shared-state/src/store/rich-filters/filter-helpers.ts
```diff
@@ -87,6 +110,41 @@ export class FilterInstanceHelper<P extends TFilterProperty, E extends TExternal
...
+  /**
+   * Toggles the visibility of the filter.
+   * @param isVisible - The visibility to set.
+   */
+  toggleVisibility: IFilterInstanceHelper<P, E>["toggleVisibility"] = action((isVisible) => {
+    if (isVisible !== undefined) {
+      this.isVisible = isVisible;
+      return;
+    }
+    this.isVisible = !this.isVisible;
```
aaryan610, line 145: Let's just use `??` here instead of an if-else.

prateekshourya29: I feel this is more readable.

PR #7916, opened 2025-10-07
improvement: enhance rich filters with new components and configurations

prateekshourya29:
- Introduced `FilterItem` and related components for better filter item management.
- Updated filter configurations to include new properties and support for multiple values.
- Improved loading states and error handling in filter components.
- Refactored existing filter logic to streamline operations and enhance performance.

### Type of Change


arpadgabor:
To be honest I can't really tell from the code, does this PR by any chance also add support for an "is empty" filter?

PR #7979, opened 2025-10-16
Update translations.ts: issue-artifacts discovered

conny3496:
### Type of Change

- [x] Code refactoring
- [ ] Documentation update

## Summary by CodeRabbit

* **Localization**
  * Updated terminology in workspace analytics empty state messaging to consistently refer to "work items" instead of "issues."


conny3496:
Was originally looking to improve on this:
image
...but it turns out that it is already fixed in the Cloud-Version and it only shows like that on my/our local instance. Found the other artifacts while double-checking.

PR #7997, opened 2025-10-23
fix: live server secret key app version env variables

sriramveeraghanta:
### Description

- Updated App Version with v1.1.0
- Added live server secrete key env to api server

### Type of Change

### DIFF
--- a/deployments/cli/community/variables.env
+++ b/deployments/cli/community/variables.env
@@ -77,6 +77,6 @@ MINIO_ENDPOINT_SSL=0
 # API key rate limit
 API_KEY_RATE_LIMIT=60/minute
 
-# Live server envs
+# Live server environment variables
 # WARNING: You must set a secure value for LIVE_SERVER_SECRET_KEY in production environments.
 LIVE_SERVER_SECRET_KEY=


deployments/cli/community/variables.env
```diff
@@ -1,5 +1,5 @@
 APP_DOMAIN=localhost
-APP_RELEASE=stable
+APP_RELEASE=v1.1.0
```
mguptahub: this is not required. Please revert. The setup.sh will take care of updating with the latest release downloaded

deployments/cli/community/variables.env
```diff
@@ -76,3 +76,6 @@ MINIO_ENDPOINT_SSL=0
 
 # API key rate limit
 API_KEY_RATE_LIMIT=60/minute
+
+# Live server envs
+LIVE_SERVER_SECRET_KEY=
```
mguptahub, line 82: Better would be fill with the default value as in docker-compose.yml

sriramveeraghanta: Default value is already handled in Docker compose file. Having this empty env variable empty indicates the user should update that with a different value.

deployments/cli/community/variables.env
sriramveeraghanta: Done.

PR #8005, opened 2025-10-23
fix: get work item activity external endpoint

sriramveeraghanta:
### Type of Change

## Summary by CodeRabbit

* **Bug Fixes**
  * Issue activity detail endpoints now return 404 when a requested activity is not found.

* **Improvements**
  * Label updates now require both external ID and source together to avoid ambiguous updates.
  * Issue activity detail views now return a single item (not a paginated list) for clearer, faster responses.

### DIFF
--- a/apps/api/plane/api/views/issue.py
+++ b/apps/api/plane/api/views/issue.py
@@ -1696,14 +1696,16 @@ class IssueActivityDetailAPIEndpoint(BaseAPIView):
         Excludes comment, vote, reaction, and draft activities.
         """
         issue_activity = (
-            IssueActivity.objects.filter(issue_id=issue_id, workspace__slug=slug, project_id=project_id)
-            .filter(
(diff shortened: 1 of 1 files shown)


apps/api/plane/api/views/issue.py
```diff
@@ -1695,7 +1695,7 @@ def get(self, request, slug, project_id, issue_id, pk):
         Retrieve details of a specific activity.
         Excludes comment, vote, reaction, and draft activities.
         """
-        issue_activities = (
+        issue_activity = (
             IssueActivity.objects.filter(issue_id=issue_id, workspace__slug=slug, project_id=project_id)
```
Prashant-Surya: Missed adding ` id=pk` in the filter

sriramveeraghanta: done.

PR #8352, opened 2025-12-17
feat: Implement webhooks using event stream. 

sriramveeraghanta:
### Description

- Created an event stream using outbox architecture.
- Adding outbox poller on poll on the table and retrieve the events from DB and push them to MQ in fanout. 
- Adding Webhook consumer to listen to these events from the exchange and process them using a queue.

### Type of Change


pySilver:
@sriramveeraghanta why it was abandoned? I've read your https://plane.so/blog/building-zero-loss-event-streaming-with-postgresql-django-and-rabbitmq blog post and it seems as good idea but I cant trace usage in the repo and this PR is itself closed. Why so?

sriramveeraghanta:
We are attempting to optimize the workflow before implementing any changes. Given the significant nature of this change, it is crucial to ensure that the community’s modifications do not cause any disruptions.

pySilver:
@sriramveeraghanta thanks to make it clear!

PR #8398, opened 2025-12-19
[WEB-5675] chore: implement `fontsource` as the fonts library

aaryan610:
### Description

This PR optimizes fonts across all apps. It introduces `@fontsource` as the font library.

1. Font imports are removed from the common `tailwind-config` package and moved to specific apps for proper resolution during build time.
2. Replaces `Inter` with `Inter Variable`.
3. Removes all local ofnt files and css declarations from the `propel` package.

### Type of Change

## Summary by CodeRabbit

* **Chores**
  * Centralized font handling and switched typography to variable Inter across apps.
  * Added font preloading for Inter and consolidated font dependencies; removed legacy font stylesheet imports.
  * Introduced material-symbols typography utility.

* **Bug Fixes**
  * Improved icon/font loading behavior so icons show reliable fallbacks while fonts load.

### DIFF
--- a/packages/propel/src/emoji-icon-picker/logo.tsx
+++ b/packages/propel/src/emoji-icon-picker/logo.tsx
@@ -34,7 +34,6 @@ export function Logo({ logo, size = 16, type = "material" }: Props) {
   const value = in_use === "emoji" ? emoji?.value : icon?.name;
 
   if (!value) return loadingSkeleton;
-  if (!isMaterialSymbolsFontLoaded && type === "material") return loadingSkeleton;
 
   // Emoji rendering
   if (in_use === "emoji") {
@@ -65,6 +64,8 @@ export function Logo({ logo, size = 16, type = "material" }: Props) {
       const LucideIconElement = lucideIcon.element;
       return <LucideIconElement style={{ color, height: size, width: size }} />;
     }
+
+    if (!isMaterialSymbolsFontLoaded) return loadingSkeleton;
 
     // Material icon
     return (


apps/admin/app/root.tsx
```diff
@@ -10,6 +10,11 @@ import { LogoSpinner } from "@/components/common/logo-spinner";
 import globalStyles from "@/styles/globals.css?url";
 import type { Route } from "./+types/root";
 import { AppProviders } from "./providers";
+// fonts
+import "@fontsource-variable/inter";
+import interVariableWoff2 from "@fontsource-variable/inter/files/inter-latin-wght-normal.woff2?url";
+import "@fontsource/material-symbols-rounded";
+import "@fontsource/ibm-plex-mono";
```
lifeiscontent, line 17: have you tried changing files and watching live reload? if it doesn't crash this is fine, but I was experiencing issues with doing side effect imports like this previously.

I would use the `?url` feature of vite and stick everything in links.

lifeiscontent: you might also try using the `@import` in the app css for fonts, as tailwind might be smart enough to merge it down into 1 sheet.

aaryan610: Yes, we tested this approach. Everything works as expected.

aaryan610: We tried using `@import`, but faced the same resolution issues during build.

lifeiscontent (approved):

PR #8463, opened 2025-12-29
[WEB-5512]feat: workspace api tokens

sangeethailango:
### Description
Implementation of workspace api tokens.

### Type of Change

## Release Notes

* **New Features**
  * Added workspace-level API token management within workspace settings, allowing admins to create, view, and delete workspace-scoped tokens.
  * Introduced dynamic rate limiting per API token based on configured limits.

* **UI/UX Improvements**
  * Updated terminology from "Personal Access Tokens" to "Access Tokens" for clarity.
  * Enhanced token management components with improved creation and deletion workflows.
  * Added new "Access Tokens" page in workspace settings for centralized token administration.
  * Updated translations across 18+ languages to reflect new token terminology.

### DIFF
--- a/apps/api/plane/app/views/api/workspace.py
+++ b/apps/api/plane/app/views/api/workspace.py
@@ -1,3 +1,7 @@
+# Copyright (c) 2023-present Plane Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
 # Python import
 from typing import Optional
 from uuid import uuid4
--- a/apps/api/plane/app/views/api/__init__.py
+++ b/apps/api/plane/app/views/api/__init__.py
@@ -1,3 +1,7 @@
+# Copyright (c) 2023-present Plane Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
 from .base import ApiTokenEndpoint
 from .service import ServiceApiTokenEndpoint
 from .workspace import WorkspaceAPITokenEndpoint
--- a/apps/api/plane/app/views/api/service.py
+++ b/apps/api/plane/app/views/api/service.py
@@ -1,3 +1,7 @@
+# Copyright (c) 2023-present Plane Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
 # Python import
 from uuid import uuid4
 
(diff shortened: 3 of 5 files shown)


apps/api/plane/app/views/api/workspace.py
```diff
@@ -0,0 +1,59 @@
...
+            description=description,
+            user=request.user,
+            user_type=user_type,
+            expired_at=expired_at,
+            workspace=workspace,
+        )
+
+        serializer = APITokenSerializer(api_token)
+
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
```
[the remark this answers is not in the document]
sangeethailango: Since the slug is in the URL, if it doesn’t exist, the request won’t even reach the method.

apps/api/plane/api/middleware/api_authentication.py
```diff
@@ -38,10 +46,12 @@ def validate_api_token(self, token):
         return (api_token.user, api_token.token)
 
     def authenticate(self, request):
+        workspace_slug = resolve(request.path_info).kwargs.get("slug")
```
pablohashescobar: is there a safer way to do this, does resolve always has kwargs?

pablohashescobar (approved):

apps/api/plane/api/middleware/api_authentication.py
sangeethailango: The resolve function is returning a ResolveMatch match object, which always returns a kwargs; if no args are given, it returns an empty {}.

sangeethailango:
https://github.com/makeplane/plane/pull/8339 needs to be merged before this, as it contains the allowed_rate_limit field, which is required here.

anmolsinghbhatia (approved):

pablohashescobar (approved):

apps/api/plane/api/middleware/api_authentication.py
```diff
@@ -22,13 +23,20 @@ class APIKeyAuthentication(authentication.BaseAuthentication):
...
-    def validate_api_token(self, token):
+    def validate_api_token(self, token, workspace_slug):
         try:
             api_token = APIToken.objects.get(
                 Q(Q(expired_at__gt=timezone.now()) | Q(expired_at__isnull=True)),
                 token=token,
                 is_active=True,
             )
+
+            if workspace_slug:
```
dheeru0198: @sangeethailango This will cause the existing tokens to not work. Check if the APIToken is created with a workspace first and then validate it's workspace id

dheeru0198 (approved):

PR #8544, opened 2026-01-15
[WIKI-874] refactor: description input component

aaryan610:
### Description

This PR refactors the description input component to pass `description_json` as a part of the `onSubmit` callback function.

### Type of Change

- [x] Code refactoring

## Summary by CodeRabbit

* **New Features**
  * Description input now supports both HTML and structured JSON content for richer editor data.

* **Refactor**
  * Unified description submission flow across multiple components to send the combined HTML+JSON payload and streamline save behavior.


apps/web/core/components/editor/rich-text/description-input/root.tsx
```diff
@@ -28,6 +28,7 @@ const workspaceService = new WorkspaceService();
 type TFormData = {
   id: string;
   description_html: string;
+  description_json?: object;
```
Palanikannan1437, line 31: [LOW] description_json is typed as bare 'object', which is overly permissive

Why: Loose typing reduces compile-time safety. The editor likely produces a specific document structure.

Suggested fix: Use a more specific type from the editor package (e.g., TipTap's JSONContent) instead of 'object'.

aaryan610: This change will require changes in the editor package as well. I'll keep a note of this and create a work item for the same to fix all such instances with a separate PR.

PR #8574, opened 2026-01-22
refactor: editor table performance optimization

aaryan610:
### Description

This PR optimizes editor performance when tables are a part of the content.

1. Drag handles are now created using Vanilla JS instead of the earlier, Tiptap's `ReactRenderer` which introduced a lot of overheads.
2. Insert plugin now batches updates instead of triggering an update for each minor update.
3. The `findAllTables` utility now has a 10x improvement with more efficient logic to look for tables instead of the earlier logic, the complexity of which was `O(table x nodes)`. The updated logic has the time complexity of `O(nodes)`.

### Type of Change

- [x] Performance improvements

## Summary by CodeRabbit

* **New Features**
  * Color customization for table cells via a drag-handle dropdown.
  * New, clearer icons for table row/column operations.

* **Performance**
  * Debounced table update scheduling for smoother responsiveness.

* **Refactor**
  * Reworked table row/column drag-handle implementation for improved reliability and lifecycle management.

* **Style**
  * Updated handle button styles and open/hover visuals for better discoverability.

### DIFF
--- a/packages/editor/src/styles/table.css
+++ b/packages/editor/src/styles/table.css
@@ -57,7 +57,7 @@
       .table-col-handle-container,
       .table-row-handle-container {
         & > button {
-          @apply py-1 opacity-0 rounded-sm outline-none;
+          @apply opacity-0 rounded-sm outline-none;
 
           &.default-state {
             @apply bg-layer-1 hover:bg-layer-1-hover border border-strong-1;
@@ -66,6 +66,22 @@
           &.open-state {
             @apply opacity-100! bg-accent-primary border-accent-strong;
           }
+        }
+      }
+
+      .table-col-handle-container {
+        & > button {
+          @apply px-1;
+
+          svg {
+            @apply rotate-90;
+          }
+        }
+      }
+
+      .table-row-handle-container {
+        & > button {
+          @apply py-1;
         }
       }
 


packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.ts
```diff
@@ -0,0 +1,459 @@
...
+
+    // Update button to open state
+    button.className =
+      "px-1 bg-layer-1 border border-strong-1 rounded-sm outline-none transition-all duration-200 opacity-100 bg-accent-primary border-accent-strong";
+
+    // Add active dropdown extension
+    editor.commands.addActiveDropbarExtension(CORE_EXTENSIONS.TABLE);
+
+    // Create backdrop
+    backdropElement = document.createElement("div");
```
Palanikannan1437, line 133: Loss of scroll lock behavior from React FloatingOverlay
The original React implementation used <FloatingOverlay lockScroll /> which prevented background scrolling while the dropdown was open. The new vanilla JS implementation creates a simple backdrop div without scroll locking (backdropElement.style.cssText = "position: fixed; inset: 0; z-index: 99;"). This is a behavioral change - users can now scroll the page while the dropdown is open, which may cause the dropdown to move or be positioned incorrectly relative to its anchor button.

packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.ts
```diff
@@ -0,0 +1,459 @@
...
+        if (dropdownElement) {
+          dropdownElement.style.left = `${x}px`;
+          dropdownElement.style.top = `${y}px`;
+        }
+        return;
+      });
+    });
+
+    // Handle keyboard events
+    const handleKeyDown = (event: KeyboardEvent) => {
```
Palanikannan1437, line 179: Keyboard handler closes dropdown on ANY key press
The handleKeyDown event listener at drag-handle.ts:179-183 and drag-handle.ts:179-183 closes the dropdown on ANY key press, not just Escape. This matches the original React implementation behavior but is unusual UX - typically only Escape would close a dropdown while other keys might navigate or be ignored. The original React code had the same behavior via the context.onOpenChange(false) call in handleKeyDown.

packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.ts
```diff
@@ -0,0 +1,459 @@
...
+      }
+
+      // Calculate drop index
+      const currentLeft = startLeft + moveEvent.clientX - startXPos;
+      dropIndex = calculateColumnDropIndex(col, columns, currentLeft);
+
+      // Update visual markers if they exist
+      if (dropMarker && dragMarker) {
+        if (!pseudoColumn) {
+          pseudoColumn = constructColumnDragPreview(editor, editor.state.selection, table);
```
Palanikannan1437, line 393: Stale table reference used in drag preview construction during handleMove
In drag-handle.ts:393 and drag-handle.ts:393, constructColumnDragPreview/constructRowDragPreview is called with the table variable captured from the initial mousedown event. If the document changes during the drag operation (e.g., collaborative editing), this table reference may become stale.

packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.ts
aaryan610: This works as expected.

Palanikannan1437 (approved):

packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.ts
aaryan610: It is being generated on mouse mouve event hence always has the updated content. I have verified it.

packages/editor/src/core/extensions/table/plugins/drag-handles/column/drag-handle.ts
aaryan610: We took this approach intentionally as we don't want to disturb the typing flow.

PR #8820, opened 2026-03-29
chore: replace MinIO with RustFS in docker-compose deployments

### LINKED ISSUE #8774: minio/minio image is now in maintenance-only mode, is there any plan to replace in docker-compose?
## Summary

The `minio/minio:latest` image used in [`deploy/selfhost/docker-compose.yml`](https://github.com/makeplane/plane/blob/develop/deploy/selfhost/docker-compose.yml) has entered a **maintenance-only stage** — meaning no new features will be developed, and only critical bug/security fixes may be applied going forward.

## Current usage in docker-compose.yml

```yaml
plane-minio:
  image: minio/minio:latest
  pull_policy: if_not_present
  restart: unless-stopped
  command: server /export --console-address ":9090"
```

## Question / Confirmation requested

Is the team aware of this change in MinIO's lifecycle status? Do you have a plan or timeline to:
- Continue using `minio/minio:latest` as-is (acknowledging the maintenance-only stance)?
- Or migrate to an actively maintained S3-compatible alternative (e.g., [SeaweedFS](https://github.com/seaweedfs/seaweedfs), [Garage](https://garagehq.deuxfleurs.fr/), or a pinned/stable MinIO release with a clear upgrade path)?

Since Plane self-hosted relies on this image for file storage, it would be good to have clarity on the long-term strategy so self-hosters can plan accordingly.

## References
- MinIO maintenance-only announcement: https://github.com/minio/minio (maintenance stage notice)

Rasaboun:
Closes #8774

## Summary

MinIO's community edition is [no longer maintained](https://github.com/minio/minio/blob/master/README.md) admin UI stripped Docker images discontinued and the GitHub repo is in maintenance mode since Feb 13 2026

This replaces "minio/minio" with [RustFS](https://github.com/rustfs/rustfs) (`rustfs/rustfs:1.0.0-alpha.90`) an S3-compatible drop-in alternative licensed under Apache 2.0

## What changed

- **`docker-compose.yml`** and **`deployments/cli/community/docker-compose.yml`**: swapped the image, updated env vars (`MINIO_ROOT_USER` → `RUSTFS_ACCESS_KEY`), changed volume mount (`/export` → `/data`), removed the MinIO `command`, added healthcheck (`/health` endpoint), added migration warning comments
- **`deployments/cli/community/install.sh`**: updated backup path from `/export` to `/data` to match the new RustFS volume mount
- **`.env.example`** and **`apps/api/.env.example`**: updated comments only
- **Admin console port changed from 9090 to 9001**: MinIO used `--console-address ":9090"`, RustFS serves its built-in console on port 9001 by default (this port was never published in the compose files, only accessible inside the Docker network or via manual port forwarding)

## What was intentionally NOT changed

- **Service name `plane-minio`**: kept for backward compatibility with `Caddyfile.ce`, `install.sh`, `swarm.sh`, and backup scripts
- **`USE_MINIO` / `MINIO_ENDPOINT_SSL` env vars**: consumed by the Python backend (`storage.py`), renaming would require code changes
- **Python backend code**: uses `boto3` (standard S3 SDK), works with any S3-compatible storage as-is
- **AIO deployment**: already uses `USE_MINIO=0` with external S3, doesn't include the storage container
- **`docker-compose-local.yml`** (local dev): still uses `minio/minio` with the `mc` bucket-init entrypoint — updating the local dev setup is left for a separate PR

## Scope

This targets **new deployments only**. Existing deployments should not upgrade in-place without migrating storage data first (MinIO and RustFS use different on-disk formats). A migration guide or script could be added in a follow-up.

## Longer-term suggestion

The storage layer could be made fully S3-agnostic by renaming `USE_MINIO` → `USE_LOCAL_S3` and removing product-specific naming throughout
This PR intentionally avoids that scope to keep things minimal and reviewable

### DIFF
--- a/deployments/cli/community/docker-compose.yml
+++ b/deployments/cli/community/docker-compose.yml
@@ -204,6 +204,9 @@ services:
       - rabbitmq_data:/var/lib/rabbitmq
 
   # Comment this if you using any external s3 compatible storage
+  # NOTE: RustFS uses a different on-disk format than MinIO.
+  # Existing MinIO deployments must migrate data before switching.
+  # Admin console is available on port 9001 (was 9090 with MinIO).
   plane-minio:
     image: rustfs/rustfs:1.0.0-alpha.90
     deploy:
@@ -214,6 +217,11 @@ services:
       <<: *storage-env
     volumes:
       - uploads:/data
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://127.0.0.1:9000/health"]
+      interval: 30s
+      timeout: 5s
+      retries: 3
 
   # Comment this if you already have a reverse proxy running
   proxy:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -140,6 +140,9 @@ services:
     volumes:
       - rabbitmq_data:/var/lib/rabbitmq
 
+  # NOTE: RustFS uses a different on-disk format than MinIO.
+  # Existing MinIO deployments must migrate data before switching.
+  # Admin console is available on port 9001 (was 9090 with MinIO).
   plane-minio:
     container_name: plane-minio
     image: rustfs/rustfs:1.0.0-alpha.90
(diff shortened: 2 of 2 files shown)


Rasaboun:
RustFS only has alpha releases right now (1.0.0-alpha.90 is the latest) so there's no stable tag to pin to

This is the most actively developed S3-compatible alternative with Apache 2.0 licensing, the other options are either AGPL (Garage) or heavier than what Plane needs (SeaweedFS/Ceph)

I pinned to the specific alpha tag rather than 'latest' so upgrades are explicit and I'll update the tag when a stable release drops

PR #8840, opened 2026-03-31
fix: filter out soft-deleted states from API endpoints

### LINKED ISSUE #8829: [bug]: Soft-deleted states returned by workspace and project state endpoints, causing deleted states to reappear in UI

### Current behavior

When a custom project state is deleted, it is correctly soft-deleted in the database (deleted_at is set). However, two API endpoints do not filter on `deleted_at__isnull=True`, so they return soft-deleted states alongside active ones. The frontend populates its state store from these endpoints, causing deleted states to reappear after navigating away and back to the page even though the delete appeared to succeed in the UI.

### Steps to reproduce

1. Go to Project Settings → States
2. Create a custom state (e.g. "Rainy Day")
3. Delete it. It disappears from the view
4. Navigate away (e.g. to the board view) and navigate back to Settings → States
5. "Rainy Day" reappears

### Environment

Production

### Browser

Google Chrome

### Variant

Self-hosted

### Version

stable v1.2.3

KanteshMurade:
Fixes #8829 (via issue endpoint filtering)

## Problem
Soft-deleted states were reappearing in the UI after navigation.

Although states were correctly soft-deleted in the database (`deleted_at` set), API responses were still returning them, causing the frontend to repopulate deleted states.

## Root Cause
The issue was initially suspected in the state API endpoints. However, `State.objects` already excludes soft-deleted records via the SoftDeletionManager.

The actual root cause was that issues referencing soft-deleted states were still being returned in the issue API responses.

## Solution
- Removed redundant `.filter(deleted_at__isnull=True)` from state endpoints
- Added filtering in issue queryset:
  `state__deleted_at__isnull=True`
- Ensured the serializer uses the filtered queryset (`issue_queryset`) for consistency

## Endpoint Clarification
Although the issue mentions state endpoints, the fix is applied in issue endpoints because they return state data indirectly through issue relationships.

## Scope
This change intentionally updates the issue endpoint (`IssueListEndpoint`) to ensure soft-deleted states are excluded from all API responses.

## Type of Change
- Bug fix

## Test Scenarios
- Create a state
- Delete the state (soft delete)
- Navigate away and return to the same page
- Verify the deleted state does NOT reappear

## Screenshots
N/A

## Result
- Soft-deleted states no longer reappear in the UI
- API responses are consistent and accurate


KanteshMurade:
@coderabbitai review

KanteshMurade:
Updated PR description and clarified scope. Ready for review.

sriramveeraghanta (approved):

PR #8848, opened 2026-04-03
fix: prevent shell injection in GHA feature-deployment workflow

NLmejiro:
### Summary

The `feature-deployment.yml` workflow directly interpolates `${{ github.* }}` context variables inside `run:` steps. Since these values are expanded server-side before the shell script executes, a user-controlled value (e.g., a branch name containing shell metacharacters) could be injected as shell commands.

### Affected code

`.github/workflows/feature-deployment.yml` — `set_env_variables` step:

```yaml
# Before (vulnerable)
run: |
  if [ "${{ github.event.inputs.base_tag_name }}" != "" ]; then
  FLAT_BRANCH_NAME=$(echo "${{ env.TARGET_BRANCH }}" | sed ...)
```

### Fix

Moved GitHub context variables to intermediate `env:` variables before using them in the shell script:

```yaml
# After (safe)
env:
  BASE_TAG_NAME: ${{ github.event.inputs.base_tag_name }}
  GH_TARGET_BRANCH: ${{ github.ref_name }}
run: |
  if [ "$BASE_TAG_NAME" != "" ]; then
  FLAT_BRANCH_NAME=$(echo "$GH_TARGET_BRANCH" | sed ...)
```

### References

- CWE-78: OS Command Injection
- [GitHub Actions Security Hardening](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections)

> This fix was identified through automated security scanning and manually verified.

### DIFF
--- a/.github/workflows/feature-deployment.yml
+++ b/.github/workflows/feature-deployment.yml
@@ -29,21 +29,24 @@ jobs:
     steps:
       - id: set_env_variables
         name: Set Environment Variables
+        env:
+          BASE_TAG_NAME: ${{ github.event.inputs.base_tag_name }}
+          GH_TARGET_BRANCH: ${{ github.ref_name }}
         run: |
           echo "BUILDX_DRIVER=docker-container" >> $GITHUB_OUTPUT
           echo "BUILDX_VERSION=latest" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


NLmejiro:
@CLAassistant check

NLmejiro:
@CLAassistant check

NLmejiro:
@CLAassistant check

PR #8924, opened 2026-04-21
🐛 Bug: Fix greeting timezone mismatch and add night greeting

MinitJain:
### Description

The home page greeting component had two bugs:

1. **Timezone mismatch**: `hour` (used to determine greeting text) was computed using the browser's local timezone, while `timeString` (displayed time) used `user.user_timezone` from the profile. These two could differ, producing a greeting like "Good morning" while showing 19:38.

2. **Missing night bucket**: `hour < 12` classified all midnight-to-noon hours as "morning", so users at 1 AM, 2 AM, 3 AM would see "Good morning".

**Root cause in one line:** the displayed time and greeting logic were reading time from two different clocks.

**Fix:**
- All four `Intl.DateTimeFormat` calls now use `timeZone: userTimezone` consistently
- Added timezone validation with try/catch — if `user_timezone` is an invalid IANA string, falls back to browser local time instead of throwing `RangeError` and crashing the component
- Added night threshold (0–4 AM → "Good night")
- Added `night` i18n key to all 19 locale files

**Greeting thresholds:**
| Hours | Greeting |
|-------|----------|
| 0–4   | Good night |
| 5–11  | Good morning |
| 12–17 | Good afternoon |
| 18–23 | Good evening |

Both `apps/web/core/components/home/user-greetings.tsx` and `apps/web/core/components/user/user-greetings.tsx` had identical bugs and were fixed identically.

### Type of Change

### Test Scenarios

- Set profile timezone to UTC, verify greeting text and displayed time both reflect UTC hour
- Set profile timezone to a valid non-local timezone (e.g. `America/New_York` while in IST), verify greeting and time both use New York time
- Set profile timezone to an invalid string, verify component does not crash and falls back to browser local time
- Log in between 00:00–04:59 in the active timezone, verify "Good night" appears
- Log in between 05:00–11:59, verify "Good morning"

## Summary by CodeRabbit

* **New Features**
  * Added a "night" greeting for late-night hours and consolidated greeting display into a single localized template that includes first and last name placeholders.

* **Bug Fixes**
  * Improved timezone handling with fallback to the browser timezone and consistent 24-hour time formatting.

* **Localization**
  * Added "night" and per‑daypart greeting entries across supported languages.

### DIFF
--- a/apps/web/core/components/home/user-greetings.tsx
+++ b/apps/web/core/components/home/user-greetings.tsx
@@ -25,7 +25,7 @@ export function UserGreetingsView(props: IUserGreetingsView) {
         {t(`greetings.${greeting}`, { first_name: user?.first_name ?? "", last_name: user?.last_name ?? "" })}
       </h2>
       <h5 className="flex items-center gap-2 font-medium text-placeholder">
-        <div>{greeting === "morning" ? "🌤️" : greeting === "afternoon" ? "🌥️" : "🌙"}</div>
+        <div aria-hidden="true">{greeting === "morning" ? "🌤️" : greeting === "afternoon" ? "🌥️" : "🌙"}</div>
         <div>
           {weekDay}, {date} {timeString}
         </div>
--- a/apps/web/core/components/home/user-greetings.tsx
+++ b/apps/web/core/components/home/user-greetings.tsx
@@ -5,6 +5,7 @@
  */
 
 // plane types
+import { useMemo } from "react";
 import { useTranslation } from "@plane/i18n";
 import type { IUser } from "@plane/types";
 // plane ui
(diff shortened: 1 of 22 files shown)


aaryan610:
Hey @MinitJain, thanks for fixing this bug! 🙌 I do have a small piece of feedback though- using “Good night” as a greeting might not be ideal since it’s usually used as a farewell rather than a greeting.

Let’s simplify the greetings to the following time ranges:

| Time | Greeting |
|--------|--------|
| 5:00-11:59 | Morning |
| 12:00-16:59 | Afternoon |
| 17:00-4:59 | Evening |

MinitJain:
Hey @aaryan610, thanks for the review : )
Makes total sense, "Good night" is definitely a farewell. 
Updated the thresholds to your suggestion (morning 5-11, afternoon 12-16, evening 17-4)
and removed the night bucket entirely. Also extracted the shared timezone and formatting logic into a useGreeting hook to avoid duplication between the two components.

MinitJain:
Hey @sriramveeraghanta
I have iterated on all CodeRabbit suggestions. Waiting for your approval. Let me know if you'd like any more changes!

MinitJain:
The PR fixes a timezone mismatch bug on the greeting component where the greeting text
and displayed time were reading from two different clocks. 
CodeRabbit's suggestions have all been addressed across multiple commits.

sriramveeraghanta:
@MinitJain Your changes looks good to me but the problem is `useCurrentTime` hook. Ideally at the place of greeting we just need to take the local time of the user not the current time from the timezone he setup.

MinitJain:
Hey @sriramveeraghanta 
Thanks for the feedback! 

I have removed the user_timezone from all Intl.DateTimeFormat calls in useGreeting. 
Now the greeting uses the browser's local timezone directly.

Also dropped the userTimezone validation logic since it's no longer needed.

PR #8935, opened 2026-04-26
fix(nginx): correct real_ip_header typo X-Forward-For → X-Forwarded-For

MinitJain:
### Description

Fixes #8934

All three nginx configs had a typo in the `real_ip_header` directive — `X-Forward-For` instead of the standard `X-Forwarded-For`. `X-Forward-For` is not a real HTTP header, so Nginx silently ignored the directive and never replaced `$remote_addr` with the actual client IP.

**Changes:**

- Fixed `real_ip_header X-Forward-For` → `real_ip_header X-Forwarded-For` in `apps/web/nginx/nginx.conf`
- Fixed `real_ip_header X-Forward-For` → `real_ip_header X-Forwarded-For` in `apps/admin/nginx/nginx.conf`
- Fixed `real_ip_header X-Forward-For` → `real_ip_header X-Forwarded-For` in `apps/space/nginx/nginx.conf`

One character added in 3 files, nothing else.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Feature (non-breaking change which adds functionality)
- [ ] Improvement (change that would cause existing functionality to not work as expected)
- [ ] Code refactoring
- [ ] Performance improvements
- [ ] Documentation update

### Test Scenarios

1. Deploy behind a proxy or CDN that sets `X-Forwarded-For`
2. Verify Nginx logs show the real client IP, not the proxy IP
3. Verify rate limiting applies per real client IP, not per proxy

### References

Closes #8934

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated server configurations across admin, space, and web applications to correctly identify real client IP addresses from incoming proxy headers. These changes standardize IP detection behavior across the platform, fixing inconsistencies and ensuring accurate and reliable client identification. All services now consistently process client IP information in a uniform manner.

PR #9019, opened 2026-05-05
fix: correct X-Forward-For typo to X-Forwarded-For in nginx configs

### LINKED ISSUE #8934: [bug]: real_ip_header directive has typo X-Forward-For instead of X-Forwarded-For in nginx configs

### Current behavior

All three nginx config files (`apps/web`, `apps/admin`, `apps/space`) have a typo in the `real_ip_header` directive:

```nginx
real_ip_header X-Forward-For;   ← wrong (not a real HTTP header)
```

`X-Forward-For` does not exist as an HTTP header. The correct standard header is `X-Forwarded-For`. Because of this typo, Nginx never replaces `$remote_addr` with the actual client IP, so rate limiting (`limit_req_zone $binary_remote_addr`) and IP logging always see the proxy/load balancer IP instead of the real client IP.

### Steps to reproduce

1. Deploy Plane behind a proxy or CDN (e.g. AWS ELB, Cloudflare)
2. Check the IP seen by the application — it will be the proxy IP, not the real client IP
3. Rate limiting (`limit_req_zone`) applies per-proxy instead of per-client, making it ineffective

### Environment

Self-hosted (Docker)

### Browser

N/A

### Variant

Community Edition / All

### Version

All

cybertron288:
Closes #8934

## What
Corrects the `real_ip_header` directive in `apps/web`, `apps/admin`, and `apps/space` nginx configs from the non-existent `X-Forward-For` header to the standard `X-Forwarded-For`.

## Why
`X-Forward-For` is not a real HTTP header, so nginx never replaces `$remote_addr` with the actual client IP. As a result, `limit_req_zone $binary_remote_addr` rate-limits per-proxy instead of per-client and access logs record the proxy IP. See #8934.

## Notes
- Opened as draft — please review before marking ready.
- Verification: changes are limited to three nginx config files; no app code touched. Did not run nginx locally.

### DIFF
--- a/apps/admin/nginx/nginx.conf
+++ b/apps/admin/nginx/nginx.conf
@@ -11,7 +11,7 @@ http {
 
   set_real_ip_from        0.0.0.0/0;
   real_ip_recursive       on;
-  real_ip_header          X-Forward-For;
+  real_ip_header          X-Forwarded-For;
   limit_req_zone          $binary_remote_addr zone=mylimit:10m rate=10r/s;
 
   access_log /dev/stdout;
--- a/apps/space/nginx/nginx.conf
+++ b/apps/space/nginx/nginx.conf
@@ -11,7 +11,7 @@ http {
 
   set_real_ip_from        0.0.0.0/0;
   real_ip_recursive       on;
-  real_ip_header          X-Forward-For;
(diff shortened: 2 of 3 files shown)


sriramveeraghanta:
This PR #8935 already fixes this.

PR #9063, opened 2026-05-13
feat(auth): Lark/Feishu OAuth provider + workspace directory invite

JOBYINC:
## What

Adds end-to-end Lark/Feishu support, on par with the existing Google/GitHub/GitLab/Gitea SSO integrations:

1. **OAuth provider** for sign-in via Lark (feishu.cn) and Lark Suite (larksuite.com). Configured under admin → authentication, with the same toggle + client_id / client_secret layout as the other providers.
2. **Invite from Lark** modal on the workspace members page. Workspace admins can browse the entire Feishu directory the app is authorised to see, multi-select teammates, pick a role, and add them directly — no email round-trip. Pre-created accounts match the same identifier the OAuth provider hands out on first sign-in, so SSO links seamlessly.
3. **Performance + i18n niceties** that came up while making this usable for a 500-person Feishu tenant.

## Why

Feishu / Lark is the de-facto chat + identity platform inside Chinese enterprises (50M+ DAU per ByteDance). Today self-hosted Plane deploys in those orgs either ship without SSO or rely on Google Workspace, which is blocked. With this PR they can stand Plane up next to Feishu with zero email round-trips.

## Commits

| # | Commit | What |
|---|--------|------|
| 1 | feat(auth): add Lark/Feishu OAuth provider | Backend OAuth provider + admin config + types |
| 2 | feat(auth/web): add "Sign in with Lark" button | Lark button on web + space sign-in pages |
| 3 | fix(auth/lark): unblock callback by removing LogRecord reserved-key collisions | Lark v2 token endpoint returns RFC 6749-style success; fixes `KeyError: msg` in extra={} |
| 4 | ci(docker): expose pnpm global bin so turbo install succeeds | pnpm 11 installs globals under `$PNPM_HOME/bin`, not `$PNPM_HOME` — unblocks frontend Docker build |
| 5 | feat(auth/lark): synthesize stable identifier when directory exposes no email | Some Feishu tenants don't surface `enterprise_email` via API — fall back to `<union_id>@lark.local` so SSO is still functional |
| 6 | feat(workspace/lark): contacts + batch invite endpoints | `/api/workspaces/<slug>/lark-contacts/` + `/lark-invite/` |
| 7 | feat(web/workspace): Invite-from-Lark modal | Multi-select directory picker; gated on `config.is_lark_enabled` + workspace-admin role |
| 8 | perf(lark-contacts): concurrent dept crawl + 10-min Redis cache | Cold 50s → Warm <5ms |
| 9 | feat(i18n): VITE_DEFAULT_LANGUAGE env + browser auto-detect | Priority chain: localStorage > env > navigator.languages > en. Backwards compatible. |

## Setup notes for reviewers

To exercise the full flow against a real Feishu tenant:

1. Create an internal app in https://open.feishu.cn → grab `App ID` + `App Secret`
2. Add the OAuth scopes: `contact:user.email:readonly contact:user.basic_profile:readonly`
3. For the directory picker, also grant tenant-token scopes: `contact:contact.base:readonly`, `contact:user.base:readonly`, `contact:user.email:readonly`, and set their data range to "All members"
4. Set `LARK_CLIENT_ID`, `LARK_CLIENT_SECRET`, `LARK_BASE_DOMAIN=feishu.cn` (or `larksuite.com`) on the API
5. Enable in admin → authentication

## Known caveats

- The synthetic `<union_id>@lark.local` email is necessary because Feishu's contact v3 API does not always expose `enterprise_email` (some tenants source the "Business Email" display from Feishu Mail rather than the contact record). Open to a different fallback shape if maintainers prefer.
- The contacts crawler runs serially within each top-level department subtree; concurrency is only at the top level. Could be deeper, but the 10-min cache makes cold paths rare in practice.
- i18n env var is opt-in (empty → browser detection only) so existing deploys see no behaviour change.

Marking this **draft** to give maintainers room to suggest the right shape before I polish anything else. Happy to split into smaller PRs (auth-only / invite-only / i18n) if that's easier to land.


ElvisZP (commented):
...

vihar:
Directory sync is part of Plane's paid plans (https://plane.so/pricing), so it isn't a change we can take into the community edition. Closing here, and appreciate the effort that went in.

PR #9094, opened 2026-05-16
fix(web): add requestIdleCallback fallback for Safari/iOS

bubacho:
## Summary
Fixes a runtime crash on Safari/iOS where `window.requestIdleCallback` is not available.

## Type of change
- [x] New feature
- [x] Breaking change
- [x] Documentation update

## Problem
On iPhone Safari, opening a project issues page can crash with:
`TypeError: window.requestIdleCallback is not a function`.

The crash happens in `RenderIfVisible` and can break project/issue views.

## Changes
- Added a small `runIdleTask` helper:
  - uses `window.requestIdleCallback` when available
  - falls back to `globalThis.setTimeout(..., 0)` otherwise
- Replaced direct `window.requestIdleCallback` calls with `runIdleTask`
- Fixed browser check to `typeof window !== "undefined"`
- Updated `IntersectionObserver` cleanup to unobserve a stable `target` reference
- Removed `children` from the observer effect dependency list to avoid unnecessary observer recreation

## Test scenarios
1. Open project issues page in iPhone Safari/iOS -> no crash, page renders normally.
2. Open the same page in Chrome/Firefox -> behavior unchanged.
3. Trigger scrolling/visibility updates -> lazy render behavior still works.

## Screenshots / media
N/A (runtime compatibility fix).

## Why this is safe
- No behavior change for browsers that support `requestIdleCallback`
- Graceful fallback for unsupported browsers
- Keeps lazy rendering logic intact while preventing runtime failures

## Summary by CodeRabbit

* **Performance**
  * Improved idle-time scheduling to reduce UI jank and provide smoother visibility updates for conditionally rendered content.
  * More efficient height measurements to prevent layout thrashing and improve scroll/column stability.

* **Bug Fixes**
  * Fixed intermittent visibility and measurement timing in board/column views so items render more reliably.

[Review Change Stack](https://app.coderabbit.ai/change-stack/makeplane/plane/pull/9094?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/web/core/components/core/render-if-visible-HOC.tsx
+++ b/apps/web/core/components/core/render-if-visible-HOC.tsx
@@ -23,6 +23,14 @@ type Props = {
   forceRender?: boolean;
 };
 
+const runIdleTask = (callback: () => void) => {
+  if (typeof window !== "undefined" && typeof window.requestIdleCallback === "function") {
+    window.requestIdleCallback(callback, { timeout: 300 });
+    return;
+  }
+  globalThis.setTimeout(callback, 0);
+};
+
 function RenderIfVisible(props: Props) {
   const {
     defaultHeight = "300px",
@@ -47,14 +55,13 @@ function RenderIfVisible(props: Props) {
 
   // Set visibility with intersection observer
   useEffect(() => {
-    if (intersectionRef.current) {
+    const target = intersectionRef.current;
+    if (target) {
       const observer = new IntersectionObserver(
         (entries) => {
           //DO no remove comments for future
-          if (typeof window !== undefined && window.requestIdleCallback && useIdletime) {
-            window.requestIdleCallback(() => setShouldVisible(entries[entries.length - 1].isIntersecting), {
-              timeout: 300,
-            });
+          if (typeof window !== "undefined" && useIdletime) {
+            runIdleTask(() => setShouldVisible(entries[entries.length - 1].isIntersecting));
           } else {
             setShouldVisible(entries[entries.length - 1].isIntersecting);
           }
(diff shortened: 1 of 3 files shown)


codingwolf-at:
@bubacho A few small suggestions while we're already touching this area:

* `useIdletime` could be renamed to `useIdleTime` for consistency/readability.
* Since the idle task is async/deferred, it may still execute after unmount or during rapid scroll updates. It might be worth storing the scheduled task id in a ref and cancelling it during cleanup via `cancelIdleCallback` (with `clearTimeout` fallback) to avoid stale state updates / queued tasks piling up.
* `runIdleTask` also feels generic enough to move into the shared `utils` folder with a small docstring explaining the function.

bubacho:
> @bubacho A few small suggestions while we're already touching this area:
> 
> * `useIdletime` could be renamed to `useIdleTime` for consistency/readability.
> * Since the idle task is async/deferred, it may still execute after unmount or during rapid scroll updates. It might be worth storing the scheduled task id in a ref and cancelling it during cleanup via `cancelIdleCallback` (with `clearTimeout` fallback) to avoid stale state updates / queued tasks piling up.
> * `runIdleTask` also feels generic enough to move into the shared `utils` folder with a small docstring explaining the function.

Thanks for the suggestions — addressed in 73c0d108b8.

Renamed useIdletime to useIdleTime.
Added cleanup cancellation for scheduled idle work (cancelIdleCallback with clearTimeout fallback) to prevent stale updates / queued tasks after unmount.
Moved runIdleTask to shared utils and added a short doc comment.

codingwolf-at (approved):

codingwolf-at (approved):

PR #9114, opened 2026-05-21
fix: add is_server parameter to S3Storage.__init__

### LINKED ISSUE #8680: [bug]: GET /api/v1/workspaces/<slug>/assets/<asset_id>/ returns 500 - S3Storage.__init__() does not accept is_server parameter

### Current behavior

The external API endpoint for fetching asset presigned URLs (`GenericAssetEndpoint.get()`) crashes with a `TypeError` on every request.

Root cause:

`GenericAssetEndpoint.get()` in `apps/api/plane/api/views/asset.py` calls:

`storage = S3Storage(request=request, is_server=True)`

But `S3Storage.__init__()` in `apps/api/plane/settings/storage.py` only accepts:

`def __init__(self, request=None):`

This raises `TypeError: S3Storage.__init__() got an unexpected keyword argument 'is_server'`, which is caught by the generic exception handler and returned as `{"error": "Internal server error"}` (HTTP 500).

The same issue occurs in `GenericAssetEndpoint.post()` and `GenericAssetEndpoint.patch()` — all three  methods pass `is_server=True`.

### Steps to reproduce

`curl -H "X-API-Key: <your_api_key>" "https://<instance>/api/v1/workspaces/<slug>/assets/<any_valid_asset_id>/"`

Returns: `{"error": "Internal server error"}` (500)

### Environment

Production

### Browser

None

### Variant

Self-hosted

### Version

v1.2.2

HermessNRJ:
### Description

asset.py calls S3Storage(request=request, is_server=True) in three places (lines 337, 441, 565), but S3Storage.__init__ in plane/settings/storage.py only accepts request=None — the is_server parameter was never added.

This causes a TypeError: S3Storage.__init__() got an unexpected keyword argument 'is_server' on every request that hits those asset endpoints, resulting in HTTP 500 errors for all asset operations (upload slot creation, presigned URL generation, attachment downloads) when USE_MINIO=1.

The fix adds is_server=False to __init__ and updates the endpoint_url condition so server-side calls use AWS_S3_ENDPOINT_URL instead of deriving the host from the HTTP request object.

### Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)

N/A

### Test Scenarios 

- Reproduced the TypeError on Plane 1.3.1 with USE_MINIO=1 / RustFS or Minio storage by hitting GET /api/v1/workspaces/{slug}/assets/{id}/
- Applied the patch to a running pod and confirmed asset upload and download endpoints return 200
- Verified that client-side presigned URL generation (where request is provided and is_server=False) still derives the endpoint from the request host as before

### References

Fixes #8680

## Summary by CodeRabbit

* **Chores**
  * Enhanced S3/MinIO storage configuration handling to support improved endpoint URL resolution for server-side operations, increasing flexibility in storage setup configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/makeplane/plane/pull/9114?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/api/plane/settings/storage.py
+++ b/apps/api/plane/settings/storage.py
@@ -22,7 +22,7 @@ class S3Storage(S3Boto3Storage):
 
     """S3 storage class to generate presigned URLs for S3 objects"""
 
-    def __init__(self, request=None):
+    def __init__(self, request=None, is_server=False):
         # Get the AWS credentials and bucket name from the environment
         self.aws_access_key_id = os.environ.get("AWS_ACCESS_KEY_ID")
         # Use the AWS_SECRET_ACCESS_KEY environment variable for the secret key
@@ -48,7 +48,7 @@ class S3Storage(S3Boto3Storage):
                 aws_access_key_id=self.aws_access_key_id,
                 aws_secret_access_key=self.aws_secret_access_key,
                 region_name=self.aws_region,
-                endpoint_url=(f"{endpoint_protocol}://{request.get_host()}" if request else self.aws_s3_endpoint_url),
+                endpoint_url=(f"{endpoint_protocol}://{request.get_host()}" if request and not is_server else self.aws_s3_endpoint_url),
                 config=boto3.session.Config(signature_version="s3v4"),
             )
         else:


ch4og:
Applied this patch and it fixed the issue completely. Looking forward to seeing this PR merged!

@sriramveeraghanta (sorry if you are the wrong person to ping), but could you please take a look at this PR? It looks like a clean fix without any side effects.

HermessNRJ:
Could you check this one @pablohashescobar / @dheeru0198 ?

misterkuka:
Confirmed still needed on **v1.3.1** / current master.

We hit this in production on self-hosted MinIO (`USE_MINIO=1`):

- Browser `/api/assets/v2/...` uploads work
- PAT `POST /api/v1/workspaces/{slug}/assets/` → 500 with  
  `TypeError: S3Storage.__init__() got an unexpected keyword argument 'is_server'`  
  at `api/views/asset.py:565`

This PR’s 2-line change is the right fix. Would appreciate a review/merge so CE can use the public assets API for automation (issue attachments, migrations, etc.).

Refs #8680

PR #9239, opened 2026-06-15
fix: strip whitespace and handle null values in instance configuration

okxint:
## What

The instance configuration PATCH endpoint stored raw values from `request.data` without sanitization, causing two issues:

1. **Null values** — `str(None)` produces the literal string `"None"` instead of an empty string, so clearing a field would store `"None"` in the DB.
2. **Whitespace** — values with leading/trailing spaces (e.g. copy-pasted URLs or tokens) were stored as-is, which can silently break downstream usage.

## Fix

```python
raw_value = request.data.get(configuration.key, configuration.value)
value = "" if raw_value is None else str(raw_value).strip()
```

Both issues are addressed in a single sanitization step before the value is encrypted or stored.

## Files changed

- `apps/api/plane/license/api/views/configuration.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed configuration value normalization to correctly handle null inputs, convert boolean values to `"1"`/`"0"`, and trim leading/trailing whitespace before storing values.

### DIFF
--- a/apps/api/plane/license/api/views/configuration.py
+++ b/apps/api/plane/license/api/views/configuration.py
@@ -45,7 +45,8 @@ class InstanceConfigurationEndpoint(BaseAPIView):
 
         bulk_configurations = []
         for configuration in configurations:
-            value = request.data.get(configuration.key, configuration.value)
+            raw_value = request.data.get(configuration.key, configuration.value)
+            value = "" if raw_value is None else str(raw_value).strip()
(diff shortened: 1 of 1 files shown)


okxint:
Fixed — boolean values are now converted to '1'/'0' before the string cast, so EMAIL_USE_TLS/SSL sent as actual booleans from the frontend will correctly match the equality checks downstream.

PR #9245, opened 2026-06-16
[GIT-248 | GIT-254] refactor: store and components consolidation to core

codingwolf-at:
### Description
Moved Components from `apps/web/ce/components` to `apps/web/core/components`

### Type of Change
- [x] Code refactoring


prateekshourya29:
@coderabbitai, review these changes. Most of the files have just been moved, so ignore them and review the rest. We are also cleaning things up as part of this refactor.

apps/web/tsconfig.json
```diff
@@ -15,6 +15,7 @@
     "noUnusedLocals": false,
     "noImplicitReturns": false,
     "noImplicitOverride": false,
+    "lib": ["DOM", "DOM.Iterable", "ES2023"],
```
prateekshourya29: Why is this needed?

codingwolf-at: Removed the ES2023 lib override. The only thing needing it was a .toSorted() call in tab-navigation-root.tsx, which I swapped for .sort() — the array comes straight from .filter(), so it's a fresh copy and sorting in place is safe. tsconfig now inherits ES2022 from the base config, and typecheck passes clean.

prateekshourya29 (approved):

PR #9253, opened 2026-06-17
Feat/OIDC free auth

MagicRB:
### Description

This PR implements a generalized OIDC auth backend which is largely a copy-paste from the Gitea on. This needs cleanup and feedback on if these changes are desired, opening for the sake of sharing.

### Type of Change

- [x] Code refactoring
- [ ] Documentation update


vihar:
Thanks for opening this to share the idea. OIDC is part of Plane's paid plans (https://plane.so/pricing), so it isn't something we'd take into the community edition. Closing it here, and thanks for the thought behind it.

MagicRB:
Please do tell your customers that you're making them pay for something that can be trivially implemented and has in fact been implemented already, I'm sure they would appreciate it. Perhaps if it is this easy for me to implement one of your paid features, it shouldn't be a paid feature? I am very disappointed this is the approach here.

In any case, we decided not self-host plane as it became obvious it's not meant to be self-hosted.

PR #9261, opened 2026-06-18
fix: Scope workspace user preference filter to current user

### LINKED ISSUE #9260: [bug]: Sidebar pin/unpin not persisted — WorkspaceUserPreference PATCH missing user filter
## Current behavior
Sidebar pin/unpin changes return 200 OK but are not saved for the requesting user. After page refresh, pins revert to previous state.

## Root cause
In `plane/app/views/workspace/user_preference.py`, the `patch` method filters preferences without scoping to `request.user`:

```python
# v1.3.1 — missing user filter
preference = WorkspaceUserPreference.objects.filter(
    key=key, workspace__slug=slug
).first()
```

In a workspace with multiple members, this returns the first matching record regardless of user — so the wrong user's preference gets updated.

## Fix
Add `user=request.user` to the filter:

```python
preference = WorkspaceUserPreference.objects.filter(
    key=key, workspace__slug=slug, user=request.user
).first()
```

## Steps to reproduce
1. Self-host Plane v1.3.1 with Docker
2. Create a workspace with 2+ members
3. Log in as any user
4. Pin a sidebar item (e.g. Analytics, Archives)
5. Refresh the page → pin reverts

## Verified
- DB shows `is_pinned: False` after PATCH despite 200 OK response
- Adding `user=request.user` to the filter fixes the issue permanently

## Version
v1.3.1 (self-hosted Docker)

okxint:
The `patch` method in `WorkspaceUserPreference` was filtering by `key` and `workspace__slug` only. In a workspace with multiple users, `.first()` could grab a different user's preference row, so updating pin/sort state would either mutate the wrong user's data or leave the current user's preference unchanged — making sidebar state look like it didn't persist after refresh.

The `get` method already had `user=request.user` on its queries. Added the same filter to `patch`.

Closes #9260

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where personal workspace preferences were not properly isolated between users within the same workspace.

Closes #9260

### DIFF
--- a/apps/api/plane/app/views/workspace/user_preference.py
+++ b/apps/api/plane/app/views/workspace/user_preference.py
@@ -85,7 +85,7 @@ class WorkspaceUserPreferenceViewSet(BaseAPIView):
             if not key:
                 continue
 
-            preference = WorkspaceUserPreference.objects.filter(key=key, workspace__slug=slug).first()
+            preference = WorkspaceUserPreference.objects.filter(key=key, workspace__slug=slug, user=request.user).first()
 
             if not preference:
                 continue


apps/api/plane/app/views/workspace/user_preference.py
```diff
@@ -85,7 +85,7 @@ def patch(self, request, slug):
             if not key:
                 continue
 
-            preference = WorkspaceUserPreference.objects.filter(key=key, workspace__slug=slug).first()
+            preference = WorkspaceUserPreference.objects.filter(key=key, workspace__slug=slug, user=request.user).first()
```
sviat-barbutsa, line 88: Nice fix. Could you, please, add a regression test for the cross-user case?

The nuance: Before this change, two users in the same workspace with the same preference key could hit `filter(key=key, workspace__slug=slug).first()`, so PATCH might update another user’s preference row.

For this case will be helpful a test where user A and user B both have `key="views"` preferences, then user B patches `sort_order` or `is_pinned`, would lock in that only user B’s row changes.

sriramveeraghanta:
Udpated some changes and merged your changes in #9279

PR #9262, opened 2026-06-18
fix: Use APP_DOMAIN env var for bot user email

### LINKED ISSUE #9251: [bug]: `plan.so` is hardcoded into bot user email

### Current behavior

When Plane creates the initial bot user called `Plane`, it sets it's email to `bot_user_{workspace_slug}@plane.so`, which is incorrect as `plane.so` is not the domain of our self-hosted instance.

### Steps to reproduce

Run Plane on your own infrastructure and create a workspace, then check the DB.

### Suggested fix

Plane would need to know it's own domain, the environment variable configuration system you have already understands `APP_DOMAIN`, if you plumb that through into the `api` django application, that could be easily used for generating the emails. I'd do this myself, but I don't know the codebase that well and it's not a blocker so I'm not feeling like spending too much time on this (when someone else can do this in 5 minutes).

### Environment

Production

### Browser

Mozilla Firefox

### Variant

Self-hosted

### Version

v1.3.1

okxint:
Bot user email was hardcoded as `bot_user_{id}@plane.so` during workspace seeding. Self-hosted instances with a custom `APP_DOMAIN` would end up with bot accounts on the wrong domain.

Replaced the hardcoded domain with `os.environ.get('APP_DOMAIN', 'plane.so')` — defaults to `plane.so` so Plane Cloud is unaffected.

Closes #9251

## Summary by CodeRabbit

* **Chores**
  * Enhanced workspace initialization to derive the bot user’s seeded email domain from the configured web URL, with a sensible fallback to the default domain when not provided.

### DIFF
--- a/apps/api/plane/bgtasks/workspace_seed_task.py
+++ b/apps/api/plane/bgtasks/workspace_seed_task.py
@@ -526,7 +526,7 @@ def workspace_seed(workspace_id: uuid.UUID) -> None:
             last_name="",
             is_bot=True,
             bot_type=BotTypeEnum.WORKSPACE_SEED,
-            email=f"bot_user_{workspace.id}@plane.so",
+            email=f"bot_user_{workspace.id}@{os.environ.get('APP_DOMAIN', 'plane.so')}",
             password=make_password(uuid.uuid4().hex),
             is_password_autoset=True,
         )
--- a/apps/api/plane/bgtasks/workspace_seed_task.py
+++ b/apps/api/plane/bgtasks/workspace_seed_task.py
@@ -10,6 +10,7 @@ import uuid
 from typing import Dict
 import logging
 from datetime import timedelta
+from urllib.parse import urlparse
 
 # Django imports
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


MagicRB:
Shouldn't this go through the proper configuration system that plane has? reading it directly from the environment there is very side effectful and possibly unexpected. As far as I am aware `APP_DOMAIN` wasn't used by anything except to set other configuration options outside of plane (in the surrounding bash scripting). As such some folks might assume it's safe to remove post initialization as, again afaik plane won't unset configuration variables, when it doesn't see the accompanying environment variable. It shouldn't affect anything as this code should only be ran on instance init, but it still feels super hacky.

okxint:
good point — updated to use `settings.WEB_URL` instead of reading `APP_DOMAIN` directly from the environment. parses the hostname out of the URL via `urlparse` and falls back to `plane.so` if `WEB_URL` isn't set. goes through the config system properly now.

sriramveeraghanta (approved):

PR #9296, opened 2026-06-23
[WEB-7847] fix: enforce workspace membership on entity-search endpoint

mguptahub:
## Summary

- `SearchEndpoint` (`/workspaces/<slug>/entity-search/`) required authentication but did **not** verify the requesting user was a member of the queried workspace
- Any authenticated Plane user could enumerate members of workspaces they don't belong to by knowing or guessing the workspace slug
- Fixes **GHSA-32q3-mqpc-3mhv** (medium — cross-workspace member enumeration)

## Fix

Added a `WorkspaceMember` guard at the top of `SearchEndpoint.get()` — returns `403 Forbidden` if the requesting user is not an active member of the target workspace.

The EE version already had this protection via `@can(WorkspacePermissions.VIEW)`. This brings OSS to parity.

**Files changed:** `apps/api/plane/app/views/search/base.py` (+10 lines)

## Test plan

- [ ] Authenticated user queries `/workspaces/<own-slug>/entity-search/` → `200 OK` with results
- [ ] Authenticated user queries `/workspaces/<other-slug>/entity-search/` for a workspace they don't belong to → `403 Forbidden`
- [ ] Unauthenticated request → `401 Unauthorized` (unchanged, handled by `BaseAPIView`)

## Summary by CodeRabbit

* **Security / Bug Fixes**
  * Updated search access controls to require proper workspace user authorization.  
  * Requests to the search feature are now restricted to active workspace members, helping prevent unauthorized access to workspace data.

### DIFF
--- a/apps/api/plane/app/views/search/base.py
+++ b/apps/api/plane/app/views/search/base.py
@@ -28,6 +28,7 @@ from rest_framework.response import Response
 
 # Module imports
 from plane.app.views.base import BaseAPIView
+from plane.app.permissions import WorkspaceUserPermission
 from plane.db.models import (
     Workspace,
     Project,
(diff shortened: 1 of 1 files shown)


apps/api/plane/app/views/search/base.py
```diff
@@ -303,6 +303,16 @@ def get(self, request, slug):
...
     def get(self, request, slug):
+        # Verify the requesting user is an active member of the target workspace.
+        # Without this guard any authenticated user can enumerate members of
+        # workspaces they do not belong to (GHSA-32q3-mqpc-3mhv).
+        if not WorkspaceMember.objects.filter(
+            member=request.user,
+            workspace__slug=slug,
+            is_active=True,
+        ).exists():
+            return Response({"error": "Forbidden"}, status=status.HTTP_403_FORBIDDEN)
```
dheeru0198:
```suggestion
class SearchEndpoint(BaseAPIView):
    permission_classes = (WorkspaceUserPermission, )
    
    def get(self, request, slug):
```

Instead of doing inline, please use the existing permission class
@mguptahub

mguptahub: Done. Swapped the inline check for `permission_classes = (WorkspaceUserPermission,)` on `SearchEndpoint`. `WorkspaceUserPermission` reads `view.workspace_slug` which resolves to `self.kwargs.get("slug")` — matching the URL kwarg — so the enforcement is identical, just in the right layer.

dheeru0198 (approved):

PR #9324, opened 2026-06-26
fix(cloudflare): route auth to worker and add post-login workspace APIs

KunanonJ:
test

Made with [Cursor](https://cursor.com)


KunanonJ:
Created by mistake against upstream; reopening on mygogocash/plane.

Issue #7022, opened 2025-05-07
[feature]: Japanese IME Input Conflicts with Label Submission via Enter Key


Aethey:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Summary

First of all, we really enjoy using Plane — our team finds it very useful and well-designed. However, we’ve encountered a small issue when using Japanese input (IME).

Currently, when creating or editing labels, pressing the Enter key immediately submits the label. But in Japanese input, Enter is used to confirm a kanji conversion, and Space is used to switch candidates. As a result, it's very easy to accidentally submit an incomplete or incorrect label while typing in Japanese.

Would it be possible to add an option or setting to prevent Enter from instantly submitting the label? For example, requiring an extra confirmation button or allowing users to choose their preferred key behavior?

This small change would greatly improve the usability for teams using Japanese or other IME-based input systems. Thanks a lot!

### Why should this be worked on?

This issue significantly affects users who type in Japanese (and other IME-based languages) when editing or creating labels.
Since the Enter key is used to confirm character conversion in Japanese, the current behavior causes unintended submissions, making it frustrating and error-prone.

Improving this would greatly enhance usability and accessibility for international teams. It’s a small change but would remove a daily pain point for many non-English-speaking users.

gijigae:
Raised #7084 to mitigate the issue.

gijigae:
hello @vihar ! Would you please review #7084 ? The fix will greatly help Japanese users adopt Plane.

PR #9331, opened 2026-06-28
fix(ui): ignore Enter during IME composition in select dropdowns

mahirhir:
### Description

`MultiSelectDropdown` and `SingleSelectDropdown` attach their `Combobox` `onKeyDown` to `useDropdownKeyPressed`, which calls `onEnterKeyDown` (`toggleDropdown`) on every `Enter`. The search input inside the dropdown only stops propagation for `Escape`, so the `Enter` that confirms an IME composition bubbles up from the input to the combobox and toggles the dropdown closed before anything is selected.

For a CJK IME (Japanese, Chinese, Korean), the first `Enter` is meant to commit the conversion, not to act on the dropdown. As a result the search field is hard to use with an IME: committing the text closes the menu.

The sibling hook `useDropdownKeyDown` already guards against this with `!event.nativeEvent.isComposing`. This change applies the same guard to `useDropdownKeyPressed`. Non-IME `Enter` is unaffected (`isComposing` is `false`), so it still toggles as before; only the `Enter` that ends a composition is ignored.

### Type of Change

### Test Scenarios

With an IME enabled (for example Japanese):

1. Open a dropdown backed by `MultiSelectDropdown` or `SingleSelectDropdown` with search enabled.
2. Type into the search field and press `Enter` to confirm the IME conversion.
3. Before this change the dropdown closes. After it, the conversion is committed and the dropdown stays open.

Without an IME, pressing `Enter` still toggles the dropdown exactly as before.

I verified by inspection that this matches the existing `useDropdownKeyDown` guard and leaves non-composing `Enter` unchanged. I don't have a CJK IME configured locally to run the end-to-end flow, so a confirmation from someone who does would be welcome.

### References

The `!event.nativeEvent.isComposing` guard already exists in `useDropdownKeyDown` (`packages/ui/src/hooks/use-dropdown-key-down.tsx`) and in the label create handlers; this brings `useDropdownKeyPressed` in line. IME Enter handling has come up before in Plane (for example #7022).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard behavior in dropdowns by preventing the Enter key action from firing during text input composition, helping users with IME input avoid accidental selections.

### DIFF
--- a/packages/ui/src/hooks/use-dropdown-key-pressed.ts
+++ b/packages/ui/src/hooks/use-dropdown-key-pressed.ts
@@ -27,7 +27,7 @@ export const useDropdownKeyPressed: TUseDropdownKeyPressed = (onEnterKeyDown, on
 
   const handleKeyDown = useCallback(
     (event: React.KeyboardEvent<HTMLElement>) => {
-      if (event.key === "Enter") {
+      if (event.key === "Enter" && !event.nativeEvent.isComposing) {
         stopEventPropagation(event);
         onEnterKeyDown();
       } else if (event.key === "Escape") {


mahirhir:
Closing this to clean up my open PRs. It's a small change and not really worth taking up your review time. Sorry for the noise.

Issue #9280, opened 2026-06-21
[bug]: Unable to use special characters such as dashes in project title


tgishor:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

When creating or updating a project title, the application does not allow special characters such as dashes/hyphens (-). This prevents users from entering normal project names that include separators or descriptive formatting.

### Steps to reproduce

1. Go to the project creation or project edit page.
2. Enter a project title that includes a dash/hyphen, for example:
    AI-Task-Tracker
    Project Dashboard - Phase 1
3. Try to save/create the project.
4. The title is rejected or the project cannot be saved.

### Environment

Production

### Browser

Google Chrome

### Variant

Local

### Version

v1.3.1

karthiksuki:
"PROJECT_NAME_CANNOT_CONTAIN_SPECIAL_CHARACTERS", @tgishor

harsha1196:
We've identified the issue and will get it fixed.

sriramveeraghanta:
Moving the conversation to #9226

PR #9341, opened 2026-07-01
fix: allow special characters in project name

### LINKED ISSUE #9226: [bug]: Project creation fails with special characters in project name

### Current behavior

When I try to create a project in my self-hosted Plane instance, the request fails with HTTP 400 and the API returns:
{"name":["PROJECT_NAME_CANNOT_CONTAIN_SPECIAL_CHARACTERS"]}

The frontend then shows React errors and the project is not created.

I expected the project to be created, or at least for the UI to show a clear validation message before sending the request.

### Steps to reproduce

1. Go to the Projects page in a workspace.
2. Click Create project.
3. Enter a name containing a special character, for example: Career Scaling & Remote Transition.
4. Submit the form.
5. The API responds with 400 and the project is not created.

### Environment

Production

### Browser

Mozilla Firefox

### Variant

Self-hosted

### Version

v1.3.1

Suryacodeshere:
Resolves #9226 and #9280.

### Description
The project name (title) is a user-facing label and should be allowed to contain special characters such as hyphens -, ampersands &, etc. However, the serializers in the backend API and App endpoints were validating the project 
ame against Project.FORBIDDEN_IDENTIFIER_CHARS_PATTERN (which is meant only for project identifiers to ensure they are valid prefix patterns). This caused project creation and update requests containing special characters in their name to fail with PROJECT_NAME_CANNOT_CONTAIN_SPECIAL_CHARACTERS.

This change removes the validation of project names against the forbidden identifier characters regex, while keeping the validation in place for project identifiers.

### Changes
- Removed name validation using Project.FORBIDDEN_IDENTIFIER_CHARS_PATTERN from ProjectCreateSerializer, ProjectUpdateSerializer, and ProjectSerializer in pps/api/plane/api/serializers/project.py.
- Removed name validation using Project.FORBIDDEN_IDENTIFIER_CHARS_PATTERN from ProjectSerializer.validate_name in pps/api/plane/app/serializers/project.py.

## Summary by CodeRabbit

* **Bug Fixes**
  * Project names no longer reject certain special characters during creation or editing.
  * Identifier validation remains unchanged, so project identifiers still follow the existing character rules.

### DIFF
--- a/apps/api/plane/api/serializers/project.py
+++ b/apps/api/plane/api/serializers/project.py
@@ -105,11 +105,7 @@ class ProjectCreateSerializer(BaseSerializer):
         ]
 
     def validate(self, data):
-        project_name = data.get("name", None)
         project_identifier = data.get("identifier", None)
-
-        if project_name is not None and re.match(Project.FORBIDDEN_IDENTIFIER_CHARS_PATTERN, project_name):
-            raise serializers.ValidationError("Project name cannot contain special characters.")
 
         if project_identifier is not None and re.match(Project.FORBIDDEN_IDENTIFIER_CHARS_PATTERN, project_identifier):
             raise serializers.ValidationError("Project identifier cannot contain special characters.")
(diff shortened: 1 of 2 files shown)


Suryacodeshere:
@dheeru0198 @pablohashescobar Whenever you have time. Could you please review it? Thank You!

PR #9358, opened 2026-07-05
fix: add requestIdleCallback polyfill at app entry for iOS WebKit (Fixes #9231)

### LINKED ISSUE #9231: iOS WebKit: gantt-layout-loader throws on missing window.requestIdleCallback — global error boundary on every issues-list view
### Description

Plane CE's lazy-loaded `gantt-layout-loader` chunk calls `window.requestIdleCallback` without feature detection or a polyfill. iOS WebKit (Safari, iOS Chrome, iOS Firefox — all WebKit-backed under Apple's App Store rules) has never implemented `requestIdleCallback` — open WebKit gap since 2016 ([webkit.org#164193](https://bugs.webkit.org/show_bug.cgi?id=164193)). The throw propagates up, React's error boundary catches it, and the user sees the generic "Looks like something went wrong" page on every issues-list / project view.

### Steps to reproduce

1. iPad (any iOS / iPadOS version), Chrome (CriOS user-agent — but iOS Safari and iOS Firefox would all fail the same way; see "Why the matrix is weird" below).
2. Sign in to a Plane CE instance.
3. Navigate to any project's `/projects/<uuid>/issues/` page.
4. The full error boundary fires.

### Console error (from Safari Web Inspector against the iPad)

```
TypeError: window.requestIdleCallback is not a function.
(In 'window.requestIdleCallback(()=>{h.current&&(S.current=`${h.current.offsetHeight}px`)})', 'window.requestIdleCallback' is undefined)
— gantt-layout-loader-BH8Iumhn.js:1:5524

React Router caught the following error during render
— componentDidCatch (chunk-EPOLDU6W-PsE1YSAx.js:4:2496)

[also] Minified React error #418 (hydration failed) and #423 (client-side recovery also failed)
— both downstream of the same throw.
```

### Why the matrix is weird (and the cause is clear)

| Browser | Result | Why |
|---|---|---|
| **iOS Chrome (CriOS)** | Error boundary | UA hints route through the bundle that lazy-loads gantt-layout-loader → hits missing API |
| **iOS Safari** | Works | UA routes through a different bundle that doesn't load the gantt chunk |
| **iOS Chrome with "Request Desktop Site"** | Works | Desktop UA routes through the desktop bundle (gantt chunk apparently feature-detects or skips on the desktop layout calculator path) |
| **macOS Safari / macOS Chrome** | Works | Engines implement `requestIdleCallback` natively |

So the bug has been sitting hidden behind a UA-sniffing branch that only one of three iOS browser paths exercises.

### Suggested fix

Either:

**A — Polyfill `requestIdleCallback` once at app entry:**

```javascript
if (typeof window.requestIdleCallback !== 'function') {
  window.requestIdleCallback = (cb) => {
    const start = Date.now();
    return setTimeout(() => cb({
      didTimeout: false,
      timeRemaining: () => Math.max(0, 50 - (Date.now() - start)),
    }), 1);
  };
  window.cancelIdleCallback = (id) => clearTimeout(id);
}
```

**B — Feature-detect inside `gantt-layout-loader` and fall back to `setTimeout(fn, 0)` for the height-measurement use case** (which doesn't actually need idle-time semantics; it's just deferring a layout measurement).

Either is a few lines. (A) catches every other `requestIdleCallback` use in the codebase.

### Environment

- Plane CE v1.3.1 (self-hosted, full upstream image set)
- Client: iPad (any), Chrome on iOS / iOS WebKit

rohanmaan07:
### Description
This PR fixes a bug where the `gantt-layout-loader` chunk throws an error on iOS WebKit devices (Safari, iOS Chrome, iOS Firefox) due to the absence of the native `window.requestIdleCallback` API. 

The missing API caused the React Error Boundary to catch the thrown error, resulting in a generic "Looks like something went wrong" page whenever users navigated to project views containing the layout loader on iPad/iOS devices.

### Changes Made
- Added a fallback polyfill for `window.requestIdleCallback` and `window.cancelIdleCallback` directly at the app entry points (`apps/web/app/entry.client.tsx`, `apps/admin/app/entry.client.tsx`, `apps/space/app/entry.client.tsx`).
- The polyfill safely checks for the function's existence and uses `setTimeout` to mimic idle time deferral for layout measurement, completely preventing the crash on iOS WebKit environments.

Fixes #9231

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved browser compatibility by adding a fallback for idle callback scheduling when the native browser support is unavailable.
  * This helps pages load and hydrate more reliably in more environments without changing the app experience.

### DIFF
--- a/apps/admin/app/entry.client.tsx
+++ b/apps/admin/app/entry.client.tsx
@@ -8,6 +8,19 @@ import { startTransition, StrictMode } from "react";
 import { hydrateRoot } from "react-dom/client";
 import { HydratedRouter } from "react-router/dom";
 
+if (typeof window !== "undefined" && typeof window.requestIdleCallback !== "function") {
+  window.requestIdleCallback = (cb: any) => {
+    const start = Date.now();
+    return setTimeout(() => {
+      cb({
+        didTimeout: false,
+        timeRemaining: () => Math.max(0, 50 - (Date.now() - start)),
+      });
+    }, 1) as any;
+  };
+  window.cancelIdleCallback = (id: any) => clearTimeout(id);
+}
+
 startTransition(() => {
   hydrateRoot(
     document,
--- a/apps/space/app/entry.client.tsx
+++ b/apps/space/app/entry.client.tsx
@@ -8,6 +8,19 @@ import { startTransition, StrictMode } from "react";
 import { hydrateRoot } from "react-dom/client";
 import { HydratedRouter } from "react-router/dom";
 
+if (typeof window !== "undefined" && typeof window.requestIdleCallback !== "function") {
+  window.requestIdleCallback = (cb: any) => {
+    const start = Date.now();
+    return setTimeout(() => {
+      cb({
+        didTimeout: false,
+        timeRemaining: () => Math.max(0, 50 - (Date.now() - start)),
+      });
+    }, 1) as any;
+  };
+  window.cancelIdleCallback = (id: any) => clearTimeout(id);
+}
+
 startTransition(() => {
   hydrateRoot(
     document,
--- a/apps/web/app/entry.client.tsx
+++ b/apps/web/app/entry.client.tsx
@@ -12,6 +12,18 @@ import polyfills from "@/lib/polyfills";
 
 void polyfills;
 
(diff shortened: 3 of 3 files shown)


MarcoMatta:
news?

PR #9367, opened 2026-07-07
[WEB-8060] fix(security): enforce authz on is_active member (de)activation

mguptahub:
## Summary

Fixes **GHSA-hpgm-9r34-c4x5** (HIGH) / **GHSA-25gg-cxm8-g7h9** (MEDIUM) — the same vulnerability under two severity framings.

`ProjectMemberViewSet.partial_update` nested **every** authorization guard inside `if "role" in request.data:`. Because `ProjectMemberSerializer` exposes `is_active` through `fields="__all__"` with no `read_only`, a request that sets `{"is_active": false}` while **omitting** `role` skips all guards. The `@allow_permission` decorator admits GUEST, and the only unconditional check blocks editing *your own* record — so a project **GUEST** could deactivate any other member, including **ADMINs**, leaving themselves the sole active participant = full project takeover.

**Endpoint:** `PATCH /api/workspaces/<slug>/projects/<project_id>/members/<pk>/`

## Fix

Add an independent `is_active` guard, placed right after the existing `role` block and mirroring both it and `destroy()`:

- only a project **admin** (or workspace admin) may (de)activate a member;
- a member whose role is **equal to or higher** than the requester's own cannot be (de)activated (matching the role block's `>=` comparison within the same method).

Workspace admins retain the intended bypass. Note: this endpoint resolves the target member with `is_active=True`, so it only governs deactivation of an *active* member (and no-op `is_active: true` on one already active); genuine reactivation of an inactive member is not reachable here — that path is the admin-gated member add/`create` flow.

## Testing

New contract regression tests — `apps/api/plane/tests/contract/app/test_project_member_is_active_authz.py`:

- `test_guest_cannot_deactivate_admin` — GUEST → ADMIN deactivation blocked (403, victim stays active)
- `test_member_cannot_deactivate_admin` — MEMBER → ADMIN blocked (403)
- `test_guest_cannot_deactivate_peer_guest` — GUEST → GUEST blocked (403)
- `test_project_admin_can_deactivate_member` — positive control: project ADMIN may deactivate a MEMBER (200)
- `test_workspace_admin_with_low_project_role_can_deactivate` — positive control: workspace ADMIN (even with a low project role) may deactivate a project ADMIN (200), locking in the intended `is_workspace_admin` bypass

Verified fail-before: the three security tests return `200` (deactivation succeeds) on pre-fix code, `403` after the fix.

- `ruff check` — clean
- `python manage.py check` — no issues
- existing `tests/contract/app/test_project_app.py` — 25/25 pass (no regression)

### DIFF
--- a/apps/api/plane/app/views/project/member.py
+++ b/apps/api/plane/app/views/project/member.py
@@ -261,6 +261,25 @@ class ProjectMemberViewSet(BaseViewSet):
                     status=status.HTTP_400_BAD_REQUEST,
                 )
 
+        # Guard privileged `is_active` mutations (member (de)activation). These are NOT
+        # covered by the role block above, so without this check a GUEST could PATCH
+        # {"is_active": false} while omitting "role" to deactivate any member — including
+        # admins — and take over the project. Mirror the role block and destroy(): only a
+        # project admin (or workspace admin) may (de)activate a member, and never one whose
+        # role is equal to or higher than the requester's own.
+        if "is_active" in request.data:
+            if requested_project_member.role < ROLE.ADMIN.value and not is_workspace_admin:
+                return Response(
+                    {"error": "You do not have permission to update member status"},
+                    status=status.HTTP_403_FORBIDDEN,
+                )
+
+            if project_member.role >= requested_project_member.role and not is_workspace_admin:
+                return Response(
+                    {"error": "You cannot update the status of a member with a role equal to or higher than your own"},
+                    status=status.HTTP_403_FORBIDDEN,
+                )
+
         serializer = ProjectMemberSerializer(project_member, data=request.data, partial=True)
 
         if serializer.is_valid():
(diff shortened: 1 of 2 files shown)


mguptahub:
Thanks @Copilot — addressed both:

1. **Reactivation wording (`member.py:272`)** — correct catch. `partial_update` resolves the target with `is_active=True`, so an already-inactive member can't be addressed here and genuine reactivation never reaches the guard; reactivation is handled by the admin-gated member-add/`create` path. This is a description inaccuracy, not a security gap — the guard still governs deactivation of active members (and no-op `is_active: true` on an active one). Updated the PR description accordingly; no code change needed.

2. **Workspace-admin bypass test (`test:145`)** — good call. Added `test_workspace_admin_with_low_project_role_can_deactivate` (commit 65239689a6): a workspace ADMIN holding only a project GUEST role can still deactivate a project ADMIN, locking in the intended `is_workspace_admin` bypass against future regressions. Suite is now 5/5.

Saurabhkmr98 (approved):

dheeru0198 (approved):

PR #9372, opened 2026-07-08
[WEB-8066] fix: scope workspace asset get/patch/delete to project membership

mguptahub:
## Summary

`WorkspaceFileAssetEndpoint` (`apps/api/plane/app/views/asset/v2.py`) is authorized at the **WORKSPACE** level. Its `get`, `patch`, and `delete` fetch `FileAsset.objects.get(id=asset_id, workspace__slug=slug)` with no project-membership check, so **any workspace member or guest could download, mark-uploaded, or delete a project-bound asset** (issue attachment/description, comment description, page description) belonging to a project they are not a member of.

This is an incomplete fix of the GHSA-qw87 asset-IDOR cluster.

Advisories: **GHSA-h7mc-p9mm-2r4w** (dup: **GHSA-cjph-cgm5-8pw8**) · MEDIUM.

## Fix

- New `project_membership_denied(request, asset)` helper: for project-bound assets (`asset.project_id is not None`), require an active `ProjectMember` of `asset.project_id`, otherwise `403`.
- Workspace-level entity types (`WORKSPACE_LOGO`, `USER_AVATAR`, `USER_COVER`) have `project_id = NULL` and remain accessible to any workspace member.
- Guard is invoked in `get`, `patch`, and `delete` immediately after the asset fetch — before the `is_uploaded` check / any mutation — so a non-member receives a uniform `403` and cannot probe upload state.
- Mirrors the strictest adjacent endpoint, `ProjectAssetEndpoint` (`level="PROJECT"`).

## Tests

`plane/tests/contract/app/test_workspace_file_asset_project_scope_app.py`:
- `get`/`patch`/`delete` on a project asset by a non-project-member → `403` (and no presigned URL minted / no mutation).
- Positive control: active project member can still download.
- Exemption control: workspace-level asset stays accessible to any workspace member.

**Fail-before verified**: with the fix stashed, the three denied cases return `302`/`204`/`204` (the vulnerable behavior); positive + exemption controls still pass.

Gates: ruff ✅ · `manage.py check` ✅ · sibling `test_generic_asset.py` regression ✅.

## Follow-up (out of scope)

Sibling workspace-scoped endpoints share the same shape and warrant a separate audit: `AssetRestoreEndpoint`, `AssetCheckEndpoint`, `DuplicateAssetEndpoint`, `WorkspaceAssetDownloadEndpoint`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened access controls for project-bound workspace file assets: only active project members can view, update, or delete them.
  * Unauthorized requests now return `403` and are blocked before any download link is generated or any asset changes occur.
  * Workspace-level assets (no project scope) remain accessible to eligible workspace members.
* **Tests**
  * Added contract tests covering project scoping denial/allow scenarios, including presigned URL generation and ensuring no changes on forbidden PATCH/DELETE.

### DIFF
--- a/apps/api/plane/app/views/asset/v2.py
+++ b/apps/api/plane/app/views/asset/v2.py
@@ -324,8 +324,14 @@ class WorkspaceFileAssetEndpoint(BaseAPIView):
         """
         if asset.project_id is None:
             return None
+        # Scope the membership lookup to the asset's workspace as well as its
+        # project, mirroring allow_permission's PROJECT branch. This prevents a
+        # member of the same project in a different workspace from passing the
+        # check should an asset row ever be inconsistent (asset.workspace_id !=
+        # asset.project.workspace_id).
         is_project_member = ProjectMember.objects.filter(
             member=request.user,
+            workspace_id=asset.workspace_id,
             project_id=asset.project_id,
             is_active=True,
         ).exists()
--- a/apps/api/plane/app/views/asset/v2.py
+++ b/apps/api/plane/app/views/asset/v2.py
@@ -312,35 +312,29 @@ class WorkspaceFileAssetEndpoint(BaseAPIView):
         else:
             return
 
-    def project_membership_denied(self, request, asset):
-        """Enforce project-level access on a workspace-scoped asset lookup.
+    def has_project_asset_access(self, request, asset):
+        """Return whether the user may access a workspace-scoped asset.
 
         This endpoint is authorized at the WORKSPACE level, so a workspace
         member/guest could otherwise reach an asset that belongs to a project
         they are not a member of. For project-bound assets, require an active
         ProjectMember of the asset's project. Workspace-level entity types
         (WORKSPACE_LOGO, USER_AVATAR, USER_COVER) have project_id=None and are
-        exempt. Returns a 403 Response when access is denied, else None.
+        always allowed.
         """
         if asset.project_id is None:
-            return None
+            return True
         # Scope the membership lookup to the asset's workspace as well as its
         # project, mirroring allow_permission's PROJECT branch. This prevents a
         # member of the same project in a different workspace from passing the
         # check should an asset row ever be inconsistent (asset.workspace_id !=
         # asset.project.workspace_id).
-        is_project_member = ProjectMember.objects.filter(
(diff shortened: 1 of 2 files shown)


apps/api/plane/app/views/asset/v2.py
```diff
@@ -312,6 +312,30 @@ def entity_asset_delete(self, entity_type, asset, request):
...
+        (WORKSPACE_LOGO, USER_AVATAR, USER_COVER) have project_id=None and are
+        exempt. Returns a 403 Response when access is denied, else None.
+        """
+        if asset.project_id is None:
+            return None
+        is_project_member = ProjectMember.objects.filter(
+            member=request.user,
+            project_id=asset.project_id,
+            is_active=True,
+        ).exists()
```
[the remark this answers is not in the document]
mguptahub: Good catch — addressed in 90b6416. The `ProjectMember` lookup now also filters `workspace_id=asset.workspace_id`, mirroring `allow_permission`'s PROJECT branch, so a membership in the same project under a different workspace can't satisfy the check for an inconsistent asset row.

apps/api/plane/app/views/asset/v2.py
```diff
@@ -312,6 +312,36 @@ def entity_asset_delete(self, entity_type, asset, request):
         else:
             return
 
+    def project_membership_denied(self, request, asset):
```
Saurabhkmr98: we should make this helper function return a boolean value and return Response in the view itself based on the value

Saurabhkmr98 (changes requested):

apps/api/plane/app/views/asset/v2.py
mguptahub: Done in the latest commit — renamed to `has_project_asset_access(request, asset)` which now returns a boolean (`True` = allowed), and each of `get`/`patch`/`delete` builds the 403 `Response` itself based on the returned value. Behaviour is unchanged (same 403 + message; workspace-level assets with `project_id=None` still allowed). Existing 5 contract tests still pass.

Saurabhkmr98 (approved):

dheeru0198 (approved):

PR #9382, opened 2026-07-09
[WEB-8103] fix: stop leaking webhook HMAC secret_key on reads

mguptahub:
Fixes WEB-8103 (MED, admin-only).

## Problem
`WebhookEndpoint` list/retrieve/patch pass a `fields=(...)` allowlist that deliberately excludes `secret_key`, but `DynamicBaseSerializer.__init__` discards the caller allowlist (`fields = self.expand`). Combined with `WebhookSerializer` using `fields="__all__"` and `secret_key` only in `read_only_fields` (read-only fields are still serialized), the HMAC signing `secret_key` was returned on every webhook read. All handlers are `@allow_permission([ADMIN], level="WORKSPACE")`, so exposure is workspace-admin-only.

## Fix
The advisory suggests `secret_key = serializers.CharField(write_only=True)`, but that would (a) let a client **inject** their own secret on create/patch (a downgrade — the secret is server-generated via `default=generate_token`) and (b) break the intended one-time reveal. Instead:

- `WebhookSerializer.to_representation` drops `secret_key` unless a `show_secret_key` context flag is set — **secure by default**. `secret_key` stays server-generated and non-writable.
- `POST create` and `WebhookSecretRegenerateEndpoint` pass `show_secret_key=True`, so the secret is still returned **once** for the caller to configure their receiver.
- list / retrieve / patch no longer emit it.

`WebhookSerializer` is only used in `app/views/webhook/base.py` (no external API / EE consumer), so the blast radius is contained.

## Tests
`tests/contract/app/test_webhook_secret_key_scope_app.py` — 6 tests: list/retrieve/patch do not leak, create + regenerate still reveal, and a serializer-level check of both branches of the flag. Fail-before verified (the four no-leak tests return `secret_key` before the fix). `ruff` + `manage.py check` green.

## Follow-up
The root `DynamicBaseSerializer.__init__` bug (silently discarding the `fields=` allowlist) also affects other serializers (workspace/project member over-exposure) — tracked as a separate audit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Webhook responses now keep `secret_key` hidden by default, including when explicitly requested.
  * Webhook creation responses include the secret for initial setup.
  * Secret regeneration returns the newly generated secret.
  * Listing, retrieving, and updating webhooks no longer expose the secret.

* **Tests**
  * Added coverage verifying secret visibility across webhook creation, listing, retrieval, updates, and regeneration.

### DIFF
--- a/apps/api/plane/app/serializers/webhook.py
+++ b/apps/api/plane/app/serializers/webhook.py
@@ -61,7 +61,7 @@ class WebhookSerializer(DynamicBaseSerializer):
         # never on list/retrieve/update.
         #
         # This context flag — not the fields= allowlists in views/webhook/base.py —
-        # is the sole enforcement point (GHSA-83rj). Those fields= kwargs are dead
+        # is the sole enforcement point. Those fields= kwargs are dead
         # on two independent levels, so do not assume fixing either one re-activates
         # them:
         #   1. DynamicBaseSerializer.__init__ pops the caller's fields= and then
--- a/apps/api/plane/app/serializers/webhook.py
+++ b/apps/api/plane/app/serializers/webhook.py
@@ -58,9 +58,22 @@ class WebhookSerializer(DynamicBaseSerializer):
         data = super().to_representation(instance)
         # secret_key is the HMAC signing secret. It is only revealed on creation
         # and secret regeneration (opt-in via the show_secret_key context flag),
-        # never on list/retrieve/update. The per-request fields= allowlist that
-        # the webhook views pass to exclude it is silently dropped by
-        # DynamicBaseSerializer.__init__, so filter it here (GHSA-83rj).
+        # never on list/retrieve/update.
+        #
+        # This context flag — not the fields= allowlists in views/webhook/base.py —
+        # is the sole enforcement point (GHSA-83rj). Those fields= kwargs are dead
+        # on two independent levels, so do not assume fixing either one re-activates
+        # them:
+        #   1. DynamicBaseSerializer.__init__ pops the caller's fields= and then
+        #      overwrites it with self.expand (serializers/base.py:16-18), so the
+        #      requested allowlist never reaches _filter_fields at all.
+        #   2. _filter_fields never *removes* anything even when it does receive a
+        #      list — it builds `allowed` purely to add expansion serializers for
+        #      names not already present, then returns self.fields unfiltered
+        #      (serializers/base.py:45-119).
+        # A future one-line `fields = fields or self.expand` fix addresses only (1);
+        # (2) still has to be made restrictive before any fields= allowlist can be
+        # relied on for confidentiality.
         if not self.context.get("show_secret_key"):
             data.pop("secret_key", None)
         return data
--- a/apps/api/plane/app/serializers/webhook.py
+++ b/apps/api/plane/app/serializers/webhook.py
@@ -54,6 +54,17 @@ class WebhookSerializer(DynamicBaseSerializer):
         if any(hostname == domain or hostname.endswith("." + domain) for domain in disallowed_domains):
             raise serializers.ValidationError({"url": "URL domain or its subdomain is not allowed."})
 
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        # secret_key is the HMAC signing secret. It is only revealed on creation
+        # and secret regeneration (opt-in via the show_secret_key context flag),
+        # never on list/retrieve/update. The per-request fields= allowlist that
+        # the webhook views pass to exclude it is silently dropped by
+        # DynamicBaseSerializer.__init__, so filter it here (GHSA-83rj).
+        if not self.context.get("show_secret_key"):
+            data.pop("secret_key", None)
+        return data
+
     def create(self, validated_data):
         url = validated_data.get("url", None)
         self._validate_webhook_url(url)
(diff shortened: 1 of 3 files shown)


apps/api/plane/app/serializers/webhook.py
```diff
@@ -54,6 +54,17 @@ def _validate_webhook_url(self, url):
         if any(hostname == domain or hostname.endswith("." + domain) for domain in disallowed_domains):
             raise serializers.ValidationError({"url": "URL domain or its subdomain is not allowed."})
 
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        # secret_key is the HMAC signing secret. It is only revealed on creation
+        # and secret regeneration (opt-in via the show_secret_key context flag),
+        # never on list/retrieve/update. The per-request fields= allowlist that
+        # the webhook views pass to exclude it is silently dropped by
+        # DynamicBaseSerializer.__init__, so filter it here (GHSA-83rj).
```
sriramveeraghanta: One detail worth folding into the follow-up (or this comment): the discarded `fields=` is only half of the root cause. Even if `__init__` honored it, `DynamicBaseSerializer._filter_fields` never *removes* fields — it builds `allowed` but only uses it to add expanded fields, then `return self.fields` unfiltered. So the `fields=(...)` kwargs still passed in `views/webhook/base.py` (list/retrieve/patch) are no-ops on two levels, and the follow-up needs to make `_filter_fields` actually restrictive (or the views should drop the dead kwarg). Worth spelling out so a future one-line `fields = self.expand` fix isn't assumed to re-activate the allowlists.

apps/api/plane/tests/contract/app/test_webhook_secret_key_scope_app.py
```diff
@@ -0,0 +1,99 @@
...
+
+        assert response.status_code == status.HTTP_200_OK
+        body = response.json()
+        assert "secret_key" in body
+        assert body["secret_key"] and body["secret_key"] != old_secret
+
+    @pytest.mark.django_db
+    def test_create_reveals_secret_key(self, session_client, workspace, monkeypatch):
+        # Bypass the network-dependent SSRF URL validation for this unit.
+        monkeypatch.setattr(WebhookSerializer, "_validate_webhook_url", lambda self, url: None)
```
sriramveeraghanta: Minor: consider patching the network boundary instead of the whole private method, e.g.

```python
monkeypatch.setattr("plane.app.serializers.webhook.validate_url", lambda *a, **k: None)
```

With default test settings (`WEBHOOK_ALLOWED_HOSTS` / `WEBHOOK_DISALLOWED_DOMAINS` empty) `example.com` still passes the domain checks and the request host (`testserver`) doesn't collide, so `_validate_webhook_url`'s disallowed-domain logic stays exercised and the test survives a rename of the private method.

apps/api/plane/tests/contract/app/test_webhook_secret_key_scope_app.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+
+        assert response.status_code == status.HTTP_201_CREATED
+        assert response.json().get("secret_key")
+
+    @pytest.mark.django_db
+    def test_serializer_hides_secret_by_default_and_reveals_with_flag(self, workspace):
+        webhook = _make_webhook(workspace)
+
+        assert "secret_key" not in WebhookSerializer(webhook).data
```
sriramveeraghanta, line 99: Optional, cheap extra assertion: `WebhookSerializer(webhook, fields=("secret_key",)).data` should also hide the key. It pins `to_representation` as the enforcement point, so when the `DynamicBaseSerializer` `fields=` handling is eventually fixed, an explicit `fields=` request can't quietly re-open the leak without the context flag.

sriramveeraghanta (commented):
Solid fix — verified the claims against the codebase before commenting:

- `DynamicBaseSerializer.__init__` does discard the caller's `fields=` (`apps/api/plane/app/serializers/base.py:16-18`), and `WebhookSerializer` has no consumer outside `app/views/webhook/base.py` (the one test usage never serializes), so the blast-radius statement holds.
- The contract matches the web frontend: `apps/web/core/store/workspace/webhook.store.ts` only reads `secret_key` off the create (`createWebhook`) and regenerate (`regenerateSecretKey`) responses — list/retrieve/patch never used it. No client breakage expected.
- Test fixtures/routes/markers all line up (`session_client` + `workspace` give a workspace ADMIN; `contract` marker is registered in `apps/api/pytest.ini`).

Three small, non-blocking suggestions inline.

mguptahub:
Thanks @sriramveeraghanta — addressed in `1936de0`:

- **Patch the network boundary, not the private method** — `test_create_reveals_secret_key` now does `monkeypatch.setattr("plane.app.serializers.webhook.validate_url", lambda *a, **k: None)`. The domain/schema checks in `_validate_webhook_url` still run for `example.com`, and the test no longer breaks if the private method is renamed.
- **Pin `to_representation` as the enforcement point** — added `assert "secret_key" not in WebhookSerializer(webhook, fields=("secret_key",)).data`, so an explicit `fields=` request can't silently re-open the leak if the base-class allowlist is fixed later.

On your first point: agreed, and to spell it out — `DynamicBaseSerializer._filter_fields` builds `allowed` but never actually restricts (`return self.fields` unfiltered), so both the discarded `fields=` in `__init__` *and* the `fields=(...)` kwargs in `views/webhook/base.py` are no-ops on two levels. That base-class redesign (and auditing the workspace/project member serializers that rely on the same allowlist) is the separate follow-up; this PR deliberately enforces at `to_representation` so it holds regardless of that fix. 6/6 contract tests green.

sriramveeraghanta (approved):

PR #9419, opened 2026-07-13
feat: OpenTelemetry (OTel) observability for the community API

sriramveeraghanta:
### Description

Adds end-to-end **OpenTelemetry (OTel) observability** to the community API. Self-hosters can get OTLP-compatible APM for the Django API — a server span per request with DB / Redis / HTTP-client child spans, a span per Celery task (traceparent propagated through the queue), HTTP server metrics, and trace-correlated JSON logs — by setting two environment variables and pointing at any OTLP-compatible backend (Jaeger, Tempo, Datadog Agent, Honeycomb, Grafana Cloud, …).

Everything is **off by default** (`OTEL_ENABLED=0`) and gated behind `OTEL_ENABLED` + `OTEL_EXPORTER_OTLP_ENDPOINT`, so existing deployments are unaffected until they opt in. The off path is byte-for-byte unchanged (no OTLP connections, log schema untouched).

**Scope is the Django API server.** Node/browser/SSR instrumentation and browser-side telemetry are out of scope for this PR.

Highlights:

- New `apps/api/plane/observability/` package — `configure_otel()` (idempotent, gated) + `flush_otel()` + `TraceContextFilter`. Sets up a `TracerProvider` (parent-based 10% head sampler) and `MeterProvider`, selects grpc/http exporters by `OTEL_EXPORTER_OTLP_PROTOCOL`, instruments Django / Celery / Psycopg / Redis / Requests / HTTPX, and pins noisy `opentelemetry.*` loggers to WARNING.
- `configure_otel()` is called at the top of `wsgi.py`, `asgi.py`, `manage.py`, and `celery.py` **before Django wires up**, so `DjangoInstrumentor` patches the WSGI/ASGI handler with **zero view/serializer changes**.
- Trace-correlated JSON logs are wired via Django's `LOGGING` dict (both settings files). Because the Celery workers attach their own JSON handlers (they don't route through Django's `dictConfig`), a small gated filter in `celery.py` gives worker logs the same `trace_id` / `span_id` / `service_name` fields — the off-path handler setup is unchanged.
- Celery `worker_process_shutdown` flushes buffered spans/metrics (prefork children exit via `os._exit` and skip `atexit`).
- Adds the 7 new `opentelemetry-instrumentation-*` / OTLP-HTTP pins to `requirements/base.txt` (the 5 base OTel packages were already present).
- Wires the OTel env into `api` / `worker` / `beat-worker` in the `cli/community` compose (not `migrator`), and documents the vars in the `cli/community` and `aio/community` `variables.env` files.
- Adds a self-hoster guide and a reference collector config under `docs/otel-api-observability/`.

### Type of Change

- [ ] Code refactoring
- [x] Documentation update

### Test Scenarios

**Off by default (no regression):**
- Boot the API/worker with `OTEL_ENABLED` unset → normal startup, no OTLP connection attempts, and JSON logs keep their existing schema (no `trace_id`/`span_id`). Verified: `manage.py check` is clean, and the full test suite collects **494 tests with no import errors**.

**Unit tests (`plane/tests/unit/observability/`, 32 tests, all green):**
- Gating (off by default, explicit `0`, enabled-without-endpoint warns and returns, truthy tokens `1/true/yes/on`), provider + instrumentation wiring, `PsycopgInstrumentor(enable_commenter=False)`, default application + operator overrides, quiet loggers, idempotency, and protocol-aware exporter selection (grpc default, http/protobuf → http). Plus `TraceContextFilter` behavior.

**Enabled (`OTEL_ENABLED=1` + collector reachable):**
- Hit any REST endpoint → a server span with `http.route`/`http.status_code` plus child Postgres/Redis/HTTP-client spans; `http.server.duration` metrics arrive.
- Trigger an action that enqueues a Celery task → the request span and the task execution span share one trace.
- Confirm stdout JSON logs (API request path and Celery workers) carry `trace_id`/`span_id`/`service_name`.
- Set `OTEL_ENABLED=1` with `OTEL_EXPORTER_OTLP_ENDPOINT` **unset** → exactly one WARNING at boot; the app continues without instrumentation.

**Deployment:**
- `docker compose -f deployments/cli/community/docker-compose.yml config` renders the `OTEL_*` vars on `api`/`worker`/`beat-worker` only (not `migrator`).

### References

- Self-hoster guide & sample collector config: `docs/otel-api-observability/README.md`, `docs/otel-api-observability/otel-collector.yaml`

---
🤖 PR description generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added optional OpenTelemetry observability for API requests, background tasks, database queries, cache operations, outbound HTTP calls, and metrics.
  * Added trace and service details to logs for easier troubleshooting and request correlation.
  * OpenTelemetry is disabled by default and can be enabled through documented configuration settings.

* **Documentation**
  * Added setup guidance, supported configuration options, troubleshooting tips, and a sample OpenTelemetry Collector configuration.

* **Tests**
  * Added coverage for telemetry setup, exporters, logging correlation, and configuration behavior.


apps/api/plane/observability/setup.py
```diff
@@ -0,0 +1,217 @@
...
+        return
+
+    _apply_defaults()
+    resource = _build_resource()
+    _setup_tracing(resource)
+    _setup_metrics(resource)
+    _instrument_libraries()
+    _quiet_otel_loggers()
+
+    _CONFIGURED = True
```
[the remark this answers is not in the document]
sriramveeraghanta: `_CONFIGURED` is used — it's the module-level idempotency guard. It's read at the top of `configure_otel()` and set to `True` after a successful bootstrap (declared via `global _CONFIGURED`), so the bootstrap can be called safely from all four entry points (`wsgi.py`, `asgi.py`, `manage.py`, `celery.py`) but runs exactly once. This is a known false positive for module globals mutated through `global`. Behaviour is covered by `test_idempotent_when_called_twice` and `test_idempotent_when_disabled`. No change needed.

apps/api/plane/settings/production.py
```diff
@@ -97,3 +97,19 @@
         },
     },
 }
+
+# OpenTelemetry APM: only when OTEL_ENABLED=1 do we extend the JSON
+# formatter and attach TraceContextFilter to every handler. Attaching at
+# the handler level (rather than the root logger) is required because
+# most plane.* loggers have propagate=False; runtime mutation also wouldn't
+# survive Django's dictConfig. Off path leaves the log schema unchanged.
+if os.environ.get("OTEL_ENABLED", "0").lower() in ("1", "true", "yes"):
```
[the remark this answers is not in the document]
sriramveeraghanta: Fixed in 9a7d840. Extracted a shared `is_otel_enabled()` into `plane.observability.logging` (single source of truth for the truthy tokens, including `on`, with `.strip()`), and reused it in the bootstrap, Celery, and both Django `LOGGING` gates. The settings block is now `extend_logging_config(LOGGING)`, so `OTEL_ENABLED=on` (and whitespace-padded values) now trace-correlate the JSON logs consistently with `configure_otel()`, matching the README.

apps/api/plane/settings/local.py
```diff
@@ -87,3 +87,19 @@
         },
     },
 }
+
+# OpenTelemetry APM: only when OTEL_ENABLED=1 do we extend the JSON
+# formatter and attach TraceContextFilter to every handler. Attaching at
+# the handler level (rather than the root logger) is required because
+# most plane.* loggers have propagate=False; runtime mutation also wouldn't
+# survive Django's dictConfig. Off path leaves the log schema unchanged.
+if os.environ.get("OTEL_ENABLED", "0").lower() in ("1", "true", "yes"):
```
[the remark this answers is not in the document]
sriramveeraghanta: Fixed in 9a7d840 via the shared `is_otel_enabled()` gate (see `plane/observability/logging.py`), now used here through `extend_logging_config(LOGGING)`. `on` and whitespace-padded values enable log correlation identically to `configure_otel()`.

apps/api/plane/celery.py
```diff
@@ -19,6 +23,54 @@
...
+# patch task execution. No-op unless OTEL_ENABLED=1.
+from plane.observability.setup import configure_otel, flush_otel  # noqa: E402
+from plane.observability.logging import TraceContextFilter  # noqa: E402
+
+configure_otel()
+
+# Whether to trace-correlate worker logs. Matches the Django LOGGING gate in
+# plane/settings/{local,production}.py; the bootstrap in configure_otel() uses
+# its own (superset) token check.
+_OTEL_LOG_ENABLED = os.environ.get("OTEL_ENABLED", "0").strip().lower() in ("1", "true", "yes")
```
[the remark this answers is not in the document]
sriramveeraghanta: Fixed in 9a7d840 — `_OTEL_LOG_ENABLED = is_otel_enabled()` now uses the shared gate, so `OTEL_ENABLED=on` also trace-correlates worker logs. The stale "superset token check" comment was removed.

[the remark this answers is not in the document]
sriramveeraghanta: Good catch — fixed in 9a7d840. Rather than just adding `on` in one place, I extracted a shared `is_otel_enabled()` into `plane.observability.logging` (owning the `("1","true","yes","on")` token set) and reused it in `setup.py` (`configure_otel`), `celery.py` (`_OTEL_LOG_ENABLED`), and both Django settings via `extend_logging_config(LOGGING)`. That removes the drift entirely — all gates now accept the same documented tokens.

apps/api/plane/observability/setup.py
```diff
@@ -0,0 +1,217 @@
...
+_CONFIGURED = False
+
+# Kept at module scope so flush_otel() can force-flush them on worker shutdown
+# (prefork children exit via os._exit and skip atexit).
+_TRACER_PROVIDER: "TracerProvider | None" = None
+_METER_PROVIDER: "MeterProvider | None" = None
+
+# Accepted "on" tokens — kept in sync with the pi/node services so OTEL_ENABLED
+# behaves identically across every runtime.
+_TRUTHY_VALUES = ("1", "true", "yes", "on")
```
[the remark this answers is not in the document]
sriramveeraghanta: Done in 9a7d840. Exported `is_otel_enabled()` from `plane.observability.logging` as the single source of truth for the truthy tokens (incl. `on`) and reused it in `setup.py`, `celery.py`, and both settings modules, so gating is identical everywhere and can't drift.

apps/api/plane/settings/local.py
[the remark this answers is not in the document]
sriramveeraghanta: Fixed in 9a7d840. Both settings files now call the shared `extend_logging_config(LOGGING)` helper, whose gate (`is_otel_enabled()`) accepts `on`, so a user setting `OTEL_ENABLED=on` gets trace-correlated logs. The README's documented token list is now honoured — no doc change needed.

deployments/cli/community/docker-compose.yml
```diff
@@ -61,6 +61,16 @@ x-app-env: &app-env
...
+# OpenTelemetry APM for the Django API (api, worker, beat-worker).
+# Read at container RUNTIME; off by default. See docs/otel-api-observability/README.md.
+x-otel-env: &otel-env
+  OTEL_ENABLED: ${OTEL_ENABLED:-0}
+  OTEL_EXPORTER_OTLP_ENDPOINT: ${OTEL_EXPORTER_OTLP_ENDPOINT:-}
+  OTEL_EXPORTER_OTLP_PROTOCOL: ${OTEL_EXPORTER_OTLP_PROTOCOL:-grpc}
+  OTEL_EXPORTER_OTLP_HEADERS: ${OTEL_EXPORTER_OTLP_HEADERS:-}
+  OTEL_SERVICE_NAME: ${OTEL_SERVICE_NAME:-plane-api}
+  OTEL_ENVIRONMENT: ${OTEL_ENVIRONMENT:-}
+
```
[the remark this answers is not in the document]
sriramveeraghanta: Fixed in 9a7d840 — added `OTEL_TRACES_SAMPLER`, `OTEL_TRACES_SAMPLER_ARG`, and `OTEL_RESOURCE_ATTRIBUTES` to the `x-otel-env` anchor (empty defaults so the app's own defaults apply when unset). The vars documented in `.env.example` now reach the `api`/`worker`/`beat-worker` containers; verified via `docker compose config`.

deployments/cli/community/docker-compose.yml
```diff
@@ -61,6 +61,19 @@ x-app-env: &app-env
   WEBHOOK_ALLOWED_IPS: ${WEBHOOK_ALLOWED_IPS:-}
   WEBHOOK_ALLOWED_HOSTS: ${WEBHOOK_ALLOWED_HOSTS:-}
 
+# OpenTelemetry APM for the Django API (api, worker, beat-worker).
+# Read at container RUNTIME; off by default. See docs/otel-api-observability/README.md.
+x-otel-env: &otel-env
+  OTEL_ENABLED: ${OTEL_ENABLED:-0}
+  OTEL_EXPORTER_OTLP_ENDPOINT: ${OTEL_EXPORTER_OTLP_ENDPOINT:-}
+  OTEL_EXPORTER_OTLP_PROTOCOL: ${OTEL_EXPORTER_OTLP_PROTOCOL:-grpc}
+  OTEL_EXPORTER_OTLP_HEADERS: ${OTEL_EXPORTER_OTLP_HEADERS:-}
```
sriramveeraghanta, line 78: **[1] Credential leak: `OTEL_EXPORTER_OTLP_HEADERS` is picked up by the license-telemetry exporter and sent to telemetry.plane.so**

Plumbing this var into the api/worker containers changes the behavior of the pre-existing exporter in `apps/api/plane/license/bgtasks/telemetry_metrics.py`: `_create_otlp_metric_exporter` constructs `OTLPMetricExporter(endpoint=..., insecure=...)` with no `headers=`, and in `opentelemetry-exporter-otlp-proto-grpc==1.28.1` `OTLPExporterMixin.__init__` falls back to `self._headers = headers or environ.get(OTEL_EXPORTER_OTLP_HEADERS)` (the HTTP exporter does the same via `environ.get(OTEL_EXPORTER_OTLP_HEADERS, "")`).

So an operator who sets `OTEL_EXPORTER_OTLP_HEADERS=Authorization=Bearer <APM token>` (exactly what `apps/api/.env.example` suggests) will have that Authorization header attached as gRPC metadata / HTTP headers on the instance metrics that the scheduled `push_instance_metrics` beat task (every 6h by default) sends to the default `https://telemetry.plane.so` endpoint — exfiltrating their private APM credential to a third party.

Fix: pass explicit `headers={}` (or a pinned header set) in both branches of `_create_otlp_metric_exporter` so the license exporter never inherits operator OTLP env.

apps/api/manage.py
```diff
@@ -8,6 +8,13 @@
 
 if __name__ == "__main__":
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "plane.settings.production")
+
+    # Bootstrap OpenTelemetry before Django imports so runserver and
+    # management commands are instrumented too. No-op unless OTEL_ENABLED=1.
+    from plane.observability.setup import configure_otel
```
sriramveeraghanta, line 36: **[2] `manage.py --settings=...` is now silently ignored (breaks the local runserver entrypoint, even with `OTEL_ENABLED=0`)**

This import pulls in the `plane` package: `plane/__init__.py` does `from .celery import app`, and `plane/celery.py` executes `ri = redis_instance()` at module scope, which reads `settings.REDIS_SSL` and materializes Django's `LazySettings` from the `DJANGO_SETTINGS_MODULE` that was setdefault'd to `plane.settings.production` two lines above — all before `execute_from_command_line` gets a chance to honor `--settings`.

`apps/api/bin/docker-entrypoint-api-local.sh` ends with `python manage.py runserver 0.0.0.0:8000 --settings=plane.settings.local`; post-PR that dev server silently runs **production** settings. Pre-PR, manage.py imported nothing from `plane` before `execute_from_command_line`, so `--settings` worked.

Side effect worth knowing: because of this same import chain, the `configure_otel()` calls in `wsgi.py`, `asgi.py`, and here are all no-ops behind the `_CONFIGURED` flag — the load-bearing bootstrap is the hidden one that runs during `plane/__init__` -> `plane/celery.py` import.

Fix: don't import anything from the `plane` package in manage.py before command-line handling (e.g. parse/apply `--settings` first, or move the observability package out of the `plane` package's eager import path).

deployments/aio/community/variables.env
```diff
@@ -72,3 +72,12 @@ WEBHOOK_ALLOWED_IPS=
 # SSRF check. Useful for trusted internal services whose container/service IPs are
 # dynamic (e.g. "silo,silo.namespace.svc.cluster.local")
 WEBHOOK_ALLOWED_HOSTS=
+
+# OpenTelemetry APM (self-hoster observability). Off by default.
+# See docs/otel-api-observability/README.md.
+OTEL_ENABLED=0
```
sriramveeraghanta: **[3] AIO image can never enable OTel: baked-in `OTEL_ENABLED=0` overrides container env**

`build.sh:97` copies this file into the image as `/app/plane.env`, and `start.sh:190` runs `export $(grep -v '^#' plane.env | xargs)` immediately before launching supervisord — i.e. **after** container env is applied. Because `OTEL_ENABLED=0` is the only *active* (uncommented) OTEL line, `docker run -e OTEL_ENABLED=1 -e OTEL_EXPORTER_OTLP_ENDPOINT=...` gets `OTEL_ENABLED` re-exported back to `0` for every supervised process (api/worker/beat/migrator). The commented OTEL vars pass through fine; only the gate is clobbered. `update_env_file()` never syncs OTEL keys, and plane.env is baked in with no volume, so the README quickstart is unfollowable on AIO while its troubleshooting step ("confirm OTEL_ENABLED=1 is in the container's env") appears satisfied.

Fix: ship `OTEL_ENABLED` commented out here (matching the other OTEL_* lines), or make start.sh prefer already-set container env over plane.env values.

apps/api/plane/celery.py
```diff
@@ -19,6 +23,54 @@
 # Set the default Django settings module for the 'celery' program.
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "plane.settings.production")
 
+# Bootstrap OpenTelemetry before Celery wires up so CeleryInstrumentor can
+# patch task execution. No-op unless OTEL_ENABLED=1.
+from plane.observability.setup import configure_otel, flush_otel  # noqa: E402
+from plane.observability.logging import TraceContextFilter, is_otel_enabled  # noqa: E402
+
+configure_otel()
```
sriramveeraghanta: **[4] Prefork fork-safety: OTLP gRPC channel created in the parent, inherited by forked task children**

`configure_otel()` runs at module import in the Celery MainProcess, before billiard forks the prefork pool (`celery -A plane worker` in `docker-entrypoint-worker.sh` uses the default prefork pool). In `opentelemetry-exporter-otlp-proto-grpc==1.28.1` the exporter creates its gRPC channel eagerly in `__init__` and registers no `os.register_at_fork` handler. The SDK's `BatchSpanProcessor` / `PeriodicExportingMetricReader` do restart their worker threads post-fork, so children happily queue spans/metrics into an exporter whose inherited gRPC channel is not fork-safe (grpc-python fork support is env-gated and heavily caveated) — child exports can hang or fail nondeterministically, silently dropping the task telemetry that is this PR's headline feature for the worker container, and `worker_process_shutdown` -> `flush_otel()` flushes into the same broken channel.

This is why the OTel Celery instrumentation docs prescribe initializing tracer/meter providers in the `worker_process_init` signal. Fix: move provider/exporter creation into `worker_process_init` for the worker process (keeping the import-time call for beat/solo, or gating by pool type).

deployments/cli/community/docker-compose.yml
```diff
@@ -61,6 +61,19 @@ x-app-env: &app-env
...
+  OTEL_SERVICE_NAME: ${OTEL_SERVICE_NAME:-plane-api}
+  OTEL_ENVIRONMENT: ${OTEL_ENVIRONMENT:-}
+  OTEL_TRACES_SAMPLER: ${OTEL_TRACES_SAMPLER:-}
+  OTEL_TRACES_SAMPLER_ARG: ${OTEL_TRACES_SAMPLER_ARG:-}
```
sriramveeraghanta: **[5] `${VAR:-}` injects empty strings that defeat the Python defaults — every compose user who enables OTel gets 100% sampling instead of the documented 10%**

With the host var unset, `${OTEL_TRACES_SAMPLER:-}` sets the container env var to the **empty string** (verified with `docker compose config`). `os.environ.setdefault("OTEL_TRACES_SAMPLER", "parentbased_traceidratio")` in `_apply_defaults()` is then a no-op (the key exists), and opentelemetry-sdk 1.28.1 `sampling._get_from_env_or_default()` treats `""` as unrecognized: it warns (a warning `_quiet_otel_loggers()` suppresses) and falls back to `parentbased_always_on` — 100% sampling. Worse, `OTEL_TRACES_SAMPLER_ARG=""` makes `float("")` raise, so even an operator who explicitly picks `parentbased_traceidratio` gets rate 1.0 unless they also set the arg.

```suggestion
  OTEL_TRACES_SAMPLER: ${OTEL_TRACES_SAMPLER:-parentbased_traceidratio}
  OTEL_TRACES_SAMPLER_ARG: ${OTEL_TRACES_SAMPLER_ARG:-0.1}
```

(Alternatively treat empty-as-unset in `_apply_defaults()`, which would also cover `OTEL_RESOURCE_ATTRIBUTES: ${OTEL_RESOURCE_ATTRIBUTES:-}` and any other empty-string injections.)

apps/api/plane/observability/setup.py
```diff
@@ -0,0 +1,222 @@
...
+
+    _apply_defaults()
+    resource = _build_resource()
+    _setup_tracing(resource)
+    _setup_metrics(resource)
+    _instrument_libraries()
+    _quiet_otel_loggers()
+
+    _CONFIGURED = True
+    logger.info(
```
sriramveeraghanta: **[6] This INFO line can never be printed, and after dictConfig all OTel/observability logging is disabled — misconfiguration becomes fully silent**

Two-stage problem:
1. `configure_otel()` runs before any logging is configured in every entrypoint; the root logger has no handlers and `logging.lastResort` only emits WARNING+, so this INFO confirmation is dropped. The README troubleshooting step ("Check API logs for the `OpenTelemetry configured` INFO line at boot") can never succeed.
2. Both `settings/local.py` and `settings/production.py` apply LOGGING with `"disable_existing_loggers": True`. By then the `plane.observability.setup` logger and the `opentelemetry.*` loggers already exist (created at import/bootstrap time, pre-dictConfig), and none are named in LOGGING — so dictConfig sets `disabled=True` on all of them. Export errors from an unreachable collector, instrumentor failures, and `flush_otel()` warnings all vanish. The README's "connection refused floods" troubleshooting entry describes messages that cannot appear.

Fix: emit the boot confirmation via an explicit stderr handler (or defer it until after logging config), and exempt these loggers from `disable_existing_loggers` (name them in LOGGING or flip the flag).

apps/api/manage.py
```diff
@@ -8,6 +8,13 @@
...
+
+    configure_otel()
```
sriramveeraghanta, line 38: **[7] With the collector unreachable, atexit OTel flush stalls every boot command and every recycled gunicorn worker**

`TracerProvider` registers an atexit shutdown whose `BatchSpanProcessor.shutdown()` does an unbounded `worker_thread.join()`, and the 1.28.1 gRPC exporter retries `UNAVAILABLE` with exponential backoff summing to ~63s per batch. Consequences with `OTEL_ENABLED=1` and the collector down (the compose files ship no collector service and no `depends_on`, so at cold start it typically is):

- `docker-entrypoint-api.sh` runs ~7 `manage.py` commands (wait_for_db, wait_for_migrations, register_instance, configure_instance, create_bucket, clear_cache, collectstatic) before gunicorn; each generates instrumented DB/Redis spans and can stall up to ~1 min at interpreter exit — minutes of added container startup, with no log output explaining why (see the silenced-loggers finding).
- Steady state: gunicorn runs `--max-requests 1200 --max-requests-jitter 1000` without `--preload`, so each worker owns its own providers and exits through the same atexit flush on recycle; the arbiter kills it after ~30s. That's a recurring capacity dip every 1200-2200 requests per worker, plus guaranteed span loss on every recycle.

Fix: bound the shutdown (`OTEL_EXPORTER_OTLP_TIMEOUT`, or a custom bounded shutdown like `flush_otel()` already does for celery children), and/or skip bootstrap for short-lived management commands.

apps/api/plane/observability/setup.py
```diff
@@ -0,0 +1,222 @@
...
+    the environment natively. We additionally set deployment.environment.name
+    from a dedicated OTEL_ENVIRONMENT (or SENTRY_ENVIRONMENT) var. Note the
+    current semconv key is `deployment.environment.name` (not the legacy
+    `deployment.environment`).
+    """
+    attributes: dict[str, str] = {}
+    environment = os.environ.get("OTEL_ENVIRONMENT") or os.environ.get("SENTRY_ENVIRONMENT") or ""
+    if environment:
+        attributes["deployment.environment.name"] = environment
+    return Resource.create(attributes)
```
sriramveeraghanta, line 117: **[8] No `service.instance.id`: every gunicorn worker exports metric streams with identical identity, corrupting `http.server.*` metrics**

`Resource.create()` sets `service.name` + `telemetry.sdk.*` but never `service.instance.id`, and `_build_resource()` doesn't add one. The API serves via `gunicorn -w $GUNICORN_WORKERS -k uvicorn.workers.UvicornWorker` with no `--preload`, so N worker processes each run `configure_otel()` and each `PeriodicExportingMetricReader` exports its own cumulative `http.server.duration` / `http.server.active_requests` sums under byte-identical resource+scope+attribute identity. That violates OTLP's single-writer principle: backends that key series on stream identity (Prometheus remote-write, Mimir, most OTLP stores) receive N interleaved conflicting cumulative points for the same series — counters sawtooth between per-process values and histograms become unusable. Celery prefork children compound it (parent and children share the inherited resource).

Fix: stamp a per-process id in `_build_resource()`, e.g. `attributes["service.instance.id"] = str(uuid.uuid4())`.

apps/api/plane/observability/setup.py
```diff
@@ -0,0 +1,222 @@
...
+is truthy and OTEL_EXPORTER_OTLP_ENDPOINT is set.
+"""
+
+import logging
+import os
+
+from opentelemetry import metrics, trace
+from opentelemetry.instrumentation.celery import CeleryInstrumentor
+from opentelemetry.instrumentation.django import DjangoInstrumentor
+from opentelemetry.instrumentation.httpx import HTTPXClientInstrumentor
```
sriramveeraghanta: **[9] All six instrumentors are imported unconditionally at module scope — and `httpx` is only a transitive dependency**

These imports run in every process (wsgi/asgi/manage/celery import this module before the `OTEL_ENABLED` gate is evaluated) and sit outside the per-instrumentor try/except in `_instrument_libraries`, defeating that isolation for import-time failures. `opentelemetry-instrumentation-httpx==0.49b1` has a bare module-level `import httpx`, and its wheel declares httpx only under the `instruments` extra — while `requirements/base.txt` pins the plain package and contains no httpx line. Today httpx arrives transitively via `openai==1.63.2` only; if that edge disappears (openai removed/replaced, or a stale venv on upgrade — the five other instrumentation packages are also new requirements), api, worker, beat, and every manage.py command crash at import with a raw `ModuleNotFoundError`, **even with `OTEL_ENABLED=0`**. In manage.py the import also runs before the try/except that converts ImportError into the friendly "couldn't import Django" message.

Fix: defer the instrumentor (and SDK) imports into `configure_otel()` after the gate — `_create_span_exporter` already demonstrates the pattern — and pin httpx explicitly (or depend on `opentelemetry-instrumentation-httpx[instruments]`). This also removes the cold-start import cost every OTEL-disabled process currently pays.

apps/api/plane/observability/logging.py
```diff
@@ -0,0 +1,76 @@
...
+    When enabled, extends the JSON formatter's fmt with the trace-context fields
+    and attaches TraceContextFilter to every handler. Attaching at the handler
+    level (rather than the root logger) is required because most plane.* loggers
+    set propagate=False; runtime mutation also wouldn't survive Django's
+    dictConfig. The off path leaves the log schema byte-for-byte unchanged.
+
+    Shared by settings/local.py and settings/production.py so the two stay in
+    lockstep.
+    """
+    if not is_otel_enabled():
```
sriramveeraghanta: **[10] Log-schema gate diverges from the bootstrap gate: `OTEL_ENABLED=1` without an endpoint changes the log schema forever, with permanently empty trace ids**

`extend_logging_config()` (and `_OTEL_LOG_ENABLED` in `celery.py`) gate only on `is_otel_enabled()`, but `configure_otel()` additionally requires `OTEL_EXPORTER_OTLP_ENDPOINT`. In the exact misconfiguration setup.py warns about — and the compose default produces when a user only flips `OTEL_ENABLED=1` (`OTEL_EXPORTER_OTLP_ENDPOINT: ${OTEL_EXPORTER_OTLP_ENDPOINT:-}`) — no TracerProvider is installed, yet every Django and Celery JSON log line permanently gains empty `trace_id`/`span_id`/`trace_flags` plus `service_name`, paying per-record filter cost for zero telemetry. That's a third schema state the README doesn't describe (it claims the schema only changes when OTel is on), and its troubleshooting tip ("trace_id is empty -> sampler") sends operators down the wrong path.

Related: `_has_endpoint()` also refuses spec-valid configs that set only the signal-specific `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` / `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` vars, which the pinned 1.28.1 exporters honor.

Fix: one shared `is_otel_active()` predicate (enabled AND any endpoint var set) used by setup.py, both settings files, and celery.py.

deployments/cli/community/docker-compose.yml
```diff
@@ -61,6 +61,19 @@ x-app-env: &app-env
   WEBHOOK_ALLOWED_IPS: ${WEBHOOK_ALLOWED_IPS:-}
   WEBHOOK_ALLOWED_HOSTS: ${WEBHOOK_ALLOWED_HOSTS:-}
 
+# OpenTelemetry APM for the Django API (api, worker, beat-worker).
+# Read at container RUNTIME; off by default. See docs/otel-api-observability/README.md.
+x-otel-env: &otel-env
```
sriramveeraghanta, line 69: **[11] `migrator` service never receives the OTel env**

The anchor was added to `api`, `worker`, and `beat-worker`, but the `migrator` service's environment (line 181) still merges only the old anchors. manage.py was explicitly instrumented so "management commands are instrumented too", yet the one container dedicated to manage.py (`docker-entrypoint-migrator.sh`: wait_for_db + migrate — often the slowest, riskiest boot phase) silently produces no telemetry however the operator configures OTel.

Fix: add `*otel-env` to migrator's `<<:` list.

docs/otel-api-observability/otel-collector.yaml
```diff
@@ -0,0 +1,63 @@
...
+      grpc:
+        endpoint: 0.0.0.0:4317
+      http:
+        endpoint: 0.0.0.0:4318
+
+  filelog:
+    include:
+      - /var/lib/docker/containers/*/*-json.log
+    operators:
+      - type: json_parser
```
sriramveeraghanta, line 20: **[12] Reference filelog pipeline can never correlate logs with traces**

Two problems:
1. Docker's json-file driver wraps every stdout line as `{"log": "<app json>\n", "stream": ..., "time": ...}`. The single `json_parser` (default `parse_from: body`) therefore produces only `attributes.log` / `attributes.stream` / `attributes.time`, with the app's JSON left as an unparsed string in `attributes.log` — `attributes.trace_id` never exists, and both `move` operators error on every entry (passing entries through untransformed under the default `on_error: send`).
2. Even with the ids extracted, `to: trace_id` is interpreted as `body.trace_id` (stanza assumes `body` for non-prefixed fields) — it never sets the entry's top-level trace context. That requires the `trace_parser` operator.

Fix: add a second `json_parser` with `parse_from: attributes.log`, then replace both `move` ops with `trace_parser` (`trace_id.parse_from: attributes.trace_id`, `span_id.parse_from: attributes.span_id`).

docs/otel-api-observability/README.md
```diff
@@ -0,0 +1,47 @@
...
+   OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4317
+   ```
+3. Restart the API/worker containers. That's it.
+
+## What you get
+
+- A server span per HTTP request, with `http.route`, `http.method`, `http.status_code`, `http.target`, and duration.
+- A span per Celery task with `celery.action` / `celery.task_name` / `celery.state`. Traceparent is propagated through the queue, so a request that enqueues a task is linked to that task's execution span in the same trace.
+- Child spans for every Postgres query, Redis op, and outbound `requests` / `httpx` call inside that request or task.
+- HTTP metrics: `http.server.duration` histogram, `http.server.active_requests`, `http.server.request.size`, `http.server.response.size`.
```
sriramveeraghanta: **[13] Two of the four promised metrics are never emitted**

`opentelemetry-instrumentation-django==0.49b1` creates only the `http.server.duration` histograms and `http.server.active_requests` (verified in contrib v0.49b1 `django/__init__.py` — no size histograms exist anywhere in the Django instrumentation on either the WSGI or ASGI path; the `http.server.request.size`/`http.server.response.size` metrics belong to the standalone ASGI `OpenTelemetryMiddleware`, which `DjangoInstrumentor` does not use). Operators building dashboards or alerts from this list will chase collector config for data the API never sends.

Fix: drop `http.server.request.size` and `http.server.response.size` from this line.

apps/api/plane/observability/setup.py
```diff
@@ -0,0 +1,222 @@
...
+)
+
+_HTTP_PROTOCOLS = ("http/protobuf", "http")
+
+
+def _has_endpoint() -> bool:
+    return bool(os.environ.get("OTEL_EXPORTER_OTLP_ENDPOINT", "").strip())
+
+
+def _protocol() -> str:
```
sriramveeraghanta, line 66: **[14] `_protocol()` lowercases but doesn't strip, unlike `is_otel_enabled()`/`_has_endpoint()`**

A whitespace-padded value (env-file values are taken literally, e.g. `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf ` with a trailing space) fails the `_HTTP_PROTOCOLS` membership test and silently selects the gRPC exporter against an HTTP :4318 endpoint — every export fails with no surfaced error (see the silenced-loggers finding).

```suggestion
    return os.environ.get("OTEL_EXPORTER_OTLP_PROTOCOL", "grpc").strip().lower()
```

apps/api/plane/tests/unit/observability/conftest.py
```diff
@@ -0,0 +1,39 @@
...
+  bookkeeping).
+"""
+
+import os
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def isolate_otel_state(monkeypatch):
```
sriramveeraghanta, line 29: **[15] Env restore is defeated by monkeypatch teardown ordering — pre-existing host `OTEL_*` vars are permanently stripped**

`isolate_otel_state` depends on `monkeypatch`, so `monkeypatch.undo()` runs **after** this fixture's post-yield restore. A test that calls `monkeypatch.setenv("OTEL_ENABLED", "1")` records the var as previously-absent (this fixture deleted it during setup); at teardown the fixture restores the developer's original value, then monkeypatch's undo deletes it again. Any `OTEL_*` var exported in the developer's shell is silently removed for the remainder of the pytest session — exactly the leak the docstring says this fixture prevents.

Fix: don't mix mechanisms — delete the vars via `monkeypatch.delenv(key, raising=False)` in setup (so undo restores them correctly), and drop the manual save/restore.

sriramveeraghanta:
Fixing:
- @coderabbitai: comment — [![Review Change Stack](https://storage.googleapis.com/coderabbit_publi…
- @copilot-pull-request-reviewer: review summary — Pull request overview
- @coderabbitai: review summary — Actionable comments posted: 4**
- @coderabbitai: review summary — <details>

Will be in the next commit.

apps/api/plane/observability/setup.py
```diff
@@ -0,0 +1,326 @@
...
+def _init_providers() -> None:
+    """Create the exporters + providers for the *current* process. Idempotent."""
+    global _PROVIDERS_READY
+    if _PROVIDERS_READY:
+        return
+    resource = _build_resource()
+    _setup_tracing(resource)
+    _setup_metrics(resource)
+    atexit.register(flush_otel)
+    _PROVIDERS_READY = True
```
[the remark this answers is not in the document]
sriramveeraghanta: False positive — `_PROVIDERS_READY` is already initialized at module scope in `apps/api/plane/observability/setup.py:36`, right after `_CONFIGURED = False`, which is exactly the change suggested here.

It is a real per-process guard: `_init_providers()` declares `global _PROVIDERS_READY`, returns early when it is `True`, and sets it after building the tracer/meter providers. That makes provider creation idempotent whether it happens at import (wsgi/asgi/manage/beat) or in a forked prefork child via `worker_process_init` → `init_process_providers()`.

Same known limitation with module globals mutated through `global` as the earlier `_CONFIGURED` report on this PR. Covered by `test_init_process_providers_creates_deferred_exporters`, which asserts a second call builds no second exporter. No change needed.

apps/api/plane/celery.py
```diff
@@ -19,6 +25,93 @@
...
+    """
+    argv = sys.argv
+    if "worker" not in argv:
+        return False
+    for index, arg in enumerate(argv):
+        if arg.startswith("--pool="):
+            return arg.split("=", 1)[1] == "prefork"
+        if arg in ("-P", "--pool"):
+            return index + 1 < len(argv) and argv[index + 1] == "prefork"
+    return True  # prefork is Celery's default pool
```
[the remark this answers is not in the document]
sriramveeraghanta: Good catch — fixed in 5791392.

`_is_prefork_worker()` read only `sys.argv`, so a pool coming from `CELERY_WORKER_POOL` was invisible to it and defaulted to `prefork`. For `threads` that meant deferring provider creation to `worker_process_init`, which that pool never dispatches, so the worker exported nothing.

Added `_effective_pool()`, which applies Celery's own precedence — `-P`/`--pool` flag, then `worker_pool` from the config object (`CELERY_WORKER_POOL` in Django settings), then the `prefork` default — and `_is_prefork_worker()` is now `"worker" in sys.argv and _effective_pool() == "prefork"`. Every non-forking pool keeps the import-time bootstrap in the main process.

The settings lookup is short-circuited behind the `"worker" in sys.argv` check so wsgi/asgi/manage.py do not resolve `settings` any earlier than they already do.

Covered by `plane/tests/unit/observability/test_celery_pool.py` (12 cases: argv forms, the settings-configured `threads` case, and CLI-overrides-settings).

deployments/cli/community/docker-compose.yml
```diff
@@ -61,6 +61,22 @@ x-app-env: &app-env
...
   WEBHOOK_ALLOWED_HOSTS: ${WEBHOOK_ALLOWED_HOSTS:-}
 
+# OpenTelemetry APM for the Django API (api, worker, beat-worker, migrator).
+# Read at container RUNTIME; off by default. See docs/otel-api-observability/README.md.
+# Note the sampler defaults are spelled out here rather than left as `${VAR:-}`:
+# an unset host var would otherwise be injected as an empty string, which the
+# OTEL SDK rejects (silently falling back to 100% sampling).
+x-otel-env: &otel-env
+  OTEL_ENABLED: ${OTEL_ENABLED:-0}
+  OTEL_EXPORTER_OTLP_ENDPOINT: ${OTEL_EXPORTER_OTLP_ENDPOINT:-}
```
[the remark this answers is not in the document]
sriramveeraghanta: Agreed — fixed in 5791392. Added `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` to the `x-otel-env` anchor, so all four services that merge it (api, worker, beat-worker, migrator) receive them.

Without this, an operator setting only a signal-specific endpoint got `is_otel_active()` false and telemetry silently off — the exact gap the shared predicate was meant to close. The README already lists these as honored, so the compose file was the odd one out.

docs/otel-api-observability/otel-collector.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - /var/lib/docker/containers/*/*-json.log
+    operators:
+      # 1. Unwrap Docker's json-file envelope:
+      #    {"log": "<the app's JSON line>\n", "stream": "stdout", "time": "..."}
+      #    This leaves the app's own JSON as a string in attributes.log.
+      - type: json_parser
+      # 2. Parse the app's JSON out of that string, so trace_id / span_id /
+      #    service_name become real attributes.
+      - type: json_parser
+        parse_from: attributes.log
```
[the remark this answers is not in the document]
sriramveeraghanta: Agreed — fixed in 5791392. Guarded the second parser:

```yaml
      - type: json_parser
        parse_from: attributes.log
        if: 'attributes.log != nil and attributes.log matches "^\\s*[{]"'
```

`filelog.include` matches every container on the host, so plaintext lines from postgres/redis/rabbitmq/minio hit the parser and, under the default `on_error: send`, logged a processing error per line. The `nil` check also covers the case where the first `json_parser` failed and left no `attributes.log`.

Also noted in a comment that narrowing `include` to the api/worker/beat container ids is the tidier option for anyone who knows them — this file is meant as a drop-in reference, so the guard is the safer default.

PR #9420, opened 2026-07-14
Payments

JoseACP:
### Description

Adds two modules on top of upstream Plane: an **expense ledger with budgets**, and a
**payroll** system. Both sit behind the per-workspace `payments` feature flag and are
**admin-only**. Everything is surfaced on one page (`/{workspaceSlug}/payments`), split
into tabs: Expenses · Employees · Payroll · Aguinaldo · Annual cost.

Nothing here moves real money — it is an internal ledger, not a payment gateway.

#### Expenses & budgets

- Budgets allocate an amount to a **category** for a period, optionally scoped to a project.
- **Spent is never stored**, it is aggregated from the ledger on read. A stored counter
  drifts away from its own rows the first time someone edits an expense.
- Invoices/receipts are attached as **library assets** (several per expense), so they reuse
  the existing presigned upload and the PDF/image viewers. Detaching a document does not
  delete the file.
- `GET budgets/summary/?from=&to=` returns budgeted vs spent per (category, currency).
  Cancelled expenses are excluded; pending ones are reported separately.

#### Payroll

- **An Employee is not a Plane user.** The people on payroll are not necessarily the people
  with accounts; tying the two would force a login for everyone paid and leak payroll into
  the member list.
- A salary belongs to an `(employee, office)` pair, so someone working for two companies
  holds **two concurrent salaries**.
- **A raise never overwrites.** It closes the running row (`effective_to`) and opens a new
  one, so the history stays answerable and the aguinaldo is computed on the rate that was
  actually in force.
- Adjustments (bonus / debt / support) carry their direction in the `kind`, not the sign —
  the amount is always positive, so a mistyped minus can't flip a bonus into a debt.
- **"Upcoming payments" is not a separate table**: it is the `PENDING` rows, so a scheduled
  payment and the payment it becomes are the same record and cannot disagree.
- **Aguinaldo** is computed, never stored (LFT art. 87):
  `daily_salary × office.aguinaldo_days × (days_worked_in_year / 365)`.
  Each office pays on its own terms — it may grant more than the legal 15 days, never less.

#### Permissions (breaking for existing payments users)

Payments and budgets were previously readable by members and guests. **They are now
admin-only** — every handler is gated to `ROLE.ADMIN` and the sidebar entry is hidden from
everyone else.

The **annual cost** report sits behind a second gate, and is hidden **even from workspace
admins** — because the person it must be hidden from *is* an admin. Therefore:

- There is **no endpoint that grants it**. An in-app toggle any admin could flip would be
  decoration, not a restriction. The grant is a `PayrollAccess` row, set from the shell.
- The API answers **404, not 403**. A 403 would confirm the report exists, which is itself
  the thing being kept quiet.

#### Money handling

- Amounts are `Decimal`, never float, and cross the wire as **strings** (`"2500.50"`).
  Handing a raw `Decimal` to DRF's JSON renderer silently turns it into a float and loses
  cents on a large ledger.
- **Amounts in different currencies are never summed.** Every total is grouped by currency.

#### Notable schema decisions

- Budget uniqueness needs **two partial constraints, not one**: Postgres treats `NULL` as
  distinct, so a single constraint spanning the nullable `project` would allow two
  workspace-level budgets for the same category and period, silently doubling the summary.
- Same reasoning for `unique_open_salary_per_office_when_not_deleted` — without it an
  employee could hold two open salaries in one office and every total would double.
- `Office` FKs use `CASCADE`, not `PROTECT`: `PROTECT` made deleting a whole workspace fail
  on the office it cascades into. An office that still pays people is guarded at the
  endpoint instead (409).

Migrations: `0127` (finance) · `0128` (expense documents) · `0129` (merge — this branch and
`preview` both grew migrations off `0126`) · `0130` (payroll) · `0131` (office FK).

Full documentation in the README (backend in depth, frontend summarized).

### Type of Change

- [x] Documentation update

> The **Improvement** box is checked because payments/budgets went from
> member-readable to admin-only. Non-admins who could previously open the page now get a
> 403 and no sidebar entry.

### Test Scenarios

Backend, exercised against a local Postgres through the real URLconf, permissions and
serializers (test client, not mocks):

- **Permissions**: a member gets 403 on expenses, budgets, employees, payroll, aguinaldo and
  the annual cost; an admin gets 200 on all but the annual cost.
- **The hidden report**: an **admin** (the "boss" case) gets **404** on `annual-cost/`;
  a user with the `PayrollAccess` grant gets 200.
- **Salary history**: a raise appends a row, closes the previous one the day before, and the
  old rate stays visible. A raise dated before the running salary is rejected (400). The
  database refuses a second open salary in the same office.
- **Two concurrent salaries** across two offices on the same employee.
- **Aguinaldo**: full-year employee → 18,000.00 MXN (uses the July raise, not the old rate);
  July hire → 184 days → 7,561.64 MXN; an office with 30 days pays on its own terms; a USD
  salary is reported in its own row and never folded into the MXN totals.
- **Budget summary**: cancelled and out-of-period expenses excluded; pending reported apart;
  MXN and USD in separate rows; money emitted as strings, and `0.10 + 0.20` sums to exactly
  `0.30`.
- **Duplicates**: a second budget for the same category+period → 409 (not a 500); a duplicate
  category name → 409.
- **Expense documents**: several per expense; a file from another workspace is rejected; an
  asset whose upload never completed is rejected; detaching keeps the file in the library;
  deleting the expense takes the links, not the files.

Frontend: typecheck and lint clean, all locales in sync.

**Not verified**: the browser upload path (bytes → MinIO) and the visual rendering — both
require a real browser session.


sriramveeraghanta:
This is not part of the roadmap and we are not accepting these features as part of Plane. Thanks for your contribution.

PR #9429, opened 2026-07-16
fix(api): return 400 for non-positive per_page instead of 500

### LINKED ISSUE #9426: [bug]: Work items API returns HTTP 500 for per_page=0 or negative values instead of a validation error
**Describe the bug**

The work items list endpoint validates the `per_page` query parameter inconsistently. Non numeric and oversized values are rejected with a clear 400 response, but zero and negative values crash with an unhandled 500.

**Steps to reproduce**

Call the public API work items list endpoint with different `per_page` values (Plane Cloud, tested 2026-07-16):

```
GET /api/v1/workspaces/{slug}/projects/{project_id}/issues/?per_page=abc
→ 400 {"detail": "Invalid per_page parameter."}

GET .../issues/?per_page=100000
→ 400 {"detail": "Invalid per_page value. Cannot exceed 1000."}

GET .../issues/?per_page=0
→ 500 {"error": "Something went wrong please try again later"}

GET .../issues/?per_page=-5
→ 500 {"error": "Something went wrong please try again later"}
```

**Expected behavior**

`per_page=0` and negative values should return a 400 with a message like the other invalid cases, for example "Invalid per_page value. Must be a positive integer."

**Why it matters**

A 500 tells the API consumer the server failed, so clients typically retry, when the real problem is invalid input that will never succeed. The existing validation already catches non numeric and too large values, so this looks like a small gap in the same check.

**Environment**

Plane Cloud (app.plane.so), public REST API v1, personal API token auth.

Syed-Ali-Abbas-Zaidi:
The work items list endpoint (and every other paginated v1 endpoint) validated per_page for non-numeric and oversized values but not for zero or negative ones. Those fell through get_per_page and crashed downstream in the paginator, surfacing as an unhelpful HTTP 500 that clients retry.

Reject per_page <= 0 in BasePaginator.get_per_page with a 400 ParseError ("Invalid per_page value. Must be a positive integer."), matching the existing non-numeric and too-large validation cases.

Add contract tests covering per_page=0, negative, non-numeric, oversized, and valid values on the work items list endpoint.

Closes #9426

### Type of Change

- [ ] Code refactoring
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for issue list pagination.
  * Invalid, non-numeric, non-positive, or excessively large `per_page` values now return a clear HTTP 400 error.
  * Valid pagination values continue to work as expected.

### DIFF
--- a/apps/api/plane/tests/contract/api/test_issues.py
+++ b/apps/api/plane/tests/contract/api/test_issues.py
@@ -94,3 +94,50 @@ class TestIssueListOrderByInjection:
             assert response.status_code == status.HTTP_200_OK, (
                 f"order_by={value!r} got {response.status_code}: {response.data!r}"
             )
+
+
+@pytest.mark.contract
+class TestIssueListPerPageValidation:
+    """Regression tests for the work-item list endpoint's ``per_page`` handling:
+    GET /api/v1/workspaces/{slug}/projects/{project_id}/issues/.
+
+    Non-numeric and oversized values already returned a clear 400, but zero and
+    negative values fell through validation and crashed downstream with a 500.
+    They should return a 400 like the other invalid cases.
+    """
+
+    def get_url(self, workspace_slug, project_id):
+        return f"/api/v1/workspaces/{workspace_slug}/projects/{project_id}/issues/"
+
+    @pytest.mark.django_db
+    @pytest.mark.parametrize("per_page", ["0", "-5"])
+    def test_non_positive_per_page_returns_400(self, api_key_client, workspace, project, issue, per_page):
+        """``per_page=0`` and negative values used to raise a 500; now a 400."""
+        url = self.get_url(workspace.slug, project.id)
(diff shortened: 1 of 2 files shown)


asish-singh:
Original reporter of #9426 here. I reran my repro today (2026-07-27) against the live API at api.plane.so and the bug still reproduces exactly as reported, `per_page=0` and `per_page=-5` both return HTTP 500 on the work items list endpoint, while `per_page=abc` returns 400 and `per_page=50` returns 200.

I read through the change and it looks right to me. The check lands in the same `get_per_page` path that already handles the non numeric and oversized cases, so the error behavior is now consistent across all invalid inputs, and the regression tests cover the two values from my report. Happy to rerun my live repro once this ships to confirm the 500 is gone. Thanks @Syed-Ali-Abbas-Zaidi for picking this up.

PR #9443, opened 2026-07-20
[WEB-8289] fix: scope draft-to-issue conversion to the draft owner

mguptahub:
## Summary

Fixes WEB-8289 (HIGH). `WorkspaceDraftIssueViewSet.create_draft_to_issue` resolved the draft with `get_queryset().filter(pk=draft_id)` — scoped only to the workspace, with **no `created_by` check** — while the decorator admits any workspace ADMIN/MEMBER.

Draft issues are **private to their creator** (the sibling `list` / `retrieve` / `partial_update` methods all filter `created_by=request.user`). So any workspace member could supply another user's `draft_id` and:
- convert that private draft into a real issue,
- **reassign (steal) the draft's file assets** to the new issue (`FileAsset.objects.filter(draft_issue_id=draft_id).update(issue_id=...)`), and
- **delete the victim's draft** (`draft_issue.delete()`).

## Fix

Scope the lookup to the owner and 404 otherwise:

```python
draft_issue = self.get_queryset().filter(pk=draft_id, created_by=request.user).first()
if not draft_issue:
    return Response({"error": "The required object does not exist."}, status=404)
```

This mirrors the exact pattern already used by `retrieve` / `partial_update`. The explicit null guard also removes a latent `AttributeError` on the subsequent `.project_id` access when no draft matches.

## Tests

New contract suite `plane/tests/contract/app/test_draft_to_issue_owner_scope_app.py` (**fail-before verified** — the security test returns 201 without the fix):
- A non-owner member converting another user's draft → **404**, and the draft + its file asset are left intact (not deleted, not reassigned).
- Positive control: the draft's owner can still convert their own draft (201, draft consumed, asset migrated to the new issue).

## Verification
- `ruff check`: clean
- `manage.py check` (test settings): no issues
- `pytest` (new suite): 2/2 green; fail-before verified

## Notes
- **Dedupe:** a second MED report shares this root cause and code path — should be closed as a duplicate of this on release. (Mapping on the ticket.)
- **EE:** EE does not override this endpoint, so this CE fix covers EE deployments too. A distinct, lower-severity owner-scope gap on the EE-only `DraftIssuePropertyValueEndpoint` (draft custom-field values, same-project) is being tracked as a separate follow-up, not part of this advisory.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted “draft-to-issue” conversion so only the draft’s owner can convert it.
  * Unauthorized attempts now return **404 Not Found** and do not alter the draft, its attached files, or create an issue.
  * Draft owners can still convert their drafts successfully, including transferring any attached files to the new issue.

* **Tests**
  * Added contract tests covering both unauthorized and authorized draft-to-issue conversion paths.

### DIFF
--- a/apps/api/plane/app/views/workspace/draft.py
+++ b/apps/api/plane/app/views/workspace/draft.py
@@ -204,7 +204,16 @@ class WorkspaceDraftIssueViewSet(BaseViewSet):
 
     @allow_permission(allowed_roles=[ROLE.ADMIN, ROLE.MEMBER], level="WORKSPACE")
     def create_draft_to_issue(self, request, slug, draft_id):
-        draft_issue = self.get_queryset().filter(pk=draft_id).first()
+        # Drafts are private to their creator; scope the lookup to the owner so a
+        # workspace member cannot convert/steal/delete another user's draft
+        # (GHSA-vfqm-7rh7-84hq). Mirrors retrieve/partial_update.
+        draft_issue = self.get_queryset().filter(pk=draft_id, created_by=request.user).first()
+
+        if not draft_issue:
+            return Response(
+                {"error": "The required object does not exist."},
+                status=status.HTTP_404_NOT_FOUND,
+            )
 
         if not draft_issue.project_id:
             return Response(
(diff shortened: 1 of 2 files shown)


mguptahub:
Thanks @copilot — both addressed in `7448c69`.

You're right: `Project.objects.create(created_by=create_user)` and `FileAsset.objects.create(created_by=create_user)` were silent no-ops because `BaseModel.save()` nulls `created_by` when there's no current request user under tests. Switched both to instantiate + `save(created_by_id=create_user.id)` — the same pattern already used for the draft — so the fixtures' ownership is real.

Note this didn't affect the test's correctness (the security assertions key off the *draft's* `created_by`, which was already set correctly, and the asset checks use `draft_issue_id`/`issue_id`), but the fixtures are now accurate. 2/2 tests still green.

sriramveeraghanta (approved):

PR #9444, opened 2026-07-20
[WEB-8291] fix: gate project invitation list/retrieve/destroy to project admins

mguptahub:
## Summary

Fixes the confirmed-vulnerable half of WEB-8291 (HIGH, CWE-862) — the **project-invitation-list authorization bypass**.

`ProjectInvitationsViewset` declared no `permission_classes` (inheriting `IsAuthenticated`) and only decorated `create` with `@allow_permission([ROLE.ADMIN])`. The default `list` / `retrieve` / `destroy` actions were therefore ungated, and `get_queryset` was scoped only by URL `slug` + `project_id` — not by the caller's membership. The serializer (`ProjectMemberInviteSerializer`, `fields="__all__"`) returns `email` and `token`.

**Impact (before fix):** any authenticated user could `GET /api/workspaces/<slug>/projects/<project_id>/invitations/` (and `/<pk>/`) for a project they don't belong to and read every pending invite's raw `token` + invitee `email` (re-exposing what #9305 stripped from the public path), and could `DELETE` arbitrary project invitations.

> The other vector in the advisory — **unauthenticated force-accept** — is already remediated: `ProjectJoinEndpoint.post` enforces token + auth + email-match (#9308) and the public GET uses the token-stripped serializer (#9305).

## Fix

Gate `list` / `retrieve` / `destroy` with `@allow_permission([ROLE.ADMIN])`, matching `create` (project admin, or a workspace-admin who is a project member — the decorator's existing semantics). This mirrors the workspace sibling `WorkspaceInvitationsViewset`, which was already admin-gated (`WorkSpaceAdminPermission`).

Chose the method decorator over a class-level `ProjectAdminPermission` deliberately: a class-level permission would have *tightened* `create` (blocking the workspace-admin-project-member fallback that `@allow_permission` allows), a behavior change. The decorator keeps all four actions consistent.

Also removes a pre-existing unused `User` import in the same file.

## Tests

New contract suite `plane/tests/contract/app/test_project_invite_list_scope_app.py` (**fail-before verified** — the 3 security tests return 200/204 without the fix):
- A workspace user who is a member of a *different* project is rejected with **403** on list / retrieve / delete, and the invitation is left intact.
- Positive control: an active project admin can still list invitations and sees the pending invite.

## Verification
- `ruff check`: clean
- `manage.py check` (test settings): no issues
- `pytest` (new suite): 4/4 green; fail-before verified

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted project invitation listing, viewing, and deletion to project administrators.
  * Prevented unauthorized workspace users from accessing invitations belonging to other projects.
  * Confirmed that project administrators can continue managing pending invitations.

* **Tests**
  * Added coverage for authorized and unauthorized invitation access scenarios.

### DIFF
--- a/apps/api/plane/app/views/project/invite.py
+++ b/apps/api/plane/app/views/project/invite.py
@@ -28,7 +28,6 @@ from plane.db.models import (
     ProjectMember,
     Workspace,
     ProjectMemberInvite,
-    User,
     WorkspaceMember,
     Project,
     ProjectUserProperty,
@@ -52,6 +51,23 @@ class ProjectInvitationsViewset(BaseViewSet):
             .select_related("project")
             .select_related("workspace", "workspace__owner")
         )
+
+    # GHSA-r68c-48rr-m67f: project invitations expose invitee email + raw token
+    # and allow deletion, so every action must be restricted to project admins
+    # (mirroring create). Without these, the default list/retrieve/destroy
+    # inherited only IsAuthenticated, letting any workspace user read/delete
+    # another project's invitations.
+    @allow_permission([ROLE.ADMIN])
+    def list(self, request, slug, project_id):
+        return super().list(request)
+
+    @allow_permission([ROLE.ADMIN])
+    def retrieve(self, request, slug, project_id, pk):
+        return super().retrieve(request)
+
+    @allow_permission([ROLE.ADMIN])
+    def destroy(self, request, slug, project_id, pk):
+        return super().destroy(request)
 
     @allow_permission([ROLE.ADMIN])
     def create(self, request, slug, project_id):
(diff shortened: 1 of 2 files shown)


mguptahub:
Good catch @copilot — that was a branching mistake, not intended scope. `web-8291` was accidentally cut from the `web-8289` branch instead of `preview`, so it carried the WEB-8289 draft-to-issue commits (which have their own PR, #9443).

Fixed by rebasing `web-8291` onto `preview` (`git rebase --onto origin/preview`), dropping the two WEB-8289 commits. This PR now contains **only** the WEB-8291 project-invitation authorization fix:
- `apps/api/plane/app/views/project/invite.py`
- `apps/api/plane/tests/contract/app/test_project_invite_list_scope_app.py`

The draft-to-issue change stays tracked/reviewed independently in #9443. 4/4 invite contract tests still green; PR is mergeable.

PR #9450, opened 2026-07-20
fix: track and subscribe inline assignees when work items are created via the external API

### LINKED ISSUE #9449: 🐛 Bug: Work items created via the external API with inline assignees never subscribe or notify the assignees

### Current behavior

Creating a work item through the external REST API (`POST /api/v1/workspaces/:slug/projects/:project_id/work-items/`) with assignees included in the create payload (`"assignees": [...]`) silently drops the assignee tracking: no assignee activity is recorded, the assignees are not auto-subscribed to the work item, and consequently they get no notification for the assignment — nor for anything that happens on the work item later, since they never became subscribers.

Assigning the same user in a separate `PATCH` call afterwards works correctly (activity + auto-subscribe + notification).

Root cause: `create_issue_activity` in `apps/api/plane/bgtasks/issue_activities_task.py` only calls `track_assignees` when the payload contains `assignee_ids` (the web app field name):

```python
if requested_data.get("assignee_ids") is not None:
```

The external API serializer uses `assignees` instead. `track_assignees` itself already reads both keys via `extract_ids(requested_data, "assignee_ids", "assignees")`, and the update path maps both field names to `track_assignees` — only the create gate misses the external API key.

Note: with #9307 the create call now dispatches `notification=True`, but that alone doesn't help this case — the notification fan-out iterates the created activities and the subscribers, and inline assignees on create produce neither.

### Steps to reproduce

1. `POST /api/v1/workspaces/:slug/projects/:project_id/work-items/` with `{"name": "Test", "assignees": ["<user-uuid>"]}` using an API token.
2. Check the work item's activity: there is no "added assignee" entry.
3. Check `issue_subscribers`: the assignee was not subscribed.
4. The assignee receives no in-app notification and no email, and won't receive any for future changes either.

### Environment

Self-hosted Community Edition, reproduced on v1.3.1; the gate is unchanged on current `preview`.

### Expected behavior

Creating a work item with inline assignees via the external API behaves like the web app: assignee activity is tracked, assignees are auto-subscribed, and they are notified.

stella-nova:
### Description

`create_issue_activity` only runs `track_assignees` when the create payload contains `assignee_ids` (the web app field name). Work items created through the external REST API (`/api/v1/...`) carry `assignees` instead, so inline assignees on create produce no assignee activity and are never auto-subscribed to the work item — meaning they get no notification for the assignment or for any later change. Assigning in a separate `PATCH` afterwards works, because the update path maps both field names to `track_assignees`, and `track_assignees` itself already reads both keys via `extract_ids(requested_data, "assignee_ids", "assignees")`. Only the create gate misses the external API key.

This PR accepts both `assignee_ids` and `assignees` in the create gate in `apps/api/plane/bgtasks/issue_activities_task.py`.

### Type of Change

### Test Scenarios

Unit tests added in `plane/tests/unit/bg_tasks/test_issue_activities_task.py`: creating with `assignee_ids`, creating with `assignees` (both must track the assignee activity and create the `IssueSubscriber`), and creating without assignees (no activities collected, no subscription, the "created" activity still recorded). Ran against PostgreSQL: 3 passed. Verified red/green: on the unpatched task exactly the `assignees` test fails. `ruff check` clean.

### References

Fixes #9449.

## Summary by CodeRabbit

* **Bug Fixes**
  * Issue creation now correctly tracks assignees and creates subscriptions when assignees are provided through either supported API format.
  * Issues created without assignees no longer generate unnecessary assignee activity or subscriptions.
  * Issue activity creation now handles missing, empty, or invalid assignment data safely.

### DIFF
--- a/apps/api/plane/bgtasks/issue_activities_task.py
+++ b/apps/api/plane/bgtasks/issue_activities_task.py
@@ -578,7 +578,7 @@ def create_issue_activity(
     issue_activity.actor_id = issue.created_by_id
     issue_activity.save(update_fields=["created_at", "actor_id"])
     requested_data = json.loads(requested_data) if requested_data is not None else None
-    if requested_data.get("assignee_ids") is not None:
+    if requested_data.get("assignee_ids") is not None or requested_data.get("assignees") is not None:
         track_assignees(
             requested_data,
             current_instance,
--- a/apps/api/plane/bgtasks/issue_activities_task.py
+++ b/apps/api/plane/bgtasks/issue_activities_task.py
@@ -578,7 +578,9 @@ def create_issue_activity(
     issue_activity.actor_id = issue.created_by_id
     issue_activity.save(update_fields=["created_at", "actor_id"])
     requested_data = json.loads(requested_data) if requested_data is not None else None
-    if requested_data.get("assignee_ids") is not None or requested_data.get("assignees") is not None:
+    if isinstance(requested_data, dict) and (
+        requested_data.get("assignee_ids") is not None or requested_data.get("assignees") is not None
+    ):
         track_assignees(
             requested_data,
             current_instance,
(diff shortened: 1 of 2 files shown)


stella-nova:
Addressed the CodeRabbit review feedback in 81eb686: `create_issue_activity` now checks `isinstance(requested_data, dict)` before the `.get()` calls, so a `None` payload (explicitly allowed by the preceding line) or a payload parsing to a JSON array skips assignee tracking instead of raising `AttributeError`. Added unit tests for both cases; they fail with `AttributeError` without the guard. Full file suite green (5 passed), ruff check/format clean.

Issue #720, opened 2023-04-05
[bug]: The initial run with docker compose failed


ajnasaboobacker:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

yarn install v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
[3/4] Linking dependencies...
warning "workspace-aggregator-3b7daf84-f8e7-4836-81de-ab25764bb7f7 > app > recharts@2.4.3" has unmet peer dependency "prop-types@^15.6.0".
warning "workspace-aggregator-3b7daf84-f8e7-4836-81de-ab25764bb7f7 > app > @blueprintjs/core > react-popper@1.3.11" has incorrect peer dependency "react@0.14.x || ^15.0.0 || ^16.0.0 || ^17.0.0".
warning "workspace-aggregator-3b7daf84-f8e7-4836-81de-ab25764bb7f7 > app > next-pwa > babel-loader@8.3.0" has unmet peer dependency "@babel/core@^7.0.0".
warning "workspace-aggregator-3b7daf84-f8e7-4836-81de-ab25764bb7f7 > app > next-pwa > babel-loader@8.3.0" has unmet peer dependency "webpack@>=2".
warning "workspace-aggregator-3b7daf84-f8e7-4836-81de-ab25764bb7f7 > app > next-pwa > clean-webpack-plugin@4.0.0" has unmet peer dependency "webpack@>=4.0.0 <6.0.0".
warning "workspace-aggregator-3b7daf84-f8e7-4836-81de-ab25764bb7f7 > app > next-pwa > terser-webpack-plugin@5.3.6" has unmet peer dependency "webpack@^5.1.0".
warning "workspace-aggregator-3b7daf84-f8e7-4836-81de-ab25764bb7f7 > app > next-pwa > workbox-webpack-plugin@6.5.4" has unmet peer dependency "webpack@^4.4.0 || ^5.9.0".
[...]

PR #9457, opened 2026-07-21
fix(security): stop logging OAuth bearer tokens and emails in clear text

sriramveeraghanta:
### Description

Closes the three open high-severity CodeQL `py/clear-text-logging-sensitive-data` alerts (#720, #721, #722) in the authentication module.

The `plane.authentication` logger runs at **INFO in production** with a `pythonjsonlogger` formatter that serialises `extra` fields, so anything handed to these log calls is written to the production log stream verbatim. That makes one of these a genuine secret leak, not just a static-analysis complaint.

**`adapter/oauth.py` — live OAuth bearer token (the real leak).** `get_user_response()` logged the outgoing request headers on failure, and those headers carry the provider access token as a bearer credential. The dict also had no diagnostic value — it has the identical shape on every call. Replaced with the provider name and the response status code, which actually say *which* IdP failed and *how*: `401` token rejected, `403` scope/rate limit, `404` wrong userinfo URL, `None` connection/TLS/timeout error. Previously the log line couldn't even tell you which of the four providers it came from.

**`adapter/base.py` — raw email in the message.** `sanitize_email()` interpolated the rejected address into the log message. It was the only f-string logger call in the whole package; every sibling logs a constant string and confines the email to the exception payload. That payload is left untouched, so the caller still receives the address — only the log stream stops seeing it.

**`provider/oauth/github.py` — email logged at DEBUG.** `set_user_data()` logged the resolved email address. The record cannot be emitted today (the console handler is pinned to `INFO`), but it would begin leaking PII the moment anyone attached a DEBUG handler. Removed rather than downgraded.

**Splitting the client-secret config reads.** The three providers now read their client secret in a `get_configuration_value()` call of its own instead of batching it with their non-secret settings. That function returns every requested value in a single tuple, which makes the secret indistinguishable from the values beside it to anything tracking where it flows — and `GITEA_HOST`/`GITLAB_HOST` build `userinfo_url`, so the shared tuple marked every response fetched with that URL as secret-derived. That is what put `user_login` on alert #721.

This matters for *how* the alert gets fixed: `user_login` is a public GitHub handle and the only per-event signal on the not-in-org line — the field that tells an admin which account was rejected. Splitting the reads keeps it, rather than deleting real audit data to appease a false positive. No suppression comments were added; all three alerts clear because the sensitive data genuinely stops reaching the sinks.

The split is structural only — `get_configuration_value()` resolves values by key in request order (`instance_value.py:23-39`), so every provider ends up with exactly the same configuration. Its signature and all other call sites are untouched. Cost is one extra config read per OAuth init.

### Type of Change
- [ ] Code refactoring
- [ ] Documentation update

### Test Scenarios

Added `apps/api/plane/tests/unit/authentication/test_adapter_logging.py` — 15 tests. Verified they are not vacuous: **7 of them fail against the pre-fix source**, the 2 that pass being the guard tests asserting unchanged behaviour.

Credential hygiene (assert against each record's full `__dict__`, not just the formatted message, since `extra` values land as record attributes):
- Access token and the string `Bearer` appear in **no** emitted record when the userinfo request fails
- `provider` and `status_code` are present as the replacement diagnostic; `status_code` is `None` on a connection error with no response
- An invalid email is never echoed into the log, but **is** still delivered to the caller via the exception payload
- `organization_id` absent from the not-in-org record while `user_login` is retained
- Email not logged on the GitHub success path

Behaviour preservation for the config split:
- Each of GitHub / Gitea / GitLab still resolves `client_id`, `client_secret` and its host or org id identically
- Gitea and GitLab still build `token_url` / `userinfo_url` from the host, with Gitea's trailing-slash normalisation intact
- The not-configured guard still raises for each provider when its client secret is missing (parametrised)

Manual verification run:
- `ruff check apps/api` — no new findings (3 pre-existing `F401`s in `app/views/issue/sub_issue.py` and `app/views/project/invite.py`, untouched by this PR and auto-fixed by CI's `--fix`)
- `pytest plane/tests/unit` — **341 passed**
- `pytest plane/tests/contract` — 8 pre-existing failures in `test_authentication.py` magic-link tests. These are **not** caused by this branch: the untouched base fails **10** of the same tests, a superset, and each passes in isolation, so they are order/state dependent.

Worth a reviewer's eye: CodeQL only runs in CI here, so the alert-clearing itself is confirmed by this PR's own CodeQL run rather than locally.

### References

- CodeQL alerts [#720](https://github.com/makeplane/plane/security/code-scanning/720), [#721](https://github.com/makeplane/plane/security/code-scanning/721), [#722](https://github.com/makeplane/plane/security/code-scanning/722) (`py/clear-text-logging-sensitive-data`, high)

---

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication logging to prevent access tokens, email addresses, organization IDs, and other sensitive details from appearing in logs.
  * Authentication diagnostics now retain safe context, such as the provider and response status.
  * OAuth provider configuration continues to load correctly while handling client secrets more securely.

* **Tests**
  * Added coverage to verify sensitive information is excluded from authentication logs and configuration remains functional.

### DIFF
--- a/apps/api/plane/authentication/adapter/base.py
+++ b/apps/api/plane/authentication/adapter/base.py
@@ -78,7 +78,7 @@ class Adapter:
         try:
             validate_email(email)
         except ValidationError:
-            self.logger.warning(f"Email is not valid: {email}")
+            self.logger.warning("Email is not valid", extra={"provider": self.provider})
             raise AuthenticationException(
                 error_code=AUTHENTICATION_ERROR_CODES["INVALID_EMAIL"],
                 error_message="INVALID_EMAIL",
--- a/apps/api/plane/authentication/adapter/oauth.py
+++ b/apps/api/plane/authentication/adapter/oauth.py
@@ -89,11 +89,15 @@ class OauthAdapter(Adapter):
             response = requests.get(self.get_user_info_url(), headers=headers)
             response.raise_for_status()
             return response.json()
-        except requests.RequestException:
+        except requests.RequestException as e:
+            # Never log `headers` - it carries the provider access token as a
+            # bearer credential. The provider and status code are the useful
+            # signals and carry nothing sensitive.
             self.logger.warning(
                 "Error getting user response",
                 extra={
-                    "headers": headers,
+                    "provider": self.provider,
+                    "status_code": getattr(e.response, "status_code", None),
                 },
             )
             code = self.authentication_error_code()
--- a/apps/api/plane/authentication/provider/oauth/gitea.py
+++ b/apps/api/plane/authentication/provider/oauth/gitea.py
@@ -22,19 +22,28 @@ class GiteaOAuthProvider(OauthAdapter):
     scope = "openid email profile read:user"
 
     def __init__(self, request, code=None, state=None, callback=None):
-        (GITEA_CLIENT_ID, GITEA_CLIENT_SECRET, GITEA_HOST) = get_configuration_value(
+        # The client secret is read separately from the non-secret settings on
+        # purpose. Batching them into one call returns them in a single tuple,
+        # which makes the secret indistinguishable from the values around it for
+        # anything tracking where it flows - GITEA_HOST builds `userinfo_url`, so
+        # a shared tuple marks every response fetched with that URL as secret.
+        (GITEA_CLIENT_ID, GITEA_HOST) = get_configuration_value(
             [
                 {
                     "key": "GITEA_CLIENT_ID",
                     "default": os.environ.get("GITEA_CLIENT_ID"),
                 },
                 {
-                    "key": "GITEA_CLIENT_SECRET",
-                    "default": os.environ.get("GITEA_CLIENT_SECRET"),
-                },
-                {
                     "key": "GITEA_HOST",
                     "default": os.environ.get("GITEA_HOST"),
+                },
+            ]
+        )
+        (GITEA_CLIENT_SECRET,) = get_configuration_value(
+            [
+                {
+                    "key": "GITEA_CLIENT_SECRET",
+                    "default": os.environ.get("GITEA_CLIENT_SECRET"),
                 },
             ]
         )
--- a/apps/api/plane/authentication/provider/oauth/github.py
+++ b/apps/api/plane/authentication/provider/oauth/github.py
@@ -31,18 +31,24 @@ class GitHubOAuthProvider(OauthAdapter):
     organization_scope = "read:org"
 
     def __init__(self, request, code=None, state=None, callback=None):
-        GITHUB_CLIENT_ID, GITHUB_CLIENT_SECRET, GITHUB_ORGANIZATION_ID = get_configuration_value([
+        # The client secret is read separately from the non-secret settings on
+        # purpose. Batching them into one call returns them in a single tuple,
+        # which makes the secret indistinguishable from the values around it for
+        # anything tracking where it flows.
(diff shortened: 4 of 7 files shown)


sriramveeraghanta:
## CodeQL verification (run locally — the PR check can't confirm this)

The `Analyze (python)` check passing does **not** confirm these alerts cleared. PR-ref analyses are diff-informed and report `results=0` regardless — every recent PR on this repo shows `results=0`, including ones that fixed nothing. Only `refs/heads/preview` carries the real alert set. So I verified locally with the CodeQL CLI instead.

**Setup:** CodeQL bundle `v2.26.1`, `codeql-suites/python-code-scanning.qls` (the default suite CI uses, since `codeql.yml` sets no custom `queries:`), source root `apps/api`, run against both `origin/preview` and this branch.

**The harness reproduces CI exactly** — the base scan finds 5 total alerts, matching GitHub's own reported `results=5` for `refs/heads/preview`, and pins the three target alerts to the same lines the alerts cite (`base.py:81`, `github.py:161`, `oauth.py:95`).

| | base (`origin/preview`) | this branch |
|---|---|---|
| Total alerts (default suite) | 5 | **2** |
| `py/clear-text-logging-sensitive-data` | 3 | **0** |

Resolved:
```
py/clear-text-logging-sensitive-data  plane/authentication/adapter/base.py:81
py/clear-text-logging-sensitive-data  plane/authentication/adapter/oauth.py:95
py/clear-text-logging-sensitive-data  plane/authentication/provider/oauth/github.py:161
```

Newly introduced: **none**. No suppression comments were added — the alerts clear because the sensitive data genuinely stops reaching the sinks.

---

### Follow-ups found while auditing, deliberately not in this PR

An audit of every log sink in `apps/api` turned up a **higher-severity leak that CodeQL does not flag**, left out to keep this PR scoped:

- **`plane/middleware/logger.py:63-66`** — `RequestLoggerMiddleware` is global and logs `request.get_full_path()` (the message) and `request.path` (the `extra`) at INFO. The password-reset route carries its token as a **path segment** (`auth/reset-password/<uidb64>/<token>/`), so live `PasswordResetTokenGenerator` tokens are written to production logs; OAuth `code`/`state` arrive there via the query string. Note the sibling `APITokenLogMiddleware` in the same file already redacts headers and HMACs the API key — this middleware has no equivalent. Both sinks would need scrubbing, and `resolver_match.kwargs` is a cleaner basis for masking than hardcoded route prefixes.
- **`plane/bgtasks/user_email_update_task.py:110`** — logs the address at INFO, and is called once with the old and once with the new email, yielding an explicit old→new correlation for a single user.
- **`plane/app/views/user/base.py:203`** — logs a cache key of the form `magic_email_update_{user.id}_{new_email}`.

Happy to take these in a follow-up.

sriramveeraghanta:
Closing as superseded: the cleartext-logging security issue this PR addressed was already fixed in #9505 ("fix: resolve open CodeQL security alerts", merged 2026-07-29), which landed on `preview`.

Verification against `origin/preview`:
- `adapter/oauth.py` no longer logs `headers` (which carried the bearer token)
- `adapter/base.py` no longer logs the raw email
- `provider/oauth/github.py` no longer logs `organization_id` or user login

The remaining conflicts on this branch come from those same lines being edited differently by #9505. The security goal is met, so this PR is redundant.

PR #9491, opened 2026-07-29
fix: handle empty EMAIL_PORT in test-email endpoint (HTTP 500 → 400)

### LINKED ISSUE #9472: [bug]: test email not sending while setting up SMTP

### Current behavior

When setting up SMTP for the first time under Instance Admin (God Mode) → Settings → Email, if you fill in the SMTP fields and click "Send test email" without clicking "Save Changes" first, the request fails with a 500 error instead of a validation message.

Backend log shows:
ValueError: invalid literal for int() with base 10: ''
  at plane/license/api/views/configuration.py, line 110, in post
    port=int(EMAIL_PORT)

The test-email endpoint (`/api/instances/email-credentials-check/`) appears to be reading EMAIL_PORT (and presumably the other SMTP fields) from the last saved instance configuration rather than the values currently typed into the form. Since nothing had been saved yet, EMAIL_PORT was an empty string, and int('') threw an unhandled ValueError, causing a 500.

What I expected: either the test should validate against the unsaved form values (so first-time setup works without a separate save step), or if it's intentional that test-email only works against saved config, the UI should block the test / show a clear "please save your settings first" message instead of silently failing with a 500.

### Steps to reproduce

1. Go to Instance Admin (God Mode) → Settings → Email on a fresh instance with no email config saved yet.
2. Fill in SMTP Host, Username, Password, Port, and Sender Email.
3. Without clicking "Save Changes", click "Send test email".
4. Request to POST /api/instances/email-credentials-check/ returns 500.
5. Backend logs show ValueError: invalid literal for int() with base 10: '' at configuration.py line 110.
6. Click "Save Changes", then click "Send test email" again — this time it succeeds.

### Environment

Production

### Browser

Google Chrome

### Variant

Self-hosted

### Version

v1.3.1

eltypical:
## Summary

Fixes #9472

### Bug

On a fresh Plane installation with `SKIP_ENV_VAR=True`, the `EMAIL_PORT` row exists in `InstanceConfiguration` with value `""` (empty string) because no SMTP settings have been saved yet.

Clicking **Send Test Email** before saving the form calls `EmailCredentialCheckEndpoint.post()`, which reaches:

```python
connection = get_connection(
    port=int(EMAIL_PORT),   # int("") → ValueError → HTTP 500
    ...
)
```

The unhandled `ValueError` propagates as an opaque HTTP 500, giving the user no indication of what went wrong.

### Root Cause

`get_email_configuration()` reads `EMAIL_PORT` from the DB via `get_configuration_value()`. When `SKIP_ENV_VAR=True` and the row exists with `""`, the function returns `""` instead of falling back to the env-var default (`587`). The call site did not guard against this before attempting `int()`.

### Fix

Wrap the `int(EMAIL_PORT)` conversion in a `try/except (ValueError, TypeError)` block that returns **HTTP 400** with a descriptive message, consistent with every other error branch in this endpoint:

```python
try:
    smtp_port = int(EMAIL_PORT)
except (ValueError, TypeError):
    return Response(
        {
            "error": "Email port is not configured. "
                     "Please save your email settings before sending a test email."
        },
        status=status.HTTP_400_BAD_REQUEST,
    )
```

### Tests

Four pytest unit tests added in `apps/api/tests/test_email_credential_check.py`:

| Test | Scenario |
|---|---|
| `test_returns_400_for_empty_email_port` | `EMAIL_PORT = ""` — the exact regression case |
| `test_returns_400_for_none_email_port` | `EMAIL_PORT = None` — missing DB row |
| `test_returns_400_for_non_numeric_email_port` | `EMAIL_PORT = "not-a-port"` — corrupt value |
| `test_returns_400_when_receiver_email_is_missing` | Pre-existing guard still works |

All tests mock `get_email_configuration` at the call site and assert `HTTP 400` + `"error"` key in the response — no network or DB required.

### Change surface

- `apps/api/plane/license/api/views/configuration.py` — 8-line guard added before `get_connection()`; `port=smtp_port` replaces inline `int()` call
- `apps/api/tests/test_email_credential_check.py` — new regression test file

## Summary by CodeRabbit

* **Bug Fixes**
  * Email credential checks now validate the configured email port before attempting an SMTP connection.
  * Invalid or missing port values return a clear HTTP 400 error instead of an internal server error.

* **Tests**
  * Added regression coverage for empty, null, and non-numeric email port configurations.
  * Confirmed requests without a recipient email still return an HTTP 400 with an error message.

### DIFF
--- a/apps/api/plane/license/api/views/configuration.py
+++ b/apps/api/plane/license/api/views/configuration.py
@@ -104,10 +104,24 @@ class EmailCredentialCheckEndpoint(BaseAPIView):
             EMAIL_FROM,
         ) = get_email_configuration()
 
+        # Guard: EMAIL_PORT may be an empty string when SMTP settings have not
+        # been saved yet (fresh install with SKIP_ENV_VAR=True). Attempting
+        # int("") raises an unhandled ValueError → HTTP 500 (Issue #9472).
+        try:
+            smtp_port = int(EMAIL_PORT)
+        except (ValueError, TypeError):
+            return Response(
+                {
+                    "error": "Email port is not configured. "
+                    "Please save your email settings before sending a test email."
+                },
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
         # Configure all the connections
         connection = get_connection(
             host=EMAIL_HOST,
-            port=int(EMAIL_PORT),
+            port=smtp_port,
             username=EMAIL_HOST_USER,
             password=EMAIL_HOST_PASSWORD,
             use_tls=EMAIL_USE_TLS == "1",
--- a/apps/api/tests/test_email_credential_check.py
+++ b/apps/api/tests/test_email_credential_check.py
@@ -1,0 +1,81 @@
+"""
+Regression tests for EmailCredentialCheckEndpoint (Issue #9472).
+
+When EMAIL_PORT is an empty string (SMTP settings not yet saved on a fresh
+Plane install with SKIP_ENV_VAR=True), int("") raised an unhandled
+ValueError that propagated as HTTP 500.  These tests verify the endpoint
+returns HTTP 400 with a descriptive error message instead.
+"""
+import pytest
+from unittest.mock import patch
+from rest_framework.test import APIRequestFactory
+from rest_framework import status
+
+
+@pytest.mark.unit
+class TestEmailCredentialCheckEmptyPort:
+    """Guards around invalid EMAIL_PORT values (Issue #9472)."""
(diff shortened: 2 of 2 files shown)


eltypical:
Update pushed to the PR branch:

- Tests now assert the exact guidance: "...save your email settings before sending a test email" for invalid/missing EMAIL_PORT in EmailCredentialCheckEndpoint.

This keeps the user-facing message stable while preventing the original 500 regression from creeping back in.

PR #9288, opened 2026-06-22
[WEB-7776] fix(security): scope FileAsset queries to prevent cross-project IDOR (Cluster F)

mguptahub:
## Summary

Multiple asset endpoints were missing project-level scoping on `FileAsset` queryset filters, allowing authenticated users to access, mark-uploaded, or restore assets belonging to other projects or workspaces they are not members of.

### Fixes applied

| Priority | Endpoint | File | Fix |
|:---------|:---------|:-----|:----|
| P0 | `ProjectBulkAssetEndpoint.post` | `app/views/asset/v2.py` | Add `project_id=project_id` to asset filter |
| P0 | `EntityAssetEndpoint.get` | `space/views/asset.py` | Add `project_id=deploy_board.project_id` |
| P0 | `EntityAssetEndpoint.patch` | `space/views/asset.py` | Add `project_id=deploy_board.project_id` |
| P1 | `AssetRestoreEndpoint.post` | `space/views/asset.py` | Add `project_id=deploy_board.project_id` |
| P1 | `FileAssetEndpoint` V1 (get/post/delete) | `app/views/asset/base.py` | Add `WorkspaceMember` membership check |
| P1 | `FileAssetViewSet.restore` V1 | `app/views/asset/base.py` | Add `WorkspaceMember` membership check |
| P2 | `WorkspaceFileAssetEndpoint.post` | `app/views/asset/v2.py` | Gate `WORKSPACE_LOGO` upload on `ROLE.ADMIN` |
| P2 | `DuplicateAssetEndpoint.post` | `app/views/asset/v2.py` | Restrict source asset lookup to same workspace |

### Advisories addressed

GHSA-r2hw (critical), GHSA-jh4v (high), GHSA-8688 (high), GHSA-3hrj, GHSA-3892, GHSA-3ggg, GHSA-gcpp, GHSA-p57q, GHSA-c68q, GHSA-8chr, GHSA-58qm, GHSA-wrrw, GHSA-j4mj, GHSA-85h2, GHSA-29q3, GHSA-mwh2, GHSA-xrpv and related duplicates.

## Test plan

- [ ] Upload an asset to Project A, then attempt `POST /api/v1/workspaces/{slug}/projects/{projectB_id}/bulk-asset-save/` with the asset ID — should return 404
- [ ] On a public board, attempt `PATCH /spaces/{anchor}/assets/{asset_from_other_project}/` — should return 404
[...]

PR #9495, opened 2026-07-29
[SECUR-242] fix(api): scope bulk-asset associate by uploader, not project_id (regression from #9288)

mguptahub:
## Problem

**Release-blocking regression in v1.4.0-rc1.** Creating a project from the frontend fails at the "enable features" step:

```
POST /api/assets/v2/workspaces/<slug>/projects/<project_id>/<entity_id>/bulk/
→ 404 {"error": "The requested asset could not be found."}
```

## Root cause

#9288 (WEB-7776, "scope FileAsset queries to prevent cross-project IDOR — Cluster F") added `project_id=project_id` to the lookup in `ProjectBulkAssetEndpoint.post`:

```python
assets = FileAsset.objects.filter(id__in=asset_ids, workspace__slug=slug, project_id=project_id)
```

But this endpoint's job is to **associate** freshly-uploaded assets with a project/entity. A cover/feature asset uploaded during project creation still has `project_id=NULL` at this point — in fact this endpoint is what *sets* it (`assets.update(project_id=project_id)` for `PROJECT_COVER`). Requiring `project_id` to already match means the asset is never found → 404. `master` (v1.3.1) had no such filter, so the flow worked there.

Present in both `preview` and `canary` → ships in v1.4.0 unless patched.

## Fix

Scope by `created_by=request.user` (ownership) instead of `project_id=project_id` (association state):

```python
assets = FileAsset.objects.filter(id__in=asset_ids, workspace__slug=slug, created_by=request.user)
```

This **still closes the IDOR** #9288 targeted — a caller can only touch assets they uploaded, and `@allow_permission([ADMIN, MEMBER, GUEST])` already scopes them to the target project — and it is **stricter than the original master code** (which had no ownership check at all), while allowing a not-yet-associated asset to be linked.

## Testing
- ruff lint + format clean.
- Needs verification on redeploy: project creation → enable-features step should complete (204, not 404). ⚠️ **Cherry-pick to `canary` for v1.4.0-rc2** (the RC is already deployed with the bug).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved bulk asset association security by restricting selectable assets to those uploaded by the requester within the target workspace.
  - Updated the bulk association behavior to support selecting assets that are either already linked to the project or not yet linked, when the requester has the required workspace access.

### DIFF
--- a/apps/api/plane/app/views/asset/v2.py
+++ b/apps/api/plane/app/views/asset/v2.py
@@ -708,8 +708,14 @@ class ProjectBulkAssetEndpoint(BaseAPIView):
         if not asset_ids:
             return Response({"error": "No asset ids provided."}, status=status.HTTP_400_BAD_REQUEST)
 
-        # get the asset id — scope to the project to prevent cross-project IDOR
-        assets = FileAsset.objects.filter(id__in=asset_ids, workspace__slug=slug, project_id=project_id)
+        # Scope to the requester's own uploads in this workspace. This endpoint's job is
+        # to *associate* freshly-uploaded assets with a project/entity, so the assets are
+        # not yet project-scoped at this point (e.g. a cover uploaded during project
+        # creation has project_id=NULL until this call sets it). Filtering by
+        # project_id=project_id here 404s that flow; scoping by created_by still prevents
+        # cross-project/user IDOR (a caller can only touch assets they uploaded, and the
+        # @allow_permission decorator already scopes them to this project).
+        assets = FileAsset.objects.filter(id__in=asset_ids, workspace__slug=slug, created_by=request.user)
 
         # Get the first asset
         asset = assets.first()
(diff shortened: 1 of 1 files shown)


mguptahub:
Thanks @coderabbitai — valid, fixed in f5a5163. Scoping by `created_by` alone did let a caller move their own asset in from another project via the `PROJECT_COVER`/`ISSUE_DESCRIPTION` update branches.

Now bounded to **unassociated or already-in-this-project** assets, in addition to the uploader check:

```python
assets = FileAsset.objects.filter(
    id__in=asset_ids,
    workspace__slug=slug,
    created_by=request.user,
).filter(Q(project_id=project_id) | Q(project_id__isnull=True))
```

- freshly-uploaded asset (`project_id=NULL` during project creation) → still links ✅ (fixes the 404)
- asset already in this project → re-linkable ✅
- asset belonging to a **different** project → rejected ✅ (no cross-project move)
- `created_by` + `@allow_permission` keep it to the caller's own uploads within a project they're a member of.

pratapalakshmi (approved):

PR #9522, opened 2026-08-01
fix: unescape HTML entities in plain-text email bodies

### LINKED ISSUE #9497: Bug: Workspace invitation plain-text fallback link contains escaped query separators

### Current behavior

```markdown
Workspace invitation emails include a fallback text link under “If the button doesn’t work, copy and paste this link into your browser”.

In the generated plain-text email body, the URL query separators are HTML-escaped as `&amp;` instead of plain `&`.

Example:

```text
https://example.com/workspace-invitations/?invitation_id=<id>&amp;slug=<slug>&amp;token=<token>
```

When this visible fallback link is copied and opened in a browser, the frontend receives query parameters like `amp;slug` and `amp;token` instead of `slug` and `token`.

As a result, the workspace invitation page stays on the loading screen and no request is made to `/api/workspaces/<slug>/invitations/<invitation_id>/`.

I searched existing issues for `&amp;slug`, `workspace-invitations amp;`, and `copy and paste this link invitation`, but did not find this exact issue.
```

### Steps to reproduce

1. Run Plane self-hosted with SMTP enabled.
2. Invite a user to a workspace.
3. Open the received workspace invitation email.
4. Copy the visible fallback URL from "If the button doesn't work, copy and paste this link into your browser".
5. Paste that URL into a browser.
6. The invitation page stays on the loading screen.

Expected behavior:

The copyable/plain-text fallback URL should contain plain query separators:

    https://example.com/workspace-invitations/?invitation_id=<id>&slug=<slug>&token=<token>

Opening the copied fallback link should load the invitation details and allow the invitee to accept or decline the workspace invitation.

Additional investigation:

The workspace invitation task appears to build the raw URL correctly:

    relative_link = (
        f"/workspace-invitations/?invitation_id={workspace_member_invite.id}&slug={workspace.slug}&token={token}"
    )
    abs_url = str(current_site) + relative_link

The workspace invitation email template then renders `{{abs_url}}` in HTML. HTML escaping is expected there, especially inside `href` attributes.

However, the plain-text email body is generated from the already-rendered HTML:

    html_content = render_to_string("emails/invitations/workspace_invitation.html", context)
    text_content = generate_plain_text_from_html(html_content)

The current `generate_plain_text_from_html()` implementation strips tags but does not decode HTML entities, so `&amp;` remains in the plain-text output.

This seems related to PR #8535:
https://github.com/makeplane/plane/pull/8535

Possible fix: decode HTML entities when generating the plain-text email body, for example by applying `html.unescape()` after stripping HTML tags.

### Environment

Production

### Browser

Google Chrome

### Variant

Self-hosted

### Version

v1.3.1

### PR DESCRIPTION
**Pull request by okxint**
`django.utils.html.strip_tags()` strips HTML markup but leaves character entities (`&amp;`, `&lt;`, etc.) in the output as literals. In invitation and authentication emails, URLs are rendered inside HTML anchor tags — so `&` in query strings becomes `&amp;` in the HTML source. After stripping tags the entity survives, producing fallback links like:

```
https://app.plane.so/workspace-invitations/?invitation_id=…&amp;slug=…
```

Email clients that display the plain-text part pass this literally to the browser, which sees `amp;slug` and `amp;token` instead of `slug` and `token`. The invitation page never fetches `/api/workspaces/<slug>/invitations/<id>/` and stays on the loading screen indefinitely.

Wrapping the result in `html.unescape()` decodes all entities before the text is stored or sent. Applied consistently across every email background task that follows the same pattern.

Fixes #9497

### DIFF
--- a/apiserver/plane/bgtasks/forgot_password_task.py
+++ b/apiserver/plane/bgtasks/forgot_password_task.py
@@ -1,4 +1,5 @@
 # Python imports
+import html
 import logging
 
 # Third party imports
@@ -43,7 +44,7 @@ def forgot_password(first_name, email, uidb64, token, current_site):
 
         html_content = render_to_string("emails/auth/forgot_password.html", context)
 
-        text_content = strip_tags(html_content)
+        text_content = html.unescape(strip_tags(html_content))
 
         connection = get_connection(
             host=EMAIL_HOST,
--- a/apiserver/plane/bgtasks/magic_link_code_task.py
+++ b/apiserver/plane/bgtasks/magic_link_code_task.py
@@ -1,4 +1,5 @@
 # Python imports
+import html
 import logging
 
 # Third party imports
@@ -33,7 +34,7 @@ def magic_link(email, key, token, current_site):
         context = {"code": token, "email": email}
 
         html_content = render_to_string("emails/auth/magic_signin.html", context)
-        text_content = strip_tags(html_content)
+        text_content = html.unescape(strip_tags(html_content))
 
         connection = get_connection(
             host=EMAIL_HOST,
--- a/apiserver/plane/bgtasks/project_invitation_task.py
+++ b/apiserver/plane/bgtasks/project_invitation_task.py
@@ -1,4 +1,5 @@
 # Python imports
+import html
 import logging
 
 # Third party imports
@@ -41,7 +42,7 @@ def project_invitation(email, project_id, token, current_site, invitor):
             "emails/invitations/project_invitation.html", context
         )
 
-        text_content = strip_tags(html_content)
+        text_content = html.unescape(strip_tags(html_content))
 
         project_member_invite.message = text_content
         project_member_invite.save()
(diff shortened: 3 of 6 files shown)


sriramveeraghanta:
Thanks for the detailed write-up here — the diagnosis is exactly right, and it matches what we confirmed independently. `django.utils.html.strip_tags()` runs its `MLStripper` with `convert_charrefs=False` and re-emits entity refs in `handle_entityref`, so `&amp;` survives into the plain-text part and the fallback invite link arrives as `amp;slug` / `amp;token`.

Closing this one for reasons that aren't about the correctness of the fix:

1. It targets `develop`, which we no longer ship from — the last commit there is `22836ea0` (2025-01-20). The default branch is `preview`.
2. It patches `apiserver/plane/bgtasks/*`, a path that no longer exists. #8535 routed every email task through a shared `generate_plain_text_from_html()` helper in `apps/api/plane/utils/email.py`, so the fix now belongs in that one place rather than per call site. As written this PR would also have missed `email_notification_task`, `project_add_user_email_task`, `user_email_update_task`, `webhook_task`, and `analytic_plot_export`.

#9805 carries the same fix against `preview` in the shared helper, with a unit test — we'll track #9497 there.

Appreciate you finding this and digging into the root cause.

PR #9531, opened 2026-08-03
[SECUR-243] fix(security): scope issue comment + relation endpoints to the URL project

mguptahub:
## Summary

Closes a cross-tenant IDOR (CWE-639) on three issue sub-resource write handlers. `ProjectEntityPermission` and the `allow_permission` decorator validate only that the caller is an active member of the URL's `project_id` — neither validates that the sibling `issue_id` path parameter belongs to that project or workspace. Any authenticated user who is a member of *one* project on the instance could therefore reach issues in other projects and other workspaces by passing their UUID in the URL.

**HIGH** (`CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:L/A:N`), reported by @Avinash9102. Ticket: SECUR-243 (carries the advisory mapping).

> **This is present in released v1.4.0** — it was not part of the v1.4.0 security batch. Verified against the `v1.4.0` tag.

## What was wrong

| Handler | Bug | Impact |
|:--|:--|:--|
| `IssueCommentViewSet.create` | `Issue.objects.get(pk=issue_id)` — no workspace/project filter | **Cross-tenant read.** The 201 response serializes the foreign issue's full `issue_detail` — `name`, `description_json`, `description_html`, `priority`, `start_date`, `target_date`, `sequence_id` — back to the caller, and plants a comment on an issue in a workspace they never joined |
| `IssueRelationViewSet.create` | Request-body `issues` were workspace-scoped, but the URL `issue_id` used as the *other* side of the relation was not | Cross-tenant write: arbitrary foreign issue linked as the counterpart. The earlier scoping fix covered one side of the relationship and missed the other |
| `IssueRelationViewSet.remove_relation` | Filtered `IssueRelation` on `workspace__slug` only | Found while reviewing the two above. A member of one project could **delete** relations between two issues of a sibling project in the same workspace |

## The fix

All three bind the lookup to `pk=issue_id, workspace__slug=slug, project_id=project_id` and return 404 otherwise — the pattern already established by `sub_issue.py`'s `get()` and `PagesDescriptionViewSet.retrieve`.

For `remove_relation`, only the **URL issue** is project-bound; the `IssueRelation` row stays workspace-scoped on purpose, because relations legitimately span projects and either participant's project must be able to remove them.

Also fixes a pre-existing 500 in `remove_relation`: it did `.first()` and then called `.delete()` on `None` (`AttributeError` → generic 500) when no relation matched. Now a 404.

## Deliberately not in this PR

- **`SubIssuesEndpoint.post`** — the advisory's third vector. Already fixed by open PR #9466 (WEB-8352); no point duplicating it.
- **Five adjacent endpoints with the same bug class** — `reaction.py`, `link.py`, `subscriber.py`, `CommentReactionViewSet.create`, `attachment.py`. All verified reachable cross-workspace, but impact is materially lower: their responses carry no victim content and rows land with the *attacker's* `project_id`, so the victim's project-scoped querysets and `notification_task` hide them (pollution, not disclosure). Tracked in SECUR-244.

## Behaviour change worth reviewing

`comment.py` uses `Issue.objects` and `relation.py` uses `Issue.issue_objects`. That asymmetry is intentional, not an oversight — `IssueManager` additionally excludes triage-state, archived, draft, and archived-project issues:

- **Comments stay on `Issue.objects`** — exactly behaviour-preserving. Intake/triage, archived and draft issues must remain commentable (intake depends on it).
- **Relations use `issue_objects`** — matches the file's existing convention (the body `issues` filter) and `SubIssuesEndpoint`. This **narrows** behaviour: creating or removing a relation whose *URL* issue is triage/archived/draft now 404s where it previously succeeded. No first-party regression — the web client renders no relations UI in intake, passes `disabled={isArchived}` on archived detail, and drafts have no relation UI — but a non-web app-API consumer doing this would see the change. Flagging explicitly in case reviewers prefer `Issue.objects` here for strict parity.

New tests pin this boundary in both directions, so a future "tidy these to match" change can't silently break commenting on intake items.

## Testing

12 new contract tests in `plane/tests/contract/app/test_issue_subresource_scope_app.py`, covering each handler with a cross-workspace case, a same-workspace-other-project case, and a positive control.

**Fail-before verified** by reverting each fix and re-running:

- comment create → **201** with `"Confidential victim issue"` and `"Secret roadmap detail"` present in the response body
- relation create → **201** with an `IssueRelation` row written against the victim issue
- relation removal → **204** with `deleted_at` set on the sibling project's relation
- missing relation → **500** `AttributeError: 'NoneType' object has no attribute 'delete'`

Positive controls stayed green throughout, so the tests exercise the scoping rather than the plumbing — and the 404s are attributable to the new lookups, not to a permission denial that would have happened anyway (the URL project is always one the attacker *is* a member of).

Full suite: **526 passed** (`plane/tests/contract` + `plane/tests/unit`). `ruff check` and `ruff format --check` clean.

## Follow-ups

- **EE parity** — plane-ee vendors its own copy of the API and is not auto-synced; these three handlers are likely vulnerable there too. Belongs to Epic WEB-8293.
- **Systemic** — this bug class recurs while `ProjectEntityPermission` ignores `issue_id`. Worth validating issue-to-project binding at the permission layer rather than per view. Also noted in SECUR-244: `permissions/base.py`'s `creator=True` branch filters on `created_by` with no workspace/project scope, which is safe only because today's callers re-scope themselves.

The advisory stays in triage until this ships in a release — it is not eligible for the v1.4.0 CVE batch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved issue comment and relation access checks to limit requests to the specified workspace and project.
  * Cross-project or cross-workspace requests now return a 404 without modifying data.
  * Missing issue relations now return a 404 instead of an error.
  * Relation operations on archived issues are prevented, while supported comments remain available.

* **Tests**
  * Added coverage for valid and invalid issue sub-resource operations, including boundary and data-protection scenarios.

### DIFF
--- a/apps/api/plane/app/views/issue/comment.py
+++ b/apps/api/plane/app/views/issue/comment.py
@@ -67,7 +67,7 @@ class IssueCommentViewSet(BaseViewSet):
         # checks that the caller is a member of `project_id`, never that `issue_id` lives in
         # it, so a bare pk lookup let any project member comment on an issue in another
         # project/workspace — and read that issue's full `issue_detail` back out of the 201
-        # response (GHSA-hvx3-58mp-5fpx).
+        # response.
         issue = Issue.objects.filter(pk=issue_id, workspace__slug=slug, project_id=project_id).first()
         if issue is None:
             return Response({"error": "Issue not found"}, status=status.HTTP_404_NOT_FOUND)
--- a/apps/api/plane/app/views/issue/comment.py
+++ b/apps/api/plane/app/views/issue/comment.py
@@ -63,7 +63,14 @@ class IssueCommentViewSet(BaseViewSet):
     @allow_permission([ROLE.ADMIN, ROLE.MEMBER, ROLE.GUEST])
     def create(self, request, slug, project_id, issue_id):
         project = Project.objects.get(pk=project_id)
-        issue = Issue.objects.get(pk=issue_id)
+        # SECURITY: bind the issue to the URL workspace + project. `allow_permission` only
+        # checks that the caller is a member of `project_id`, never that `issue_id` lives in
+        # it, so a bare pk lookup let any project member comment on an issue in another
+        # project/workspace — and read that issue's full `issue_detail` back out of the 201
+        # response (GHSA-hvx3-58mp-5fpx).
+        issue = Issue.objects.filter(pk=issue_id, workspace__slug=slug, project_id=project_id).first()
+        if issue is None:
+            return Response({"error": "Issue not found"}, status=status.HTTP_404_NOT_FOUND)
         if (
             ProjectMember.objects.filter(
                 workspace__slug=slug,
--- a/apps/api/plane/app/views/issue/relation.py
+++ b/apps/api/plane/app/views/issue/relation.py
@@ -34,15 +34,23 @@ from plane.utils.issue_relation_mapper import get_actual_relation
 from plane.utils.host import base_host
 
 
-def issue_in_project(issue_id, slug, project_id):
+def issue_in_project(issue_id, slug, project_id, *, include_archived=False):
     """Whether ``issue_id`` belongs to the URL workspace + project.
 
     ``ProjectEntityPermission`` only checks that the caller is a member of the URL's
     ``project_id`` — it never binds the sibling ``issue_id`` path parameter to that
     project. Every handler taking both must therefore check this itself, or it is
     reachable cross-project/cross-tenant.
+
+    ``Issue.issue_objects`` (the default) additionally excludes archived/draft/triage
+    issues, which is correct for the write paths below — but a read path like
+    ``list()`` must still find archived issues in the URL's own project, the same way
+    ``IssueViewSet.retrieve()`` does with ``Issue.objects``. Pass
+    ``include_archived=True`` for that case; the project/workspace binding itself is
+    unchanged either way.
     """
-    return Issue.issue_objects.filter(pk=issue_id, workspace__slug=slug, project_id=project_id).exists()
+    manager = Issue.objects if include_archived else Issue.issue_objects
+    return manager.filter(pk=issue_id, workspace__slug=slug, project_id=project_id).exists()
 
 
 class IssueRelationViewSet(BaseViewSet):
@@ -55,7 +63,12 @@ class IssueRelationViewSet(BaseViewSet):
         # it a member of project A could list the relations of an issue in project B of
         # the same workspace and receive that issue's name, priority, assignees and
         # labels in the response.
-        if not issue_in_project(issue_id, slug, project_id):
+        #
+        # include_archived=True: this is a read, not a write, and the issue detail page
+        # fetches relations unconditionally for archived issues too. Scoping this lookup
+        # with the write-path's issue_objects (which excludes archived issues) would 404
+        # a legitimate same-project read instead of only closing the cross-project hole.
+        if not issue_in_project(issue_id, slug, project_id, include_archived=True):
             return Response({"error": "Issue not found"}, status=status.HTTP_404_NOT_FOUND)
 
         issue_relations = (
--- a/apps/api/plane/app/views/issue/relation.py
+++ b/apps/api/plane/app/views/issue/relation.py
@@ -217,6 +217,15 @@ class IssueRelationViewSet(BaseViewSet):
         issues = request.data.get("issues", [])
         project = Project.objects.get(pk=project_id)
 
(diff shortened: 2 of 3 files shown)


mguptahub:
Thanks — one of these was a real hole I'd missed, one doesn't reproduce. Addressed in 9fd1d53.

### 1. `list()` unscoped — valid, fixed ✅

Correct, and the more serious of the two. I reproduced it before fixing: a member of project A calling `GET .../issues/<project-B-issue-id>/issue-relation/` got **200** with the foreign issue's `name`, `priority`, `sequence_id`, `assignee_ids` and `label_ids` in the body. Same root cause as the write handlers, and I'd only looked at writes.

`list()` now applies the same binding and 404s. Two new contract tests cover it (cross-project 404 with no name leak, plus an in-project positive control), and I verified both fail without the fix — the failure output literally contains `'name': 'Sibling relation target'`.

### 2. Non-UUID `related_issue` → 500 — doesn't reproduce ❌

I probed this directly rather than reasoning about it:

```
POST .../remove-relation/  {"related_issue": "not-a-uuid"}
-> 400 {'error': 'Please provide valid detail'}
```

Django raises `django.core.exceptions.ValidationError` on the UUID coercion, and `BaseViewSet.handle_exception` (`app/views/base.py:85-89`) already maps that to a 400 — `base.py` imports Django's `ValidationError`, not DRF's, so the branch does catch it. Same applies to the `pk__in=issues` filter in `create`.

No code change, but I added a test pinning the 400 so a future change to that handler can't quietly turn malformed input into a 500. Missing key is also covered: `{}` → 404 via the new relation guard, not a crash.

### 3. `remove_relation` boundary test — valid, added ✅

Fair point, and it applied to `list()` too once that was fixed. The `IssueManager` boundary is now pinned on all three relation handlers (create, remove, list) plus both comment cases, so the `Issue.objects` vs `Issue.issue_objects` asymmetry can't drift in either direction:

- comments on archived/draft issues → **201** (must stay commentable; intake depends on it)
- relation create/remove/list with an archived URL issue → **404** (the `issue_objects` narrowing, deliberate)

### Also, per CodeRabbit

Extracted the binding into a module-level `issue_in_project(issue_id, slug, project_id)` helper now that three handlers in this file need it — one place to get right, and the next handler added here has an obvious thing to call.

---

**Suite:** 17 contract tests in this file, **531 passed** overall (`plane/tests/contract` + `plane/tests/unit`), ruff clean on all changed files.

The two remaining `ruff` F401s in `sub_issue.py` (`Func`, `Q` unused) are pre-existing on `preview` and in a file this PR doesn't touch — #9466 is the one editing it, so they belong there rather than here.

Worth noting the advisory itself listed three vectors; this PR now closes four, because `list` and `remove_relation` were found by reading around the reported endpoints rather than trusting the report's inventory. Adjacent endpoints with the same bug class (`reaction.py`, `link.py`, `subscriber.py`, `CommentReactionViewSet.create`, `attachment.py`) are tracked separately in SECUR-244 — lower impact, since those responses carry no victim content and rows land with the attacker's `project_id`.

mguptahub:
Follow-up fix for a regression found in review of the relation `list()` scoping guard.

**The regression:** the project-scope check added to `IssueRelationViewSet.list()` looked up the parent issue via `Issue.issue_objects`, which excludes archived issues. That made listing relations on an archived issue 404 even for a member of its own project. This is a real read-path break, not just closing the hole: the issue detail page fetches relations unconditionally on load (including for archived issues) and renders the relations panel read-only rather than hiding it there.

**The fix:** `issue_in_project()` now takes an `include_archived` flag. `list()` passes `include_archived=True` and looks the issue up via `Issue.objects` (the same archive-inclusive manager `IssueViewSet.retrieve()` already uses for reads), while `create()`/`remove_relation()` keep the narrower `issue_objects` manager unchanged. The actual project/workspace scoping check — the fix for the cross-project IDOR — is identical in all three; only the archive-inclusivity of the parent-issue lookup changed for `list()`.

**Tests** (`test_issue_subresource_scope_app.py::TestIssueManagerBoundary`):
- `test_relation_list_allows_archived_issue_in_own_project` — archived issue in the URL's own project now returns 200 (previously incorrectly 404'd)
- `test_relation_list_rejects_archived_issue_in_other_project` — archived issue in a sibling project still 404s, confirming the archive-inclusive lookup didn't reopen the cross-project hole

All 18 tests in the file pass, including the pre-existing write-path archived-issue tests (unchanged behavior) and the create/remove-relation cross-project tests.

PR #9532, opened 2026-08-03
[GIT-229] fix: bucket cycle/module analytics chart by Issue.created_at

### LINKED ISSUE #9177: Analytics Charts for Cycles and Modules Show Wrong Data — Queryset Anchored to Wrong Model

### Current behavior

When I view analytics charts filtered by a cycle or module, the daily stats chart shows incorrect data. The queryset is reassigned from Issue objects to a flat .values_list("issue_id", flat=True) queryset anchored to CycleIssue/ModuleIssue. Subsequent .values("created_at__date") calls then use the join table's created_at instead of Issue.created_at, producing wrong date groupings in the chart.

Code pattern:
cycle_issues = CycleIssue.objects.filter(...).values_list("issue_id", flat=True)
queryset = cycle_issues  # queryset is now CycleIssue IDs, not Issue objects
queryset.values("created_at__date")  # uses CycleIssue.created_at, not Issue.created_at

What I expect is that analytics charts for cycles and modules correctly reflect issue creation dates.

apps/api/plane/app/views/analytic/project_analytics.py, lines 202 and 214

### Steps to reproduce

1. Open Analytics for a project
2. 2. Filter the chart by a specific cycle or module
3. 3. Observe the daily chart data doesn't match the actual issue creation dates

### Environment

Production

### Browser

Google Chrome

### Variant

Cloud

### Version

v0.17.0-dev

blockgroot:
### Description
`work_item_completion_chart()` (`apps/api/plane/app/views/analytic/project_analytics.py`) reassigned its queryset to a bare `CycleIssue`/`ModuleIssue` id projection when scoped to a cycle or module, instead of an `Issue` queryset. The subsequent `.values("created_at__date")` grouping then resolved against the **join table's** `created_at` (when the issue was added to the cycle/module) rather than the issue's own creation date — so the "work items created" line on every cycle/module analytics chart was bucketed under the wrong date.

The `completed_count` annotation's `issue__state__group` traversal was itself a symptom — it only made sense because the queryset was still `CycleIssue`-shaped.

Fix: keep the queryset as `Issue.issue_objects.filter(id__in=cycle_issues/module_issues)`, matching the pattern already used a few lines above in `get_work_items_stats()` in the same file, and correct the `completed_count` filter from `issue__state__group` to `state__group` to match.

Reproduced and verified against the real Postgres-backed test stack (`docker-compose-test.yml`) — full details and before/after test output posted on the linked issue.

Closes #9177 (GIT-229)

### Type of Change
- [ ] Code refactoring
- [ ] Documentation update

This is a backend data-bucketing bug (dates on a chart resolve against the wrong column) rather than a visual glitch, so a screenshot wouldn't show the discrepancy — see the automated reproduction in the Test Scenarios section and on the linked issue instead.

### Test Scenarios
Added `apps/api/plane/tests/unit/views/test_project_analytics_chart.py`:
- Builds a real `Workspace -> Project -> Cycle -> Issue -> CycleIssue` graph with an issue created 10 days ago but added to its cycle 2 days ago.
- Calls `work_item_completion_chart()` directly and asserts the "created" count lands under the issue's actual creation date, not the date it was added to the cycle.
- Before the fix: fails with `assert 0 == 1` (work item missing from its true creation-date bucket).
- After the fix: passes.
- Full unit suite (`pytest -m unit`) re-run after the fix: 343 passed, 0 failed — no regressions.

### References
- Closes #9177
- Plane work item: GIT-229

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed completion charts filtered by cycles or modules to accurately aggregate issue states and completion counts.
  * Corrected chart date bucketing so items are grouped by when the issue was created, rather than when it was added to a cycle.

### DIFF
--- a/apps/api/plane/app/views/analytic/project_analytics.py
+++ b/apps/api/plane/app/views/analytic/project_analytics.py
@@ -199,7 +199,7 @@ class ProjectAdvanceAnalyticsChartEndpoint(ProjectAdvanceAnalyticsBaseView):
                 end_date = cycle.end_date.date()
             else:
                 return {"data": [], "schema": {}}
-            queryset = Issue.issue_objects.filter(id__in=cycle_issues)
+            queryset = queryset.filter(id__in=cycle_issues)
 
         elif module_id is not None:
             module_issues = ModuleIssue.objects.filter(**self.filters["base_filters"], module_id=module_id).values_list(
@@ -211,7 +211,7 @@ class ProjectAdvanceAnalyticsChartEndpoint(ProjectAdvanceAnalyticsBaseView):
                 end_date = module.target_date
             else:
                 return {"data": [], "schema": {}}
-            queryset = Issue.issue_objects.filter(id__in=module_issues)
+            queryset = queryset.filter(id__in=module_issues)
 
         else:
             project = Project.objects.filter(id=project_id).first()
--- a/apps/api/plane/app/views/analytic/project_analytics.py
+++ b/apps/api/plane/app/views/analytic/project_analytics.py
@@ -199,7 +199,7 @@ class ProjectAdvanceAnalyticsChartEndpoint(ProjectAdvanceAnalyticsBaseView):
                 end_date = cycle.end_date.date()
             else:
                 return {"data": [], "schema": {}}
-            queryset = cycle_issues
+            queryset = Issue.issue_objects.filter(id__in=cycle_issues)
 
         elif module_id is not None:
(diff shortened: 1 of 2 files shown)


blockgroot:
Addressed in 398b55e — filtering the existing project-scoped `queryset` (with `select_related`/`prefetch_related` intact) by `id__in=cycle_issues`/`module_issues` instead of rebuilding a fresh `Issue.issue_objects` queryset. Re-ran the full unit suite after the change: 343 passed, 0 regressions.

PR #9536, opened 2026-08-03
fix: handle desktop OAuth handoff fallback

### LINKED ISSUE #9535: 🐛 Bug: macOS Google OAuth callback returns to sign-in

- [x] I searched the existing issues.

### Current behavior

On Plane for macOS, Google OAuth completes successfully in the browser and macOS returns control to Plane, but the desktop app remains on the sign-in screen.

Browser sign-in for the same account works normally.

### Steps to reproduce

1. Install Plane for macOS v3.0.0.
2. Choose Plane Cloud.
3. Select **Continue with Google**.
4. Complete Google sign-in in the browser.
5. Allow the browser to return to Plane.
6. Observe that Plane returns to the sign-in screen instead of completing the session.

### Root cause

The OAuth flow now returns a versioned desktop handoff envelope:

```
plane://open?v=1&o=<base64url-origin>&p=<base64url-path>
```

The v3.0.0 desktop client receives the deep link, but interprets the scheme host (`open`) as a literal web path. It therefore navigates the current instance to `/open` instead of decoding the origin and enclosed `/d/auth/?token=...` path.

The web auth guard then redirects the unauthenticated `/open` request back to sign-in. Local macOS unified logs confirm that the `GURL` AppleEvent reaches Plane, and Plane's persisted tab state ends at `/?next_path=/open/`, so this is not a Google OAuth or macOS protocol-registration failure.

### Expected behavior

A valid version 1 handoff for the current Plane instance should continue to its enclosed same-origin path and complete desktop authentication. Malformed, unsupported, cross-instance, and cross-origin handoffs should be rejected.

### Environment

- Plane desktop: v3.0.0
- Build: `260714bxbai235k`
- Bundle ID: `com.todesktop.2405288gp0ar3qc`
- Platform: macOS
- Variant: Plane Cloud

### Proposed compatibility fix

Add a public web fallback for `/open` that:

1. accepts only handoff version 1;
2. base64url-decodes `o` and `p`;
3. requires the embedded origin and destination origin to match the current instance; and
4. redirects invalid envelopes to sign-in.

A focused regression patch with origin-safety tests is ready. The desktop client should also decode the envelope directly in a future release.

kankunnawat:
### Description

Fixes #9535.

Plane Cloud returns successful desktop OAuth sessions through a versioned deep link:

`plane://open?v=1&o=<base64url-origin>&p=<base64url-path>`

Plane Desktop v3 currently treats the deep-link host (`open`) as a web path and navigates to `/open?... ` instead of decoding the handoff. Because the web app had no compatibility route for that path, authentication middleware sent the user back to sign-in even though Google OAuth had completed successfully.

This change adds a narrow `/open` fallback that:

- decodes version 1 origin and path parameters;
- accepts the handoff only when its decoded origin matches the current Plane instance;
- accepts only same-origin absolute paths;
- redirects invalid or unsupported payloads to the app root.

The desktop protocol handler can still be corrected independently; this fallback restores login for already-released clients and self-hosted instances without weakening the origin boundary.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Test plan

- `./node_modules/.bin/turbo run test --filter=web` — 11/11 tasks; 4/4 handoff cases
- `./node_modules/.bin/turbo run check:types --filter=web` — 11/11 tasks
- `./node_modules/.bin/turbo run build --filter=web` — 11/11 tasks
- Changed files pass `oxfmt --check`
- Changed source and test files pass `oxlint --deny-warnings`
- Browser verification against the production build:
  - valid same-origin handoff reached `/d/auth/?token=one-time-token`;
  - different-instance, cross-origin, and malformed handoffs were rejected to `/`.

### Test scenarios

- Valid v1 handoff for the current instance redirects to the decoded OAuth completion path.
- Handoff for another Plane instance is rejected.
- Cross-origin path payload is rejected.
- Unsupported versions and malformed base64url payloads are rejected.

## Summary by CodeRabbit

* **New Features**
  * Added support for desktop deep-link handoffs through a dedicated route.
  * Valid handoffs now redirect to the requested in-app destination.
  * Invalid, unsupported, malformed, or cross-origin handoffs safely fall back to the home page.

* **Tests**
  * Added coverage for valid handoffs and rejected redirect scenarios.

### DIFF
--- a/apps/web/app/routes/core.ts
+++ b/apps/web/app/routes/core.ts
@@ -367,6 +367,9 @@ export const coreRoutes: RouteConfigEntry[] = [
   // ========================================================================
   // Legacy URL redirects for backward compatibility
 
+  // Desktop deep-link fallback: legacy clients navigate plane://open handoffs to /open
+  route("open", "routes/redirects/core/desktop-handoff.tsx"),
+
   // --------------------------------------------------------------------
   // REDIRECT ROUTES
   // --------------------------------------------------------------------
--- a/apps/web/app/routes/redirects/core/desktop-handoff.tsx
+++ b/apps/web/app/routes/redirects/core/desktop-handoff.tsx
@@ -1,0 +1,22 @@
+/**
+ * Copyright (c) 2023-present Plane Software, Inc. and contributors
+ * SPDX-License-Identifier: AGPL-3.0-only
+ * See the LICENSE file for details.
+ */
+
+import { redirect } from "react-router";
+// helpers
+import { resolveDesktopHandoffRedirect } from "@/helpers/desktop-handoff.helper";
+// types
+import type { Route } from "./+types/desktop-handoff";
+
+export const clientLoader = ({ request }: Route.ClientLoaderArgs) => {
+  const handoffUrl = new URL(request.url);
+  const destination = resolveDesktopHandoffRedirect(handoffUrl.searchParams, handoffUrl.origin);
+
+  throw redirect(destination ?? "/");
+};
+
+export default function DesktopHandoff() {
+  return null;
+}
--- a/apps/web/helpers/desktop-handoff.helper.spec.mjs
+++ b/apps/web/helpers/desktop-handoff.helper.spec.mjs
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) 2023-present Plane Software, Inc. and contributors
+ * SPDX-License-Identifier: AGPL-3.0-only
+ * See the LICENSE file for details.
+ */
+
+import assert from "node:assert/strict";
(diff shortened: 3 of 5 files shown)


prateekshourya29:
Hey @kankunnawat, thanks for this, and for the tests and browser verification.

We've already got a fix for this on our side, just done a different way. Rather than adding a fallback for `/open`, we changed the OAuth flow to send the old deep link again, so 3.0.0 gets a format it already understands and `/open` never gets hit. Newer desktop builds accept both.

So this one isn't needed anymore. Closing it, but really appreciate you digging into it.

PR #9539, opened 2026-08-04
fix: expand=state returns bare UUID or crashes for issues with null or soft-deleted state

### LINKED ISSUE #9534: Work Items API returns bare `state` UUID despite `?expand=state`
Fetching a single work item with `GET /workspaces/{slug}/projects/{project_id}/work-items/{id}/?expand=state` **usually** returns the expanded `state` object, but for some work items it returns `state` as a bare UUID string instead — even with `expand=state` set. This makes it impossible to resolve the state name from the response alone; a separate `/states/` lookup is required only for those items.

### Expected
`state` is consistently expanded to an object when `expand=state` is passed:

```json
{
  "id": "11111111-1111-1111-1111-111111111111",
  "name": "PR Created",
  "state": {
    "id": "22222222-2222-2222-2222-222222222222",
    "name": "PR Created",
    "group": "started"
  }
}
```

### Actual (intermittent)
Same request, different work item — `state` comes back as a bare UUID:

```json
{
  "id": "33333333-3333-3333-3333-333333333333",
  "name": "Some ticket",
  "state": "22222222-2222-2222-2222-222222222222"
}
```

Consumers that read `state.name` get nothing for these items and have to fall back to a `/states/` lookup to map the UUID → name.

### Details
- **Endpoint:** `GET /workspaces/{slug}/projects/{project_id}/work-items/{id}/?expand=state`
- **API:** REST API (api.plane.so, cloud)
- **Impact:** intermittent — most items expand correctly, a minority return the bare UUID.

ahmadmustafa02:
Closes #9534

## Root cause
`IssueDetailAPIEndpoint.get()` and `WorkspaceIssueAPIEndpoint.get()` build one-off querysets that skip `.select_related("state")` (unlike their sibling `get_queryset()` methods). When `expand=state` is requested, `BaseSerializer.to_representation()` resolves the related object via `getattr(instance, "state")`, which goes through Django's forward-FK descriptor and `StateManager` — a `SoftDeletionManager` subclass filtering `deleted_at__isnull=True`.

This means:
- For issues whose state has been soft-deleted, this lookup raises `State.DoesNotExist`, surfaced as a 404/500-style error response instead of a usable payload.
- More generally, any endpoint route that skips `select_related` for an expand-eligible relation is exposed to the same class of bug.

## Fix
1. Added `.select_related("state")` to the inline querysets in `IssueDetailAPIEndpoint.get()` and `WorkspaceIssueAPIEndpoint.get()`, matching the pattern already used in their `get_queryset()` methods. This performs a plain SQL `LEFT JOIN` that bypasses the soft-delete-filtering manager, resolving the state correctly even when it's been soft-deleted.
2. Added a defensive fallback in `BaseSerializer.to_representation()`: if the related object can't be resolved (e.g. `None`), it now falls back to the raw `{expand}_id` value instead of raising an unhandled exception. This guards against the same issue in any other expand path that might miss `select_related`.

## Testing
Added two regression tests in `plane/tests/contract/api/test_work_item_state_expand.py`:
- Issue with `state=None` → `expand=state` returns `200` with `"state": null`.
- Issue whose state is soft-deleted → `expand=state` returns `200` with the fully expanded state object (not a bare UUID, not a crash).

Verified manually against a local seeded DB and via the full contract test suite (`docker compose -f docker-compose-test.yml run --rm --build api-tests pytest`) — both new tests pass; no regressions introduced.

## Scope note
A related issue exists in `IssueListCreateAPIEndpoint`'s `external_id`/`external_source` lookup branch (uses `Issue.objects` instead of `Issue.issue_objects`, missing `select_related`), but it's a distinct bug (different manager, different filtering semantics) — filing separately rather than bundling here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved issue detail responses when expanding state information.
  * Correctly returns `null` for issues without a state.
  * Preserves expanded state details when an assigned state has been soft-deleted.
  * Improved reliability and performance when loading issue state data.

### DIFF
--- a/apps/api/plane/api/serializers/base.py
+++ b/apps/api/plane/api/serializers/base.py
@@ -106,11 +106,18 @@ class BaseSerializer(serializers.ModelSerializer):
                     }
                     # Check if field in expansion  then expand the field
                     if expand in expansion:
-                        if isinstance(response.get(expand), list):
-                            exp_serializer = expansion[expand](getattr(instance, expand), many=True)
+                        is_many = isinstance(response.get(expand), list)
+                        related_obj = getattr(instance, expand, None)
+                        if related_obj is not None:
+                            if is_many:
+                                exp_serializer = expansion[expand](related_obj, many=True)
+                            else:
+                                exp_serializer = expansion[expand](related_obj)
+                            response[expand] = exp_serializer.data
                         else:
-                            exp_serializer = expansion[expand](getattr(instance, expand))
-                        response[expand] = exp_serializer.data
+                            # Related object is null or could not be resolved
+                            # (e.g. soft-deleted) - fall back to the raw id
+                            response[expand] = getattr(instance, f"{expand}_id", None)
                     else:
                         # You might need to handle this case differently
                         response[expand] = getattr(instance, f"{expand}_id", None)
--- a/apps/api/plane/api/serializers/base.py
+++ b/apps/api/plane/api/serializers/base.py
@@ -2,6 +2,9 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 # See the LICENSE file for details.
 
+# Django imports
+from django.core.exceptions import ObjectDoesNotExist
+
 # Third party imports
 from rest_framework import serializers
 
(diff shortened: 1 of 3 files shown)


ahmadmustafa02:
Addressed both CodeRabbit review points in the latest commit:

1. Explicit `ObjectDoesNotExist` handling - added a try/except around the 
related object fetch instead of relying on `getattr`'s default only 
incidentally covering `RelatedObjectDoesNotExist` (which happens to 
subclass `AttributeError`). This makes the intent explicit and decouples 
correctness from that implementation detail.

2. Added test coverage for `WorkspaceIssueAPIEndpoint`'s identifier-based 
route (workspace_slug + project_identifier + sequence_id), mirroring the 
existing null-state and soft-deleted-state tests for the UUID-based route.

All 4 tests pass locally via the full contract suite. Ready for review.

PR #9540, opened 2026-08-04
[SECUR-245] fix(security): guard PDF image srcs and authenticate /convert-document/

mguptahub:
## What

Fixes the two issues reported under SECUR-245 in `apps/live`, plus the one API-side change needed to keep the existing caller working.

Work item: **SECUR-245**

## 1. SSRF via PDF image rendering

The `image` node renderer handed `node.attrs.src` straight to `@react-pdf/image`, which `fetch()`es any URL with a host and `fs.readFile()`s a bare path. The Live container shares a Docker network with `api`, `plane-db`, `plane-redis`, `plane-mq` and `plane-minio`, so page content could drive requests at internal-only services.

Adds `apps/live/src/lib/url-security.ts` and routes **both** `<Image>` call sites through it; anything unsafe renders a placeholder.

> **The existing `imageComponent` guard was not a usable model.** Its check is `!src.startsWith("http") && !src.startsWith("data:")` — and `http://api:8000/` and `http://plane-minio:9000/` both *pass* it. The scheme is irrelevant here; the destination is what has to be judged. That check is replaced rather than copied.

Rejected: non-`http(s)`/`data` schemes · bare and relative filesystem paths · loopback · RFC1918 · CGNAT `100.64/10` · link-local incl. `169.254.169.254` · multicast/reserved/test ranges · IPv6 ULA, link-local, site-local, multicast, NAT64, 6to4, Teredo, IPv4-mapped · obfuscated encodings (`2130706433`, `0x7f000001`, `127.1`) · single-label hosts (the shape of a Compose service name) · `.local`/`.internal`/`.lan` suffixes · embedded credentials · control-character scheme smuggling.

The IPv6 ranges are deliberately kept in step with the Python guard's `_BLOCKED_NETWORKS` in `apps/api/plane/utils/ip_address.py`. The first draft here was missing Teredo `2001::/32` and `fec0::/10` — two implementations of one policy drifting apart is how this class of bug keeps coming back.

## 2. Unauthenticated `/convert-document/`

`requireSecretKey` existed but was applied to **no controller**, leaving an expensive HTML → Y.js conversion open to anyone who could reach the service (CWE-306). It is now applied.

Its only caller is the API's `copy_s3_object` duplication task, which sent `headers=None`. So this PR also:
- sends the shared secret from that task, and
- wires `LIVE_SERVER_SECRET_KEY` into Django settings — it was previously only in `.env.example`.

A missing key short-circuits with a logged misconfiguration instead of firing a request that can only 401.

## Two corrections to the advisory

**`/pdf-export/` is not unauthenticated** and is deliberately untouched. `parseRequest` rejects requests with no `Cookie` and forwards that cookie to the API to fetch the page, so the API enforces page permissions. Gating it on the shared secret would break the browser client that design implies.

**The pre-auth chain does not connect.** `/convert-document/` performs no outbound fetch, and its output is never fed to `/pdf-export/`, which reads content from the API by `pageId`. Triggering the SSRF requires a valid session, so this is **`PR:L`**, not the reported pre-auth. Suggest reconciling severity before publishing — GitHub metadata says CVSS 8.6 while the advisory body says 7.5.

## Known residual — please read before approving

**DNS rebinding on the `http(s)` path is NOT closed here.** An attacker-controlled hostname that resolves to a blocked address still gets fetched: the renderer is synchronous and the `fetch()` happens inside `@react-pdf/image`, so the resolved address cannot be pinned. This is documented in the helper's doc comment.

The proper fix is to pre-fetch raw `image` nodes into `data:` URIs exactly as `imageComponent` already pre-fetches assets (`pdf-export.service.ts:229-230`), then accept only `data:` at render time. Tracked as a follow-up to SECUR-245. Flagging it so this is not read as "SSRF fully fixed".

## Deployment note

Gating `/convert-document/` means duplication degrades if the API lacks the key: `if external_data:` prevents corruption, but the duplicate keeps a stale `description_binary`, so its images break.

Verified `LIVE_SERVER_SECRET_KEY` reaches `api`/`worker`/`beat-worker` via `x-app-env` in `deployments/cli/community/docker-compose.yml`, and that aio auto-generates it on first boot. **Not verified: the Helm charts** (separate repo) — worth confirming before release so self-hosted K8s users don't hit this.

## Testing

- **78 new Live tests** (`apps/live/tests/lib/url-security.test.ts`) covering every payload named in the advisory, plus range boundaries (e.g. `100.63.x` allowed / `100.64.x` blocked) so the CIDR edges are pinned.
- **6 new API tests** (`test_copy_s3_object_auth.py`) for the header contract and the missing-key / no-live-url paths. The pre-existing duplication test mocks `sync_with_external_service` entirely, so this path had no coverage.
- Full Live suite: **110 pass**. `tsc --noEmit`: **24 errors before and after** — zero introduced, all pre-existing unbuilt `@plane/logger`/`@plane/decorators`. oxlint 0 errors, oxfmt and ruff clean.

## Out of scope, worth follow-ups

- `requireSecretKey` compares with `!==` (`auth-middleware.ts:38`) — **not constant-time**. Pre-existing, but it now actually guards something, so `crypto.timingSafeEqual` is warranted. Its docstring also claims it accepts `x-admin-secret-key` "(preferred)" while the code only reads `live-server-secret-key`.
- `<Link src={href}>` (`node-renderers.tsx:65`) is unguarded. Not SSRF (no fetch), but it embeds arbitrary URIs including `javascript:` into PDF link annotations.

Co-authored-by: Plane AI <noreply@plane.so>

## Summary by CodeRabbit

* **Security**
  * Protected document conversion requests with server authentication.
  * Blocked unsafe, internal, malformed, or non-image URLs, including unsafe redirects, from PDF rendering.
  * Added clear placeholders for unavailable images.

* **Bug Fixes**
  * Improved conversion reliability with bounded timeouts and better error handling.
  * Added graceful handling for missing configuration and unsuccessful responses.
  * Improved rendering of externally hosted images in exported PDFs.

* **Tests**
  * Added comprehensive coverage for authentication, URL validation, redirects, image sources, timeouts, and error scenarios.

### DIFF
--- a/apps/api/plane/bgtasks/copy_s3_object.py
+++ b/apps/api/plane/bgtasks/copy_s3_object.py
@@ -77,7 +77,19 @@ def sync_with_external_service(entity_name, description_html):
 
         url = normalize_url_path(f"{live_url}/convert-document/")
 
-        response = requests.post(url, json=data, headers=None)
+        # The Live service authenticates this endpoint with a shared secret
+        # (GHSA-55gq-rf47-9pqx). Without the header the request is rejected as 401.
+        secret_key = settings.LIVE_SERVER_SECRET_KEY
+        if not secret_key:
+            log_exception(
+                Exception(
+                    "LIVE_SERVER_SECRET_KEY is not configured; skipping document conversion "
+                    "for duplication. Set it to the same value as the Live service."
+                )
+            )
+            return {}
+
+        response = requests.post(url, json=data, headers={"live-server-secret-key": secret_key})
         if response.status_code == 200:
             return response.json()
     except requests.RequestException as e:
--- a/apps/api/plane/bgtasks/copy_s3_object.py
+++ b/apps/api/plane/bgtasks/copy_s3_object.py
@@ -19,9 +19,8 @@ from celery import shared_task
 from plane.utils.url import normalize_url_path
 
 # (connect, read) timeout for the Live service call. `requests` has no default
-# timeout, so omitting this lets a duplication task occupy a Celery worker
-# indefinitely if Live accepts the connection and then stops responding. The read
-# budget is generous because converting a large document is genuinely slow.
+# timeout, so without one a Live service that stalls after accepting the connection
+# pins a Celery worker forever. The read budget is wide: conversion is genuinely slow.
 LIVE_REQUEST_TIMEOUT = (5, 30)
 
 
@@ -83,8 +82,8 @@ def sync_with_external_service(entity_name, description_html):
 
         url = normalize_url_path(f"{live_url}/convert-document/")
 
-        # The Live service authenticates this endpoint with a shared secret
-        # (GHSA-55gq-rf47-9pqx). Without the header the request is rejected as 401.
+        # The Live service authenticates this endpoint with a shared secret.
+        # Without the header the request is rejected as 401.
         secret_key = settings.LIVE_SERVER_SECRET_KEY
         if not secret_key:
             log_exception(
--- a/apps/api/plane/bgtasks/copy_s3_object.py
+++ b/apps/api/plane/bgtasks/copy_s3_object.py
@@ -18,6 +18,12 @@ from plane.settings.storage import S3Storage
 from celery import shared_task
 from plane.utils.url import normalize_url_path
 
+# (connect, read) timeout for the Live service call. `requests` has no default
+# timeout, so omitting this lets a duplication task occupy a Celery worker
+# indefinitely if Live accepts the connection and then stops responding. The read
+# budget is generous because converting a large document is genuinely slow.
+LIVE_REQUEST_TIMEOUT = (5, 30)
+
 
 def get_entity_id_field(entity_type, entity_id):
     entity_mapping = {
@@ -89,7 +95,12 @@ def sync_with_external_service(entity_name, description_html):
             )
             return {}
 
-        response = requests.post(url, json=data, headers={"live-server-secret-key": secret_key})
+        response = requests.post(
+            url,
+            json=data,
+            headers={"live-server-secret-key": secret_key},
+            timeout=LIVE_REQUEST_TIMEOUT,
+        )
         if response.status_code == 200:
             return response.json()
     except requests.RequestException as e:
--- a/apps/api/plane/settings/common.py
+++ b/apps/api/plane/settings/common.py
@@ -418,6 +418,10 @@ LIVE_BASE_PATH = os.environ.get("LIVE_BASE_PATH", "/live/")
 
 LIVE_URL = urljoin(LIVE_BASE_URL, LIVE_BASE_PATH) if LIVE_BASE_URL else None
 
+# Shared secret for server-to-server calls into the Live service. Must match the
+# Live container's LIVE_SERVER_SECRET_KEY.
+LIVE_SERVER_SECRET_KEY = os.environ.get("LIVE_SERVER_SECRET_KEY")
+
 # WEB URL
 WEB_URL = os.environ.get("WEB_URL")
 
--- a/apps/api/plane/tests/unit/bg_tasks/test_copy_s3_object_auth.py
+++ b/apps/api/plane/tests/unit/bg_tasks/test_copy_s3_object_auth.py
@@ -3,15 +3,11 @@
 # See the LICENSE file for details.
 
 """
-Authentication of the API -> Live `/convert-document/` call (GHSA-55gq-rf47-9pqx).
+Authentication of the API -> Live `/convert-document/` call.
 
-The Live service previously served `/convert-document/` to anyone who could reach
-it (`requireSecretKey` was defined but never applied to a controller). Now that the
-endpoint is gated on the `live-server-secret-key` header, this background task is
-the one caller that has to present it — so the header must actually be sent, and a
-missing key must fail loudly rather than firing a request that 401s.
-
-These are pure unit tests: no database, no network.
+Live now gates the endpoint on the `live-server-secret-key` header, and this task is
+its only caller: the header must actually be sent, and a missing key must fail loudly
+rather than firing a request that can only 401. Pure unit tests — no DB, no network.
 """
 
 from unittest.mock import MagicMock, patch
--- a/apps/api/plane/tests/unit/bg_tasks/test_copy_s3_object_auth.py
+++ b/apps/api/plane/tests/unit/bg_tasks/test_copy_s3_object_auth.py
@@ -16,9 +16,10 @@ These are pure unit tests: no database, no network.
 
 from unittest.mock import MagicMock, patch
 
+import requests
 from django.test import override_settings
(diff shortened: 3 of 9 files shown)


mguptahub:
Both review findings addressed in `7dd98832f5`. Thanks — one of them was a genuine bug in the new guard, and it was worse than reported.

## @copilot — over-blocking reserved ranges ✅ fixed, and it was 3 cases not 1

Correct, and the pattern held in two more places I'd written the same way. The blocks in question are `/24`s sitting inside otherwise-**public** `/16`s, but I was matching on the second octet only:

| Range | Intended | Actually blocked | Collateral |
|:---|:---|:---|:---|
| `192.0.0.0/24` + `192.0.2.0/24` | two /24s | **all of `192.0.0.0/16`** | `192.0.1.x`, `192.0.3.x` … |
| `198.51.100.0/24` | one /24 | **all of `198.51.0.0/16`** | `198.51.99.x`, `198.51.101.x` … |
| `203.0.113.0/24` | one /24 | **all of `203.0.0.0/16`** | `203.0.112.x`, `203.0.114.x` … |

Now matched on the third octet. Failure mode was silent — legitimate images would just stop rendering with no error.

**Verified against the Python guard** rather than by inspection. Ran `is_blocked_ip` from `apps/api/plane/utils/ip_address.py` over all 15 boundary cases; TS and Python verdicts now agree exactly:

```
192.0.0.1      blocked      192.0.1.1     allowed
192.0.2.1      blocked      192.0.3.1     allowed
198.51.100.1   blocked      198.51.99.1   allowed
203.0.113.1    blocked      203.0.114.1   allowed
100.64.0.1     blocked      100.63.0.1    allowed
172.31.255.254 blocked      172.32.0.1    allowed
```

Worth naming: `url-security.ts` carries a comment warning that the TS and Python lists must not drift — and it shipped with three drifts already in it. The comment was aspirational, not true. It is now actually verified.

## @copilot — missing test coverage for those ranges ✅ added

Added assertions in **both** directions, which is what matters here: the four reserved `/24`s must be rejected, **and** the six adjacent public `/24`s must still be allowed. The second group is what caught the bug — I added the tests before the fix and watched `still allows the public space surrounding those /24s` fail. A reject-only test would have passed against the broken code.

## @coderabbitai — unbounded `requests.post` ✅ fixed

Correct and a real availability issue: `requests` has no default timeout, so a Live server that accepts the connection then stalls would pin a Celery worker indefinitely. This pre-dated the PR (the line was `headers=None`), but I touched it, so it's in scope.

Added `LIVE_REQUEST_TIMEOUT = (5, 30)` — short connect, generous read since converting a large document is legitimately slow — and asserted the argument is passed, as suggested.

Also added `test_timeout_is_swallowed_not_raised`: `requests.Timeout` subclasses `RequestException`, so the existing handler already returns `{}` and duplication degrades rather than the task failing. Now pinned by a test instead of relying on the exception hierarchy holding.

## Gates

Live **112 pass** (80 in `url-security.test.ts`) · API **8 pass** · oxlint 0 errors · oxfmt and ruff clean. Re-ran everything after the pre-commit hook's auto-fixes.

The **DNS-rebinding residual on the `http(s)` path is still open** and still deliberate — see the original description. Neither review touched it, so flagging again that this PR does not close it.

mguptahub:
Follow-up commit addresses a redirect-follow gap in the image SSRF guard that a review found: `isSafeImageSrc` only validated the URL handed to `<Image src>` at render time. The actual fetch happens inside `@react-pdf/image`'s `fetchRemoteFile`, which does a plain `fetch()` with Node's default `redirect: "follow"` — so a URL on an otherwise-safe public host could respond `302 Location: http://169.254.169.254/...` (or any other blocked address) and the fetch would follow it transparently, since `isSafeImageSrc` was never re-consulted on the redirect target. No DNS control needed, just one crafted HTTP response from a host that passes the initial check.

**Fix**: extended the pre-fetch pipeline that already exists for `imageComponent` (Plane's own uploaded assets) to cover raw `image` nodes (external URLs pasted into content) too, so the render pass never touches a raw external URL:

- `apps/live/src/lib/url-security.ts`: added `fetchImageSrcSafely(uri)`. Fetches with `redirect: "manual"`, and on every 3xx response resolves the `Location` header against the current URL and re-validates it with the existing `isSafeImageSrc` before following it, capped at 5 redirects. Any failure (initial URL, a redirect target, or the hop cap) resolves to `null` — same outcome as `isSafeImageSrc` returning `false` today.
- `apps/live/src/services/pdf-export/pdf-export.service.ts`: added `extractExternalImageSrcs` (collects `image`-node URLs the existing `extractImageAssetIds` skips) and `processExternalImages` (resolves them through `fetchImageSrcSafely` into data URIs, merged into the same `resolvedImageUrls` map `imageComponent` assets already populate), wired into the same Effect pre-fetch step before rendering starts.
- `apps/live/src/lib/pdf/node-renderers.tsx`: the `image` renderer now looks up the pre-resolved value from `ctx.metadata.resolvedImageUrls`, mirroring how `imageComponent` already does it, instead of calling `isSafeImageSrc` against the raw URL and handing that straight to `<Image>`. Missing/failed resolution renders the existing `[Image unavailable]` placeholder.
- Also switched the pre-existing per-asset URL resolution loop in `processImages` to `Promise.all` (it was already independent per asset; needed to satisfy the same lint rule as the new, genuinely-sequential redirect-following code).

**Tests**: the existing 78+ `isSafeImageSrc` cases in `apps/live/tests/lib/url-security.test.ts` pass untouched. Added cases for `fetchImageSrcSafely` covering: a safe host redirecting to link-local/loopback/RFC1918/Docker-service-name addresses (refused, without a second fetch), a safe-to-safe redirect (followed, body returned), relative `Location` resolution, a redirect chain exceeding the cap (refused, bounded fetch count), no-`Location`/unparseable-`Location`/non-ok/network-error cases, and a direct 200. Added renderer-level cases in `apps/live/tests/lib/pdf/pdf-rendering.test.ts` covering: pre-resolved src renders, unresolved src renders the placeholder (never fetches the raw URL), a failed-validation src renders the placeholder, and inline `data:` URIs still render without needing a pre-fetch entry. Full suite: 131/131 passing. Lint and typecheck clean.

**Residual DNS-rebinding risk**: not fully closed by this fix, and I want to be explicit about that rather than overclaim. `isSafeImageSrc` validates a hostname once; the actual fetch resolves DNS independently, on its own schedule. Re-validating every redirect hop closes the easy exploit path (a single crafted 3xx response), but on the final, non-redirect response, there's still a window between validation and the underlying DNS resolution performed by `fetch()` itself — a name that resolves to a public IP at validation time and a blocked IP by the time the connection is actually opened (classic DNS rebinding) is not pinned by this change. That gap requires an attacker to control DNS for a domain and win a timing race, which is meaningfully harder to pull off than the redirect bypass this commit closes, and is documented inline in `isSafeImageSrc`'s docstring as a known, still-open limitation.

PR #9560, opened 2026-08-06
feat(web): allow attaching multiple files at once to a work item

### LINKED ISSUE #9544: [feature]: allow attaching multiple files at once to a work item

### Summary

Work item attachments can only be added one file at a time. In `apps/web/core/components/issues/attachment/attachment-upload.tsx` the dropzone is configured with `multiple: false`, and `onDrop` takes only the first entry:

```tsx
const onDrop = useCallback((acceptedFiles: File[]) => {
  const currentFile: File = acceptedFiles[0];
  ...
});

const { getRootProps, ... } = useDropzone({
  onDrop,
  maxSize: maxFileSize,
  multiple: false,
  disabled: isLoading || disabled,
});
```

Two consequences:

1. The file picker does not let you select more than one file.
2. If you drag and drop several files at once, only the first is uploaded and the rest are silently discarded — no error, no indication. That part reads more like a bug than a limitation.

Attaching a set of files is the normal case, not the exception: a handful of screenshots for a bug report, a log bundle, several exported documents. Right now each file is a separate drag-and-wait cycle, and dropping them together looks like it worked while quietly dropping all but one.

### Backend impact

None. `IssueAttachmentV2Endpoint.post` already creates exactly one `FileAsset` and one presigned POST per call, so the client can simply fan out N calls — no bulk endpoint is needed. (Note `ProjectBulkAssetEndpoint` is not applicable here: it binds already-uploaded description/cover assets to an entity and has no `ISSUE_ATTACHMENT` branch.) The per-file size limit still applies unchanged via `maxFileSize` / `FILE_SIZE_LIMIT`.

### Proposed change

In `attachment-upload.tsx`:

- set `multiple: true`
- iterate `acceptedFiles` with bounded concurrency instead of taking `[0]`
- show progress as `Uploading 2/5...` while the batch runs
- treat per-file failures as non-fatal: report which files were rejected (size / MIME) and keep the successful ones, rather than failing the whole batch

Happy to open the PR if maintainers are open to it.

### Version

- Self-hosted (Kubernetes), code reviewed against v1.4.0

Syed-Ali-Abbas-Zaidi:
Work item attachments were limited to one file at a time. All three attachment dropzones passed `multiple: false` to react-dropzone and read only `acceptedFiles[0]`, so dragging several files uploaded the first and silently discarded the rest with no error shown.

Upload every accepted file with bounded concurrency (3 in flight) so a large drop does not open one request per file. Per-file failures are non-fatal: the batch keeps its successes and names only the files that failed. Files rejected for exceeding the size limit are reported separately, also by name.

The single-file logic was duplicated across the upload card, the quick action button and the attachment list, so it now lives in one shared `useAttachmentDropHandler` hook and all three surfaces stay in sync.

No API change is needed: the existing endpoint already creates one asset and one presigned POST per call, so the client simply fans out N calls. The per-file size limit still applies unchanged.

i18n: add six attachment keys across all 19 locales and remove `only_one_file_allowed` and `file_size_limit`, which no longer apply.

Closes #9544

### Type of Change

- [x] Code refactoring
- [ ] Documentation update

## Summary by CodeRabbit

- **New Features**
  - Upload multiple attachments at once with progress tracking and up to three concurrent uploads.
  - Continue processing valid files when individual uploads fail, with success, partial-failure, and failure notifications.
  - Identify and report oversized files without blocking other uploads.
  - Refresh attachment views automatically after uploads complete.

- **Bug Fixes**
  - Improved upload-state handling and disabled interactions during uploads or when no workspace is selected.

- **Localization**
  - Updated attachment upload and failure messages across supported languages.


Syed-Ali-Abbas-Zaidi:
Closes https://github.com/tylerryu-oai/plane/issues/9

PR #9569, opened 2026-08-08
fix: order legacy filter date range by parsed datetime instead of string comparison

### LINKED ISSUE #9567: Legacy filter converter orders date ranges by string comparison, reversing non-ISO date inputs
### Summary

`LegacyToRichFiltersConverter._convert_date_value` builds a date range by comparing the raw date **strings** with `min()` / `max()`, which only matches chronological order for zero-padded ISO `YYYY-MM-DD` values. The field validator (`_validate_date`) accepts any `dateutil`-parseable format, so a non-ISO input produces a reversed `[start, end]` range.

`apps/api/plane/utils/filters/converters.py`:

```python
# _validate_date accepts any dateutil format
dateutil_parse(value)              # line ~178
...
# but the range is ordered by string comparison
start_date = min(after_dates[0], before_dates[0])   # line 309
end_date   = max(after_dates[0], before_dates[0])   # line 310
```

### Why it is wrong

`min`/`max` on strings sorts lexicographically. That equals chronological order only for ISO `YYYY-MM-DD`. Since `_validate_date` uses `dateutil_parse` and passes many other formats (`M/D/YYYY`, `D-M-YYYY`, `Jan 15 2023`, non-zero-padded months, etc.), any such value can be ordered wrongly.

### Example

Legacy filter for `target_date`:

```
["9/1/2023;after", "10/1/2023;before"]
```

Both pass `_validate_date`. Then:

- `min("9/1/2023", "10/1/2023")` -> `"10/1/2023"` (because `'1' < '9'`)
- `max(...)` -> `"9/1/2023"`

So the range becomes `["10/1/2023", "9/1/2023"]`, i.e. start = Oct 1, end = Sep 1, which is reversed and matches nothing. Chronologically the correct range is Sep 1 -> Oct 1.

### Scope

This only triggers when a non-ISO date string reaches this path. If every caller is guaranteed to send `YYYY-MM-DD`, it is latent, but the converter and its validator are internally inconsistent: the validator accepts formats the ordering logic then mishandles.

### Suggested fix

Order by parsed dates rather than by string, e.g. parse both bounds with `dateutil_parse` and pick min/max of the `datetime` values (or normalize to ISO before comparing), keeping the emitted values in whatever format the rich filter expects.

eeshsaxena:
Fixes #9567.

### Problem

`LegacyToRichFiltersConverter._convert_date_value` builds a date range with `min()` / `max()` on the raw date **strings**:

```python
start_date = min(after_dates[0], before_dates[0])
end_date   = max(after_dates[0], before_dates[0])
```

String comparison only matches chronological order for zero-padded ISO `YYYY-MM-DD`. But `_validate_date` uses `dateutil_parse`, so it accepts many other formats (`M/D/YYYY`, non-zero-padded months, etc.). A non-ISO input like `9/1/2023` vs `10/1/2023` sorts lexicographically (`"10..." < "9..."`) and produces a reversed `[start, end]` range that matches nothing.

### Fix

Order the two bounds by their parsed `datetime` values instead of by string, keeping the emitted values in their original form. This makes the range correct for every format `_validate_date` accepts, and is a no-op for already-ISO inputs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Date range filters now correctly order start and end bounds chronologically, including non-ISO date formats.

### DIFF
--- a/apps/api/plane/utils/filters/converters.py
+++ b/apps/api/plane/utils/filters/converters.py
@@ -306,8 +306,13 @@ class LegacyToRichFiltersConverter:
         result = {}
         if len(after_dates) == 1 and len(before_dates) == 1 and len(exact_dates) == 0:
             # Simple range: one after and one before
-            start_date = min(after_dates[0], before_dates[0])
-            end_date = max(after_dates[0], before_dates[0])
+            # Order the bounds chronologically by comparing parsed datetimes, not the
+            # raw strings: _validate_date accepts non-ISO formats that do not sort
+            # lexicographically (e.g. "9/1/2023" vs "10/1/2023").
+            if dateutil_parse(after_dates[0]) <= dateutil_parse(before_dates[0]):
+                start_date, end_date = after_dates[0], before_dates[0]
+            else:
+                start_date, end_date = before_dates[0], after_dates[0]
             self._add_rich_filter(result, field_name, "range", [start_date, end_date])
         elif len(exact_dates) == 1 and len(after_dates) == 0 and len(before_dates) == 0:
             # Single exact date


eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #9603, opened 2026-08-13
[SECUR-248] fix(security): guard ProjectViewSet PUT — routed verb falls through to DRF

mguptahub:
## Summary

`app/urls/project.py:40` maps `"put": "update"`, but `ProjectViewSet` never defined `update` — so PUT fell through to DRF's `ModelViewSet.update`. The class sets **no `permission_classes`**, so that generic handler ran under the project default of `IsAuthenticated`, while `partial_update` on the *identical* URL does an inline workspace-admin / project-admin check.

PATCH was authorised. PUT was not.

Verified against `preview` @ `1c8a60f858`. Closes 2 HIGH advisories (SECUR-248 carries the mapping).

## Impact — reproduced

A workspace MEMBER with no `ProjectMember` row for a `network=0` (Secret) project:

1. `PUT .../projects/<id>/` with `network: 2` → **`200 OK`**, project becomes Public
2. Join via the legitimate public-project API → `ProjectMember(role=15)`
3. Full read plus member-level write on a project they were never in

Step 1 confirmed in the fail-before run: an unpatched non-member PUT returns `200` and flips `network`.

## The change

```python
def update(self, request, slug, pk=None):
    return self.partial_update(request, slug=slug, pk=pk)
```

Both verbs now enforce the same rule. A full PUT-replace is not a meaningful operation for projects — the client only ever sends PATCH.

This follows the pattern #9461 established for the work-item / module / intake instances of the same defect. It does **not** need that PR's `partial_update.__wrapped__` trick: `ProjectViewSet.partial_update` carries no decorator (it authorises inline), so calling it directly does not double-authorise.

## Tests

Four contract tests:

- non-member PUT is denied
- a denied PUT **writes nothing** — a 403 that still flipped `network` would leave the project published, which is the actual damage
- **PUT and PATCH reach the same verdict** — the defect was precisely that they disagreed
- **positive control** — a workspace admin PUT still succeeds

**Fail-before verified:** 3 failed / 1 passed unpatched → 4 passed patched. The positive control passes in both runs, confirming it is independent of the change.

⚠️ **The payload must be complete.** DRF's generic `update` runs the serializer with `partial=False`, so an incomplete body 400s on validation *before* authorization differs at all — with a partial body the tests passed with and without the fix and proved nothing. `full_put_body()` exists for that reason and the docstring says so.

## Related — a broader finding

The ticket asked for a sweep for other instances. Checking every `as_view({...})` mapping in `app/urls/` against whether the viewset defines that action, then filtering to classes with **no class-level `permission_classes`**:

| | |
| :--- | ---: |
| routed DRF-generic actions the viewset does not define | 68 |
| of those, on classes with no class-level `permission_classes` | 34 |
| already covered by open PR #9461 | 4 |
| `ProjectViewSet` PUT (this PR) | 1 |
| **unaddressed** | **29** |

Examples, all on classes whose *defined* actions carry `@allow_permission`/`@can` while the undefined one gets nothing: `CycleViewSet` PUT, `IssueViewViewSet` POST, `WorkspaceViewViewSet` POST, `NotificationViewSet` DELETE, `StateViewSet` GET, `ModuleIssueViewSet` PUT/PATCH/GET.

I have **not** verified each is exploitable — some may be protected by `get_queryset` scoping, and `create`/`destroy` on a generic mixin may fail for unrelated reasons. That verification is the work. Tracked separately rather than expanding this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved project update authorization for PUT requests.
  - Unauthorized project updates are rejected without changing project details.
  - Workspace administrators can continue updating projects as expected.
  - PUT and PATCH requests now apply consistent authorization rules.

- **Tests**
  - Added coverage for unauthorized member access and protected project updates.

### DIFF
--- a/apps/api/plane/app/views/project/base.py
+++ b/apps/api/plane/app/views/project/base.py
@@ -311,6 +311,14 @@ class ProjectViewSet(BaseViewSet):
             return Response(serializer.data, status=status.HTTP_201_CREATED)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def update(self, request, slug, pk=None):
+        # urls/project.py routes PUT here, but this viewset never defined `update`,
+        # so it fell through to DRF's ModelViewSet.update under the default
+        # IsAuthenticated while partial_update on the same URL requires workspace
+        # or project admin. Route PUT through partial_update so both verbs enforce
+        # the same rule; a full PUT-replace is not meaningful for projects.
+        return self.partial_update(request, slug=slug, pk=pk)
+
     def partial_update(self, request, slug, pk=None):
         # try:
         is_workspace_admin = WorkspaceMember.objects.filter(
--- a/apps/api/plane/tests/contract/app/test_project_put_authz.py
+++ b/apps/api/plane/tests/contract/app/test_project_put_authz.py
@@ -1,0 +1,140 @@
+# Copyright (c) 2023-present Plane Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
+"""Contract tests for ``ProjectViewSet`` PUT authorization (SECUR-248).
+
+``urls/project.py`` maps ``"put": "update"``, but the viewset never defined
+``update`` — so PUT fell through to DRF's ``ModelViewSet.update``. The class sets
+no ``permission_classes``, so that generic handler ran under the project default
+of ``IsAuthenticated``, while ``partial_update`` on the *identical* URL requires
+workspace admin or project admin.
+
+A workspace member who is not in a Secret project could therefore PUT it to
+``network=2`` (Public) and then use the legitimate public-join API to add
+themselves — full read plus member-level write on a project they were never in.
+"""
+
+from uuid import uuid4
+
+import pytest
+from rest_framework import status
+from rest_framework.test import APIClient
+
+from plane.db.models import Project, ProjectMember, User, WorkspaceMember
+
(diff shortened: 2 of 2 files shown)


mguptahub:
Addressed the review blocker: the test module's docstring embedded the internal ticket reference directly in source. Reworded it to describe the invariant under test instead — no ticket ID anywhere in the file now (double-checked with a repo-wide grep restricted to files this branch touches).

Old: `"""Contract tests for \`\`ProjectViewSet\`\` PUT authorization (SECUR-248)."""`
New: `"""Contract tests for \`\`ProjectViewSet\`\` PUT authorization.` followed by a sentence stating the invariant (PUT must enforce the same authorization as PATCH, not fall through to DRF's generic handler under the bare default permission class), then the existing explanatory paragraphs unchanged.

Ran `ruff check` on the file (clean) and the full contract test module via `docker compose -f docker-compose-test.yml run --rm --build api-tests pytest plane/tests/contract/app/test_project_put_authz.py -m contract -v` — all 4 tests still pass, confirming the docstring change has no behavioral effect.

mguptahub:
Closing in favour of #9652, which supersedes this — it covers the same ProjectViewSet PUT fall-through plus every other undecorated routed verb across the URLconf (225 actions audited, 27 refused vs. this PR's single-viewset point-fix), from the same structural guard. No functionality here is lost; it's a strict subset of #9652's diff. 

Thanks for the fix — the audit approach is what's landing, not the specific patch.

PR #9605, opened 2026-08-13
fix(cycle): ignore_conflicts on add-issue bulk_create to avoid 500 on concurrent adds

### LINKED ISSUE #9598: cycle: concurrent add-issue requests cause IntegrityError - bulk_create missing ignore_conflicts unlike module equivalent
﻿## Bug Description

CycleIssueViewSet.create uses ulk_create without ignore_conflicts=True. The CycleIssue model has a partial unique constraint cycle_issue_when_deleted_at_null on (cycle, issue) where deleted_at IS NULL. Two concurrent POST requests adding the same issue to the same cycle both pass the existing-issue check (plain ilter, no row lock) and then both call ulk_create. The second request raises an IntegrityError and returns HTTP 500.

The inconsistency is clear: ModuleIssueViewSet.create_module_issues in module/issue.py (line 237) performs the identical operation with ignore_conflicts=True. The cycle equivalent never received the same fix.

## Affected file

pps/api/plane/app/views/cycle/issue.py, line 264:

`python
created_records = CycleIssue.objects.bulk_create(
    [CycleIssue(...) for issue in new_issues],
    batch_size=10,
    # missing ignore_conflicts=True
)
`

For comparison, pps/api/plane/app/views/module/issue.py line 237:
`python
CycleIssue.objects.bulk_create(records, batch_size=10, ignore_conflicts=True)
`

## Failure scenario

Two browser tabs open the same cycle simultaneously. Both click "Add issues" with overlapping issue selections. Both POST requests read no existing CycleIssue rows (race), both call ulk_create, the second throws IntegrityError, user sees a 500 error and must retry.

## Fix

`python
created_records = CycleIssue.objects.bulk_create(
    [CycleIssue(...) for issue in new_issues],
    batch_size=10,
    ignore_conflicts=True,
)
`

## Environment

Plane develop branch (2026-08-13).

eeshsaxena:
Fixes #9598.

Adding an issue to a cycle can return a 500 under concurrency. `CycleIssueViewSet.create` filters for issues already in the cycle, subtracts them, and bulk-inserts the rest:

```python
existing_issues = [str(ci.issue_id) for ci in cycle_issues]
new_issues = list(set(issues) - set(existing_issues))
...
created_records = CycleIssue.objects.bulk_create([...], batch_size=10)
```

That existing-issue check is a plain filter with no row lock, so two requests adding the same issue to the same cycle can both compute the same `new_issues` and both reach the insert. `CycleIssue` has a partial unique constraint `cycle_issue_when_deleted_at_null` on `(cycle, issue)` where `deleted_at IS NULL`, so the second insert raises `IntegrityError` and the user gets a 500.

The same operation elsewhere in the codebase already guards against this with `ignore_conflicts=True`:

- `module/issue.py` (`ModuleIssue` add-issue), both bulk_create calls
- `api/views/cycle.py` (the external API's cycle add-issue)

Only this app-API path was missed. This adds `ignore_conflicts=True` to match, so a racing duplicate insert is skipped instead of blowing up the request.

I checked the one place the return value is used: `created_records` is serialized into the cycle activity payload, and the consumer (`create_cycle_issue_activity` in `bgtasks/issue_activities_task.py`) only reads `fields.cycle` and `fields.issue` from it, never the primary key. So the fact that `ignore_conflicts=True` leaves the instance PKs unset does not affect anything downstream. Behavior in the normal (non-racing) case is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when adding issues to cycles concurrently.
  * Duplicate cycle-issue entries are now skipped instead of producing an error.

### DIFF
--- a/apps/api/plane/app/views/cycle/issue.py
+++ b/apps/api/plane/app/views/cycle/issue.py
@@ -282,6 +282,12 @@ class CycleIssueViewSet(BaseViewSet):
             ignore_conflicts=True,
         )
 
+        # ON CONFLICT DO NOTHING does not populate PKs on skipped rows, so a
+        # concurrent request that lost the race gets its duplicates back here
+        # with pk=None. Drop them before the activity log below, otherwise it
+        # records a "created" entry for an insert that was silently discarded.
+        created_records = [record for record in created_records if record.pk is not None]
+
         # Updated Issues
         updated_records = []
         update_cycle_issue_activity = []
--- a/apps/api/plane/app/views/cycle/issue.py
+++ b/apps/api/plane/app/views/cycle/issue.py
@@ -274,6 +274,12 @@ class CycleIssueViewSet(BaseViewSet):
                 for issue in new_issues
             ],
             batch_size=10,
+            # Concurrent add-issue requests can both pass the unlocked
+            # existing-issue check above and insert the same (cycle, issue),
+            # violating the cycle_issue_when_deleted_at_null unique constraint.
+            # Skip the duplicates instead of raising, matching the module and
+            # external-API equivalents.
+            ignore_conflicts=True,
         )
 
         # Updated Issues


harsh4vardhan:
The `ignore_conflicts=True` fix is correct and matches the module and external-API paths. One gap worth addressing: PostgreSQL's `ON CONFLICT DO NOTHING` does not populate PKs on skipped rows. Django still returns those objects in `created_records` with `id = None`, and they flow into `create_cycle_issue_activity`, producing an activity entry for an insert that was silently discarded. A second concurrent request that lost the race will log "issue added" even though it added nothing. Filtering before the activity call would close this:

```python
created_records = [r for r in created_records if r.pk is not None]
```

Without that, the fix trades a 500 for duplicate activity history under concurrent adds. Worth addressing in this PR or a fast follow.

eeshsaxena:
Good point to raise. In the rare concurrent case the losing request's `bulk_create` skips the duplicate but still returns it in `created_records`, so that issue would get a second "created" cycle activity entry. I left it that way on purpose: it is a duplicate log line rather than a duplicate row (the DB constraint still guarantees a single CycleIssue), and the activity consumer keys off `fields.issue`, so it is cosmetic. That trade is strictly better than the current behaviour, which is a 500 for the whole request.

Cleanly logging only the truly-inserted rows isn't straightforward with `ignore_conflicts=True`, since Postgres doesn't report back which rows were inserted vs skipped, and re-querying can't distinguish this request's inserts from the racing one's without its own race. The module add-issue path sidesteps this by discarding the result entirely. Happy to follow that pattern here if you'd prefer, but it felt out of scope for the crash fix.

eeshsaxena:
Good call, you're right that this shouldn't just be logged as a duplicate line. Under a concurrent add the losing request gets its skipped rows back with `pk=None`, and serializing those into `created_cycle_issues` records a "created" activity entry for an insert that was actually discarded, which is misleading history rather than harmless noise.

Pushed the filter: `created_records` now drops `pk is None` rows before `create_cycle_issue_activity`. Thanks for the catch.

eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`fix(cycle): ignore_conflicts on add-issue bulk_create to avoid 500 on concurrent adds`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining plane!

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #9670, opened 2026-08-22
fix(api): redirect instead of 500 on an invalid password reset link

### LINKED ISSUE #9172: Password Reset with Valid But Non-Existent User ID Returns Unhandled 500

### Current behavior

When I follow a password reset link that contains a valid base64-encoded but non-existent user ID, the server crashes with an unhandled DoesNotExist exception and I see the error message attached below:

User.DoesNotExist: User matching query does not exist.

What I expect is a redirect to the error page with a clear message that the reset link is invalid.

apps/api/plane/authentication/views/space/password_management.py — ResetPasswordSpaceEndpoint.post() calls User.objects.get(id=id) but only catches DjangoUnicodeDecodeError, leaving User.DoesNotExist unhandled

### Steps to reproduce

1. Construct a password reset URL with a valid base64-encoded UUID that doesn't correspond to any user
2. 2. Submit the reset form
3. 3. See 500 error instead of a redirect to the error page

### Environment

Production

### Browser

Google Chrome

### Variant

Cloud

### Version

v0.17.0-dev

TemoSulava:
### Description

`ResetPasswordSpaceEndpoint.post()` looked the user up inside a `try` block that only caught `DjangoUnicodeDecodeError`, so a reset link whose `uidb64` decodes to something unusable crashed with an unhandled 500 instead of redirecting to the invalid-link page:

| `uidb64` decodes to | raised | before (space) | before (app) | after (both) |
| --- | --- | --- | --- | --- |
| a UUID with no matching user | `User.DoesNotExist` | 500 | 302, `error_code=5125` | 302, `error_code=5125` |
| a non-UUID string | `django.core.exceptions.ValidationError` | 500 | 500 | 302, `error_code=5125` |
| undecodable base64 | `ValueError` (binascii) | 500 | 302, `error_code=5125` | 302, `error_code=5125` |
| invalid utf-8 | `DjangoUnicodeDecodeError` | 302, `error_code=5130` | 302, `error_code=5125` | 302, `error_code=5125` |

Both endpoints now catch `(ValueError, ValidationError, User.DoesNotExist)` around the decode and the lookup only, and answer identical input identically.

`DjangoUnicodeDecodeError` subclasses `UnicodeDecodeError` → `UnicodeError` → `ValueError`, so the single tuple clause covers the invalid-utf-8 case too. On `preview` the space endpoint answered it with `5130 EXPIRED_PASSWORD_TOKEN` while the app endpoint's identical handler sat *below* an `except (ValueError, ...)` on an inner `try` and was therefore dead code, answering `5125`. Per review, that handler is now deleted from both files rather than hoisted: an undecodable `uidb64` was never a valid link that later expired, so `5125 INVALID_PASSWORD_TOKEN` is the accurate answer, and it is the code the app endpoint already returned.

Behaviour delta versus `preview`, in full: the four 500s above become 302s, and the space endpoint's invalid-utf-8 response changes from `5130` to `5125`. `error_code=5130` is no longer emitted by the API; the frontend enum entries for it are left untouched. The rest of each method is dedented out of the outer `try` verbatim (`git diff -w` shows only the `except` clauses moving), and the catch is *narrower* than before: it now wraps only the decode and the lookup, not `set_password`/`save`.

### Type of Change

### Test Scenarios

New contract tests in `apps/api/plane/tests/contract/app/test_password_reset.py` — 20 cases covering every branch of both endpoints (unknown user, non-UUID id, malformed base64, undecodable utf-8, bad token, missing password, weak password, successful reset, token replay), plus two anchors that pin the raise site each rejected-uidb64 fixture reaches and the error-code literals that packages/constants/src/auth/index.ts hardcodes.

```
docker compose -f docker-compose-test.yml run --rm api-tests \
  pytest plane/tests/contract/app/test_password_reset.py
...
20 passed in 19.62s
```

Reverting only the two view files and rerunning gives `5 failed, 15 passed` — the three space-endpoint crash paths, the app endpoint's non-UUID crash, and the space endpoint's old 5130 response — so the tests pin the actual behaviour, not just the happy path.

`ruff check` and `ruff format --check` are clean on all three files.

### References

Fixes #9172

### DIFF
--- a/apps/api/plane/authentication/views/app/password_management.py
+++ b/apps/api/plane/authentication/views/app/password_management.py
@@ -18,7 +18,7 @@ from django.contrib.auth.tokens import PasswordResetTokenGenerator
 from django.core.exceptions import ValidationError
 from django.core.validators import validate_email
 from django.http import HttpResponseRedirect
-from django.utils.encoding import DjangoUnicodeDecodeError, smart_bytes, smart_str
+from django.utils.encoding import smart_bytes, smart_str
 from django.utils.http import urlsafe_base64_decode, urlsafe_base64_encode
 from django.views import View
 
@@ -107,18 +107,9 @@ class ResetPasswordEndpoint(View):
             # Decode the id from the uidb64
             id = smart_str(urlsafe_base64_decode(uidb64))
             user = User.objects.get(id=id)
-        except DjangoUnicodeDecodeError:
-            exc = AuthenticationException(
-                error_code=AUTHENTICATION_ERROR_CODES["EXPIRED_PASSWORD_TOKEN"],
-                error_message="EXPIRED_PASSWORD_TOKEN",
-            )
-            url = urljoin(
-                base_host(request=request, is_app=True),
-                "accounts/reset-password?" + urlencode(exc.get_error_dict()),
-            )
-            return HttpResponseRedirect(url)
         except (ValueError, ValidationError, User.DoesNotExist):
-            # Malformed base64, a non-UUID id or an id that matches no user
+            # Malformed base64, bytes that are not utf-8 (DjangoUnicodeDecodeError subclasses ValueError),
+            # a non-UUID id or an id that matches no user
             exc = AuthenticationException(
                 error_code=AUTHENTICATION_ERROR_CODES["INVALID_PASSWORD_TOKEN"],
                 error_message="INVALID_PASSWORD_TOKEN",
--- a/apps/api/plane/authentication/views/app/password_management.py
+++ b/apps/api/plane/authentication/views/app/password_management.py
@@ -100,70 +100,8 @@ class ResetPasswordEndpoint(View):
     def post(self, request, uidb64, token):
         try:
             # Decode the id from the uidb64
-            try:
-                id = smart_str(urlsafe_base64_decode(uidb64))
-                user = User.objects.get(id=id)
-            except (ValueError, User.DoesNotExist):
-                exc = AuthenticationException(
-                    error_code=AUTHENTICATION_ERROR_CODES["INVALID_PASSWORD_TOKEN"],
-                    error_message="INVALID_PASSWORD_TOKEN",
-                )
-                params = exc.get_error_dict()
-                url = urljoin(
-                    base_host(request=request, is_app=True),
-                    "accounts/reset-password?" + urlencode(params),
-                )
-                return HttpResponseRedirect(url)
-
-            # check if the token is valid for the user
-            if not PasswordResetTokenGenerator().check_token(user, token):
-                exc = AuthenticationException(
-                    error_code=AUTHENTICATION_ERROR_CODES["INVALID_PASSWORD_TOKEN"],
-                    error_message="INVALID_PASSWORD_TOKEN",
-                )
-                params = exc.get_error_dict()
-                url = urljoin(
-                    base_host(request=request, is_app=True),
-                    "accounts/reset-password?" + urlencode(params),
-                )
-                return HttpResponseRedirect(url)
-
-            password = request.POST.get("password", False)
-
-            if not password:
-                exc = AuthenticationException(
-                    error_code=AUTHENTICATION_ERROR_CODES["INVALID_PASSWORD"],
-                    error_message="INVALID_PASSWORD",
-                )
-                url = urljoin(
-                    base_host(request=request, is_app=True),
-                    "accounts/reset-password?" + urlencode(exc.get_error_dict()),
-                )
-                return HttpResponseRedirect(url)
-
-            # Check the password complexity
-            results = zxcvbn(password)
-            if results["score"] < 3:
-                exc = AuthenticationException(
-                    error_code=AUTHENTICATION_ERROR_CODES["PASSWORD_TOO_WEAK"],
-                    error_message="PASSWORD_TOO_WEAK",
-                )
-                url = urljoin(
-                    base_host(request=request, is_app=True),
-                    "accounts/reset-password?" + urlencode(exc.get_error_dict()),
-                )
-                return HttpResponseRedirect(url)
-
-            # set_password also hashes the password that the user will get
-            user.set_password(password)
-            user.is_password_autoset = False
-            user.save()
-
-            url = urljoin(
-                base_host(request=request, is_app=True),
-                "sign-in?" + urlencode({"success": True}),
-            )
-            return HttpResponseRedirect(url)
+            id = smart_str(urlsafe_base64_decode(uidb64))
+            user = User.objects.get(id=id)
         except DjangoUnicodeDecodeError:
             exc = AuthenticationException(
                 error_code=AUTHENTICATION_ERROR_CODES["EXPIRED_PASSWORD_TOKEN"],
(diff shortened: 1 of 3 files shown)


TemoSulava:
Thanks @coderabbitai — both points addressed in d0c09b6.

**Nitpick (app-route coverage):** valid. `TestResetPasswordAppEndpoint` only exercised uid handling and the success path, so `ResetPasswordEndpoint`'s invalid-token, missing-password and weak-password branches were untested. Added all three, plus an assertion that a rejected reset leaves `is_password_autoset` untouched and that the redirect lands on `accounts/reset-password` — the error code alone would not catch a branch redirecting to the wrong page. 16 passed.

**Docstring coverage check:** the five undocumented functions in the diff (`post()` on both endpoints, and the three test helpers) now have docstrings.

One correction worth recording: an earlier revision of the helper docstring claimed `reverse()` "resolves to the wrong view" for these routes. That is wrong — Django disambiguates duplicate url names by argument count, so `reverse("forgot-password", args=[uidb64, token])` does resolve to the reset-password pattern. Verified against a reconstruction of `plane/authentication/urls.py:122-138` on Django 5.2:

```
forgot-password        args=[]             -> /auth/change-password/
forgot-password        args=[UID, TOK]     -> /auth/reset-password/UID/TOK/
space-forgot-password  args=[]             -> /auth/spaces/forgot-password/
space-forgot-password  args=[UID, TOK]     -> /auth/spaces/reset-password/UID/TOK/
```

The tests still build the paths by hand — these are contract tests, so pinning the literal URL is the point — but the docstring now gives the accurate reason.

Separate from this PR, for maintainers: three patterns share `name="forgot-password"` and two share `name="space-forgot-password"` (`plane/authentication/urls.py:122-138`), so `reverse()` on those names is resolved purely by arity. Nothing in the repo reverses them today, so it is latent rather than broken. Happy to send a follow-up giving each pattern a unique name if that is wanted — left out here to keep this PR to the crash fix.

TemoSulava:
Both findings check out — fixed in 1a2ecbe. 18 tests pass.

**Rejected resets preserving the original password** — correct, and the gap was real. `assert not user.check_password(candidate)` only proves the *submitted* password was not stored; a bug that wrote some other value (an unusable hash, a stale password) passed it. Every rejected-reset test now snapshots `user.password` before the request and asserts the refreshed hash is byte-identical, keeping the `is_password_autoset is True` assertion.

**Exact redirect destination** — also correct: `"error_code=5125" in Location` passes no matter which page the endpoint redirected to, since the code lives in the query string. Redirects are now parsed and compared on origin, path and query:

```python
assert (location.scheme, location.netloc) == EXPECTED_ORIGIN
assert location.path.replace("/spaces//", "/spaces/", 1).rstrip("/") == expected_path
assert parse_qs(location.query, keep_blank_values=True) == expected_query
```

One deliberate deviation from "compare the exact path": the space endpoint emits `/spaces//accounts/reset-password/` — `base_host()` already returns a trailing slash and the f-strings at `space/password_management.py:127,135,145,155,165` add another. That predates this branch, so the helper normalizes *that one boundary* rather than freezing the quirk in an expected constant. Everything else is compared as emitted, so the same `//` defect appearing anywhere else — including the app endpoint — fails.

Also added while in here:
- `SPACE_BASE_PATH` is pinned alongside the host. It is read from the environment at settings-import time, so without pinning the expected paths depended on the ambient env (`SPACE_BASE_PATH=""` would have collapsed the space path onto the app one and failed 8 tests for a config reason).
- Token replay is now covered on both endpoints: a spent link must be rejected, since `PasswordResetTokenGenerator` hashes over the stored password. That is the security property this PR reshuffled control flow around, and nothing pinned it.

To confirm the new assertions actually bite rather than just passing: mutating `SPACE_ERROR_PATH` and `APP_SUCCESS_PATH` to wrong destinations fails 8 of the 18. Reverting the two view files fails the crash-path tests, as before.

For maintainers, unchanged from above and still out of scope here: the `/spaces//accounts/reset-password/` double slash, and the duplicated `name="forgot-password"` / `name="space-forgot-password"` url names. Happy to send either as a follow-up.

TemoSulava:
Valid — applied as proposed in cd35735.

`rstrip("/")` erased the one real difference between the two endpoints: the space endpoint builds its redirects with f-strings and emits `/spaces/accounts/reset-password/`, while the app endpoint uses `urljoin()` and emits `/accounts/reset-password` with no trailing slash. Stripping meant either could drift onto the other's form without a test noticing. The path is now compared as emitted, with the trailing slash carried in the space constants; the targeted `/spaces//` normalization stays.

```python
SPACE_ERROR_PATH = "/spaces/accounts/reset-password/"
SPACE_SUCCESS_PATH = "/spaces/"
APP_ERROR_PATH = "/accounts/reset-password"
APP_SUCCESS_PATH = "/sign-in"
...
assert location.path.replace("/spaces//", "/spaces/", 1) == expected_path
```

18 passed. Confirmed the assertion now discriminates on the trailing slash: adding one to `APP_ERROR_PATH` and removing one from `SPACE_SUCCESS_PATH` fails 9 of the 18.

apps/api/plane/tests/contract/app/test_password_reset.py
```diff
@@ -0,0 +1,326 @@
...
+from plane.db.models import User
+
+EXPECTED_ORIGIN = ("http", "testserver")
+
+STRONG_PASSWORD = "correct-horse-battery-staple-9x"
+WEAK_PASSWORD = "password"
+
+# Where each endpoint sends a rejected reset, and where it sends a successful one
+SPACE_ERROR_PATH = "/spaces/accounts/reset-password"
+SPACE_SUCCESS_PATH = "/spaces"
```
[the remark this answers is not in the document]
TemoSulava: Applied as proposed in cd35735 — space constants now carry the trailing slash, `rstrip("/")` is gone, targeted `/spaces//` normalization stays. 18 passed; mutating the trailing slash on `APP_ERROR_PATH`/`SPACE_SUCCESS_PATH` fails 9 of them, so the comparison discriminates on it.

apps/api/plane/authentication/views/app/password_management.py
```diff
@@ -98,79 +98,84 @@ def post(self, request):
...
-                )
-                return HttpResponseRedirect(url)
-
-            # set_password also hashes the password that the user will get
-            user.set_password(password)
-            user.is_password_autoset = False
-            user.save()
+            id = smart_str(urlsafe_base64_decode(uidb64))
+            user = User.objects.get(id=id)
+        except DjangoUnicodeDecodeError:
```
sriramveeraghanta: **low** -- Behavior change on the app endpoint: an undecodable-utf8 `uidb64` (e.g. `/auth/reset-password/not/<token>/`) now returns `error_code=5130 EXPIRED_PASSWORD_TOKEN` where `preview` returns `5125 INVALID_PASSWORD_TOKEN`, because this previously-dead handler is now reachable ahead of the tuple clause.

Concrete effect: a user who mangles a reset URL (or whose mail client truncates it) sees "Expired password token. Please try again." for a link that was never valid -- mildly misleading, and it may send them to re-request a link they already have. Both codes render as the same banner in `apps/web/helpers/authentication.helper.tsx:292-298`, so there is no functional breakage.

The PR description already flags this and offers to drop the handler instead. If you'd rather preserve `5125`, delete the `except DjangoUnicodeDecodeError` clause from both files (the tuple clause catches it via `ValueError`) and update the two `test_undecodable_uidb64_redirects` cases, which currently pin the new code.

sriramveeraghanta (commented):
Reviewed the full head revision of both view files, not just the diff. The core change is correct: `ValidationError` is already imported in both files; `DjangoUnicodeDecodeError` -> `UnicodeDecodeError` -> `ValueError`, so ordering that clause first is required and is done right in both; `git diff -w` confirms the rest of each method is a verbatim dedent with the `try` now strictly narrower. I reproduced Django's decode paths locally -- `"a"` -> `binascii.Error`/`ValueError`, `"not"` -> `b'\x9e\x8b'` -> `DjangoUnicodeDecodeError`, valid-base64 non-UUID -> `ValidationError` from `UUIDField.to_python` -- and all three now redirect instead of 500ing. The test fixtures match repo conventions (`contract` marker registered, `get_error_dict()` returns exactly the two asserted keys, no `set_password` override on `User` that would clobber `is_password_autoset`), and `_assert_redirect`'s `/spaces//` normalization tolerates both today's double slash and a future fix.

One low-severity finding inline. Two non-blocking notes:

- The space-endpoint tests live in `plane/tests/contract/app/test_password_reset.py`. Purely organizational -- the file covers both endpoints and is otherwise fine.
- The pre-existing double slash in the space redirect (`/spaces//accounts/reset-password/`, from `base_host()` already ending in `/` plus the f-string's leading `/`) is untouched by this PR and worth a separate one-line fix.

apps/api/plane/authentication/views/app/password_management.py
TemoSulava: Done — dropped the handler from both files in 67d08b9, `5125` preserved on the app endpoint.

Agreed on the reasoning: an undecodable `uidb64` was never a valid link that later expired, so `INVALID_PASSWORD_TOKEN` is the accurate answer, and the remaining `except (ValueError, ValidationError, User.DoesNotExist)` covers it for free (`DjangoUnicodeDecodeError` → `UnicodeDecodeError` → `UnicodeError` → `ValueError`, and `force_str` is its only raise site). The import is gone from both files with `smart_bytes`/`smart_str` still in use, so it stays F401-clean.

One consequence worth stating outright, since it inverts which endpoint moves: on `preview` the space endpoint *did* answer `5130` here — its `except` wrapped the whole method body, so unlike the app one it was reachable. Deleting from both therefore restores the app endpoint to its exact `preview` behaviour and changes the space endpoint from `5130` to `5125`. That is the direction I think is right, and it makes the two endpoints agree, which is the point of the PR — but it is a user-visible copy change on the space side ("Expired password token. Please try again." → "Invalid password token."), so flagging it rather than burying it. The two `test_undecodable_uidb64_redirects` cases now pin `5125` and the PR description is updated to match.

`5130` is no longer emitted anywhere in `apps/api`. I left the enum entries in `packages/constants/src/auth/index.ts:151` and both `authentication.helper.tsx` copies alone — they are `Record<Enum, …>` definitions, so a dead key breaks nothing, and removing them would strand any link already in flight from an older API.

Two test anchors added alongside, since both rejected-uidb64 fixtures now produce the same response and the old `5130` expectation was the only thing distinguishing them:

- `test_uidb64_fixtures_reach_their_branches` pins that `"a"` raises inside `urlsafe_base64_decode` (binascii, never reaches `smart_str`) while `"not"` decodes to `b"\x9e\x8b"` and raises `DjangoUnicodeDecodeError`. Without it, a change that stopped `"not"` from raising at all would leave the utf-8 branch untested with every test green — confirmed by mutation: pointing the fixture at `"aGk"` fails it.
- `test_error_code_wire_values` pins `5125`/`5020`/`5021` as literals, since the tests otherwise read the expected code from the same dict the view writes; renumbering would stay green here and break the hardcoded TS constants.

20 passed. Reverting only the two view files gives `5 failed, 15 passed`.

TemoSulava:
Thanks for reading the whole head revision rather than just the diff — the inline finding is addressed in 67d08b9 and answered in the thread. Short version: `except DjangoUnicodeDecodeError` is deleted from both files, so the app endpoint keeps `5125` exactly as `preview` returns it, and the space endpoint moves `5130` → `5125`. That direction is deliberate (an undecodable link was never valid, so it did not expire) and the PR description now states the full behaviour delta instead of the earlier note.

On the two non-blocking points:

**Test location.** Fair. I kept both endpoints in `contract/app/test_password_reset.py` because they share the fixtures, the `_assert_redirect` helper and the error-query builder — splitting them would either duplicate that or need a shared module for two files, and there is no `contract/space/` package today to move into. Happy to add one and split if you would rather have the layout mirror the view packages; say the word and it is a follow-up commit here or a separate PR, whichever you prefer.

**The `/spaces//accounts/reset-password/` double slash.** Agreed it is a one-liner and agreed it does not belong in a crash-fix PR. `base_host()` already returns a trailing slash, and the five f-strings at `space/password_management.py:127,135,145,155,165` add another. The test helper deliberately normalizes only that one boundary, so the fix will not need any test churn beyond deleting that `.replace(...)` — every other path is compared as emitted. Happy to send it as a separate PR against `preview` whenever you want it.

The duplicated `name="forgot-password"` / `name="space-forgot-password"` in `plane/authentication/urls.py:122-138` is the other latent one from earlier — nothing in the repo reverses those names today, so it is a rename plus a `reverse()` audit rather than a behaviour change. Also happy to take it separately.

PR #9673, opened 2026-08-23
chore(ci): build images natively per architecture and merge by digest

sriramveeraghanta:
### Description

Every image in `Branch Build CE` now builds **one platform per job on a runner native to that architecture**, pushes an untagged manifest addressed only by its digest, and is stitched into the tagged multi-arch index by a paired `branch_merge_*` job using `docker buildx imagetools create`.

**Why** — arm64 was previously produced either under QEMU emulation or on the shared `makeplane/plane-dev` Docker Build Cloud builder. Emulated arm64 runs 5–10× slower than native, and the Build Cloud builder is shared across repos, has no owner here, and is load-bearing: arm64 is forced on for non-prerelease Releases, so a broken shared builder blocks releases specifically. Both paths are gone — `buildx-driver` is pinned to the local `docker-container` driver, since a leg building its own architecture natively gains nothing from a remote builder.

The arm64 trigger condition is unchanged (`arm64: true`, or a non-prerelease `Release`), and **nothing externally visible changes** — the images carry the same names and the same tags, so every consumer (compose files, the AIO Dockerfile, install scripts) keeps referencing them exactly as before.

#### What changed

- **Setup job.** `BUILDX_DRIVER` / `BUILDX_VERSION` / `BUILDX_PLATFORMS` / `BUILDX_ENDPOINT` are replaced by `gh_build_matrix` (a JSON array of `{arch, platform, runner}`) and `gh_expected_platforms`. The Docker Build Cloud endpoint is removed outright rather than left as a no-op switch.
- **Six build jobs** (admin, web, space, live, api, proxy) become matrix jobs with `push-by-digest`, a per-architecture cache scope, and a per-architecture digest artifact.
- **Six new merge jobs** collect those digests, apply the tags, and verify the resulting index.
- **AIO is split in two.** `prepare_aio_assets` runs `build.sh` once and uploads `aio-assets-dist`; `branch_build_push_aio` is the matrix build; `branch_merge_aio` publishes the index.
- Runners move to GitHub-hosted `ubuntu-24.04` / `ubuntu-24.04-arm` (both 4 vCPU / 16 GB, and free on public repos), the shared build actions are bumped to `v1.6.0`, and every job gets a `timeout-minutes`.

#### Three details that are load-bearing rather than cosmetic

- **`cache-scope-suffix` per architecture.** The registry cache manifest is not platform-indexed, so both legs writing `:buildcache` would overwrite each other and roughly half the legs would go cold every run — which could make the split *slower* than a single job.
- **The merge asserts the platform set.** A merge that produced a single-platform index still exits `0`, so it is checked rather than assumed.
- **Tag consumers depend on merge jobs, not build jobs.** The tag does not exist until the merge runs. AIO pulls the other images in by tag, so it now needs the six merges; `upload_build_assets` and `publish_release` were repointed the same way. GitHub does not warn when a `needs:` entry is missing — the dependent job simply runs early against a stale or absent tag.

`publish_release` deliberately keeps **no** status function: a failed build leg skips its merge job, and the default `success()` gate rejects a skipped need. `branch_merge_aio` is intentionally absent from its `needs` so one flaky AIO leg cannot silently discard a finished release.

### Type of Change

- [x] Performance improvements

### Test Scenarios

`actionlint` passes clean on the workflow. The rest can only be proven by running it:

1. Dispatch `Branch Build CE` on this branch with defaults (`arm64: false`). Each image's matrix should have a single amd64 leg, and each `branch_merge_*` should publish an index containing only `linux/amd64`.
2. Re-dispatch with `arm64: true`. Confirm two legs per image, the arm64 one on `ubuntu-24.04-arm`, then:
   ```
   docker buildx imagetools inspect makeplane/plane-frontend:<branch>
   ```
   Expect exactly `linux/amd64` and `linux/arm64`. Spot-check `plane-backend` and `plane-proxy` too.
3. Confirm the arm64 legs built **natively** — no `qemu-x86_64` or binfmt lines in the log, and wall time in line with the amd64 leg rather than 5–10× it.
4. Confirm the legs wrote separate cache refs (`:buildcache-amd64`, `:buildcache-arm64`) and that a second run imports rather than rebuilding from scratch.
5. Dispatch with `aio_build: true` and confirm `prepare_aio_assets` runs once, both AIO legs consume the same `aio-assets-dist` artifact, and the AIO build starts only after all six merges have published their tags.
6. Deliberately fail one build leg and confirm its merge job is skipped and `publish_release` does not run.
7. Run a `Release` build end-to-end and confirm the release is created with the expected assets.

### References

- Depends on `makeplane/actions@v1.6.0` (`build-push` digest mode + the new `merge-manifest` action).

https://claude.ai/code/session_01U45dsLe5GFmTsfpifuNcGQ

## Summary by CodeRabbit

* **Build & Release**
  * Improved multi-architecture image builds and manifest publishing.
  * Updated build runners and streamlined architecture-specific caching and artifacts.
  * Ensured AIO builds and asset uploads run after the relevant images are assembled.
  * Improved release gating so releases proceed only after required build steps succeed.

### DIFF
--- a/.github/workflows/build-branch.yml
+++ b/.github/workflows/build-branch.yml
@@ -50,13 +50,14 @@ env:
 jobs:
   branch_build_setup:
     name: Build Setup
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     outputs:
       gh_branch_name: ${{ steps.set_env_variables.outputs.TARGET_BRANCH }}
-      gh_buildx_driver: ${{ steps.set_env_variables.outputs.BUILDX_DRIVER }}
-      gh_buildx_version: ${{ steps.set_env_variables.outputs.BUILDX_VERSION }}
-      gh_buildx_platforms: ${{ steps.set_env_variables.outputs.BUILDX_PLATFORMS }}
-      gh_buildx_endpoint: ${{ steps.set_env_variables.outputs.BUILDX_ENDPOINT }}
+      # Per-architecture native build matrix. Every image builds one platform per
+      # job on a runner native to it, so there is no single buildx driver/platform
+      # pair for the run and no Docker Build Cloud endpoint.
+      gh_build_matrix: ${{ steps.set_env_variables.outputs.BUILD_MATRIX }}
+      gh_expected_platforms: ${{ steps.set_env_variables.outputs.EXPECTED_PLATFORMS }}
 
       dh_img_web: ${{ steps.set_env_variables.outputs.DH_IMG_WEB }}
       dh_img_space: ${{ steps.set_env_variables.outputs.DH_IMG_SPACE }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-branch.yml
```diff
@@ -153,18 +176,47 @@ jobs:
...
+        with:
+          build-release: ${{ needs.branch_build_setup.outputs.build_release }}
+          build-prerelease: ${{ needs.branch_build_setup.outputs.build_prerelease }}
+          release-version: ${{ needs.branch_build_setup.outputs.release_version }}
+          dockerhub-username: ${{ secrets.DOCKERHUB_USERNAME }}
+          dockerhub-token: ${{ secrets.DOCKERHUB_TOKEN }}
+          docker-image-owner: makeplane
+          docker-image-name: ${{ needs.branch_build_setup.outputs.dh_img_admin }}
+          digest-artifact-pattern: digest-admin-*
+          expected-platforms: ${{ needs.branch_build_setup.outputs.gh_expected_platforms }}
```
lovelace3, line 206: Non-blocking: this passes `expected-platforms` to the merge action, but its verifier only checks that every expected platform is present; it accepts a manifest with extra platforms. Since this workflow describes the check as asserting the platform *set*, could we either compare the complete deduplicated list in `merge-manifest`, or soften the contract to say it verifies the expected platforms are included? An exact check would also catch an unexpected digest artifact being included.

lovelace3 (commented):
## Review summary

Nice refactor: the native per-architecture matrix, digest artifacts, and merge dependencies correctly remove the tag-publication race for AIO and release consumers.

I left one non-blocking inline comment about making the manifest validation match the stated "platform set" contract.

Validation run: `actionlint` passed; workflow YAML parsed successfully; all PR checks are green.

lovelace3 (approved):
Approved after review. The manifest platform-set validation observation is non-blocking.

PR #9686, opened 2026-08-26
feat: allow private project views for the owning user

IMustafaZeynali:
### Description
Project views already had a private/public `access` field and list filtering by owner, but `access` was read-only on the API and the create/edit form had no access control. Users could not actually create personal private views.

This PR enables private project views end-to-end:
- Make `access` writable on `IssueViewSerializer` so clients can create/update Private (`0`) or Public (`1`) views
- Add a Public/Private access selector to the project view form (same pattern as Pages)
- Return `404` when retrieving a view that is not visible to the requesting user (e.g. another member’s private view)

Private views remain visible only to the owner; public views stay shared with project members.

### Type of Change
- [ ] Code refactoring
- [ ] Documentation update

### Test Scenarios
- Create a project view with **Private** access → it appears in the creator’s Views list with the lock icon
- As another project member, open the same project’s Views list → the private view is not listed
- Open the private view URL directly as another member → receives `404`
- Create a **Public** view → both members can see and open it
- Owner edits a private view and switches access to **Public** → other members can see it afterward
- Owner can still update/delete their own private view

## Summary by CodeRabbit

* **New Features**
  * Added access controls to view forms, including localized access labels.
  * View access settings can now be updated when editing a view.

* **Bug Fixes**
  * Requests for unavailable views now return a clear “Not Found” response instead of continuing with missing data.

### DIFF
--- a/apps/api/plane/app/serializers/view.py
+++ b/apps/api/plane/app/serializers/view.py
@@ -64,7 +64,6 @@ class IssueViewSerializer(DynamicBaseSerializer):
             "project",
             "query",
             "owned_by",
-            "access",
             "is_locked",
         ]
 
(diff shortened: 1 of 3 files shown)


vihar:
Appreciate this. Private views sit in Plane's paid plans (https://plane.so/pricing), so it isn't a change we can take into the community edition. Closing here, thank you for the contribution.

PR #9706, opened 2026-08-28
[INFRA-777] fix(security): scope ProjectArchiveUnarchiveAPIEndpoint to the URL project

mguptahub:
## Summary
- `ProjectArchiveUnarchiveAPIEndpoint` used `ProjectBasePermission`, whose POST branch is written for project creation and checks only workspace-level role (ADMIN/MEMBER) with no `project_id` binding at all.
- Since archive is a POST but is not creation, any active workspace member could archive (and delete every `UserFavorite` row on) any project in the workspace — including a fully private one they have no `ProjectMember` row on and cannot otherwise read.
- DELETE (unarchive) was never affected — it falls through to `ProjectBasePermission`'s non-POST branch, which does check project membership. So this was specifically an asymmetry between the two verbs of the same endpoint. That branch also turned out to be stricter than intended (project ADMIN, or workspace-ADMIN-plus-project-membership — not just an active project MEMBER), which this fix corrects too.
- Fix: a dedicated `ProjectArchiveUnarchivePermission`, scoped to `project_id`, mirroring the app-layer twin's gate (`allow_permission([ROLE.ADMIN, ROLE.MEMBER])`) on both `post` and `delete` — instead of touching `ProjectBasePermission` itself, which is correctly relied on by the actual project-create endpoint and by `ProjectDetailAPIEndpoint`'s patch/delete.
- Checked thoroughly for an existing fix first: `ProjectMemberPermission` has the same POST-branch shape but is a different class, already tracked and being fixed by the open `f63c` cluster's PR #9596 — confirmed by reading that PR's actual diff, which explicitly leaves `ProjectBasePermission` untouched.

## Test plan
- [x] New regression tests in `apps/api/plane/tests/contract/api/test_project_archive_scope.py`: a workspace member with no project membership cannot archive or unarchive a private project; an active project member (not just admin) can do both
- [x] Verified fail-before/pass-after via `git stash` — the archive-bypass test fails against pre-fix code, and the positive-control test also caught a genuine pre-existing over-restriction (a plain project member couldn't unarchive their own project before this fix)
- [x] Related project contract test suites still green (18 passed)
- [x] `ruff check` / `ruff format --check` clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Strengthened project archive and unarchive access controls.
  - Workspace members without project-level membership can no longer archive or unarchive projects.
  - Authorized project members retain access to both actions.

### DIFF
--- a/apps/api/plane/api/views/project.py
+++ b/apps/api/plane/api/views/project.py
@@ -49,7 +49,11 @@ from plane.api.serializers import (
     ProjectCreateSerializer,
     ProjectUpdateSerializer,
 )
-from plane.app.permissions import ProjectBasePermission, WorkSpaceAdminPermission
+from plane.app.permissions import (
+    ProjectArchiveUnarchivePermission,
+    ProjectBasePermission,
+    WorkSpaceAdminPermission,
+)
 from plane.utils.openapi import (
     project_docs,
     PROJECT_ID_PARAMETER,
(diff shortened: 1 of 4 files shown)


mguptahub:
/code-review: clean, no findings. Verified the permission class's query logic against both ProjectBasePermission's non-POST branch and the app-layer allow_permission([ROLE.ADMIN, ROLE.MEMBER]) decorator, confirmed view.project_id/view.workspace_slug are valid BaseAPIView properties, and confirmed ProjectBasePermission/ProjectMemberPermission are correctly left untouched.

PR #9709, opened 2026-08-28
[INFRA-779] fix(security): reject authority-relative next_path redirects

mguptahub:
## Summary
- **Server** (`apps/api/plane/utils/path_validator.py`, `validate_next_path`): calls `urlparse(next_path)` and only extracts `.path` when `scheme` or `netloc` is truthy. For `"///example.com/"` (three or more leading slashes), `urlparse()` returns both `scheme` and `netloc` empty — verified directly — so that branch never fires and the raw string passes every remaining check unchanged. Fixed by rejecting any `next_path` starting with `//` outright, right after the existing "must start with `/`" check.
- **Client** (`apps/web/core/lib/wrappers/authentication-wrapper.tsx`, `isValidURL`): only regex-blocked a literal `http(s)`/`ftp` scheme prefix, so the same authority-relative string passed and was handed to `router.push()`. Fixed by resolving the URL against `location.origin` and requiring the result to actually still be same-origin, instead of pattern-matching the input string.
- Browsers resolve a leading `//` as authority-relative even when neither validator's own URL parsing detected a host — the accepted value silently navigates off-domain post-login, a same-origin-trust phishing vector (attacker link visibly belongs to the Plane host; victim is already authenticated; click lands on attacker content).
- Checked the advisory's other listed `next_path` consumers (auth-form components, oauth hooks, `api.service.ts`) — they only forward the value to a server-side auth redirect or a hidden form field, no independent client-side navigation, so they're covered by the server-side fix and don't need changes.

## Test plan
- [x] 8 new server-side tests in `apps/api/plane/tests/unit/utils/test_path_validator.py`: rejects 3/4/5-leading-slash forms, positive controls confirming the pre-existing 2-slash and scheme-prefixed cases still safely downgrade to a bare path (not a new rejection), genuine relative paths still accepted
- [x] Verified fail-before/pass-after via `git stash` — all 3 malicious-path cases fail against pre-fix code
- [x] Client-side fix verified empirically via Node's URL parser (WHATWG-compliant, matches browser behavior) — confirmed it rejects all authority-relative forms and the original blocked absolute-scheme forms, while still accepting genuine relative paths. No test harness exists for `apps/web` in this repo to add an automated regression test.
- [x] `ruff check`/`ruff format --check` clean (Python); `oxlint`/`oxfmt --check` clean (TypeScript); the file doesn't appear in `check:types` output (all pre-existing errors there are unrelated missing-export issues in `@plane/constants`/`@plane/types`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redirect validation to block authority-relative paths that could send users to external domains.
  * Strengthened URL validation against obfuscated paths containing tab, carriage return, or line feed characters.
  * Ensured navigation stays within the current site while preserving valid relative paths and supported URL formats.

* **Tests**
  * Added coverage for unsafe multi-slash and obfuscated paths, supported URL cases, and genuine relative paths.

### DIFF
--- a/apps/api/plane/tests/unit/utils/test_path_validator.py
+++ b/apps/api/plane/tests/unit/utils/test_path_validator.py
@@ -68,3 +68,25 @@ class TestValidateNextPathAuthorityRelative:
         same-origin path — this fix must not change that behavior."""
         assert validate_next_path("https://evil.com/phish") == "/phish"
         assert validate_next_path("http://evil.com/phish") == "/phish"
+
+    @pytest.mark.parametrize(
+        # A tab between each slash defeats both urlparse()'s own netloc
+        # detection (verified directly: "/\t/\t/evil.com" -> scheme='',
+        # netloc='') AND a literal next_path.startswith("//") check, since
+        # the second character is a tab, not a slash. Per the WHATWG URL
+        # spec, browsers strip every ASCII tab/CR/LF from a URL before
+        # parsing it, so what the browser actually navigates on is
+        # "///evil.com" — authority-relative, off-origin — even though this
+        # function never sees a literal "//" prefix.
+        "obfuscated_next_path",
+        [
+            "/\t/\t/evil.com",
+            "/\r/\r/evil.com",
+            "/\n/\n/evil.com",
+        ],
+    )
+    def test_rejects_tab_cr_lf_obfuscated_authority_relative_paths(self, obfuscated_next_path):
+        assert validate_next_path(obfuscated_next_path) == "", (
+            f"{obfuscated_next_path!r} must be rejected — browsers strip tab/CR/LF before parsing, "
+            "so this collapses to an authority-relative '///evil.com' navigation"
+        )
--- a/apps/api/plane/tests/unit/utils/test_path_validator.py
+++ b/apps/api/plane/tests/unit/utils/test_path_validator.py
@@ -1,0 +1,70 @@
+# Copyright (c) 2023-present Plane Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
+"""Regression test for authority-relative open-redirect via next_path.
+
+Root cause: urlparse("///example.com/") returns both scheme and netloc
+empty (a quirk of Python's URL parser for exactly-three-or-more leading
+slashes), so validate_next_path's "extract only the path component" branch
+(gated on scheme or netloc being truthy) never fires, and the original,
+unmodified "///example.com/" string passes every remaining check unchanged.
+Browsers still resolve a leading "//" as authority-relative against an
+http(s) base, so the accepted value silently navigates off-domain.
+
+Fixed by rejecting any next_path starting with "//" outright, after the
+existing "must start with /" check.
+"""
+
+import pytest
+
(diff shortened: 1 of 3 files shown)


mguptahub:
Confirmed and fixed in 838341158f — verified the exact bypass empirically before fixing: `urlparse("/\t/\t/evil.com")` does return `scheme='', netloc=''` (matching Python's parsing, not what I'd first assumed for a single-tab variant), and per the WHATWG URL spec browsers strip every ASCII tab/CR/LF from a URL before parsing it — so what actually gets navigated to is the tab-stripped `///evil.com`, an authority-relative bypass, obfuscated with whitespace instead of extra literal slashes.

Fixed by stripping tab/CR/LF alongside the existing backslash removal, before both the `urlparse()` netloc check and the new `//` check run. 3 new regression tests, fail-before verified against the code before this commit.

Good catch — this is exactly the same bug class the PR set out to close, just via a different obfuscation technique.

mguptahub:
/code-review (re-run after the tab/CR/LF fix): clean, no findings. Tested an extensive battery of bypass attempts (multi-slash, tab/CR/LF, backslash+tab, userinfo tricks, percent-encoded forms, scheme-only forms, Unicode division-slash) against both the Python and TS logic empirically — nothing survived. Also confirmed the check ordering matches WHATWG's actual pre-parse normalization steps, and that location.origin is unreachable during SSR in the component (returns its loading state before getWorkspaceRedirectionUrl() is ever called).

mguptahub:
Proactive follow-up (not a finding on this PR, but the plane-ee port's /code-review flagged it and it applies equally here): replaced the char-by-char tab/CR/LF-stripping rebuild with `str.translate(str.maketrans("", "", "\t\r\n"))` in 124103743e — verified equivalent, single-pass, matches the terse style of the adjacent `.replace("\\", "")` line.

Also verified (same review raised delegating to Django's `url_has_allowed_host_and_scheme` instead of hand-rolling): Django's own `_url_has_allowed_host_and_scheme` doesn't strip tab/CR/LF before its `startswith("///")` check either, so that would reintroduce the bypass this PR fixed — declined for the same reason on both repos.

mguptahub:
Proactive follow-up (findings from the plane-ee port's /code-review, confirmed applicable here too):

- The client-side fix had a narrower but real analogous gap: it resolves against the real `location.origin`, so it only bypasses when the actual deployment's origin scheme matches the forged scheme (e.g. `next_path=http:evil.com` on a deployment actually served over plain http — not uncommon for self-hosted instances). Verified directly. Fixed in 43f4dc0a6a by delegating to the shared `isValidNextPath` (`@plane/utils`, already used by `apps/space`) instead of a local reimplementation — it requires a literal leading `/` before any URL-based comparison, so it doesn't depend on which scheme the real origin uses.
- Filed **INFRA-781** for a separate, pre-existing issue found in the same review pass: `get_safe_redirect_url` doesn't escape `next_path` before interpolating it into a query string, so `&`/`=`/`#` can inject extra query params. Not introduced by this PR, needs its own impact investigation.

mguptahub:
/code-review (re-run after the delegation fix): clean, no findings. Independently verified via runtime execution (Python + Node) that all 8 tests pass against the real implementation and no other client-side next_path consumer retains the old vulnerable pattern.

PR #9726, opened 2026-08-31
refactor: migrate web @plane/propel Tooltip to @makeplane/propel

anmolsinghbhatia:
### Description

Moves all **144 `<Tooltip>` call sites in `apps/web`** off the in-repo `@plane/propel/tooltip` and onto the published `@makeplane/propel/components/tooltip`. Part of the in-repo → published Propel wave (PR 3 in `docs/plane-propel-to-makeplane-plan.md`). No `@plane/ui` migration is mixed in, and `packages/propel` is untouched.

**Prop mapping, applied identically on every site:**

| In-repo | Published |
|---|---|
| `tooltipContent` | `label` (string) |
| `tooltipHeading` | folded into `label` + `layout="stacked"` |
| `position` | `side` / `align` — `top`/`center` omitted, it's the published default |
| `isMobile` | `disabled` (OR'd with any existing `disabled`) |
| `renderByDefault` | dropped — the in-repo Tooltip declared it but never read it |
| `className` | dropped, or mapped to `sideOffset` / `alignOffset` (3 sites) |

Resulting distribution: 144 `label`, 92 `disabled`, 71 `layout`, 26 `side`, 13 `align`, 3 `shortcut`, 2 `sideOffset`, 1 `alignOffset`, **0 `className`**.

**`layout` choice.** `single` is `whitespace-nowrap`, so a long label would render as one very wide chip and overflow the viewport — the in-repo tooltip wrapped at `max-w-xs`. So `single` is used only for short, bounded copy (39 sites); `stacked` wherever the label interpolates unbounded data such as work item names, page titles, URLs, or joined label lists.

**⚠️ Behavior change — tooltip open delay is now 600ms, was 200ms.** The in-repo component defaulted `openDelay = 200` and no call site overrode it. `delay` is deliberately not set per site, so every tooltip now takes Base UI's published `OPEN_DELAY` of 600ms. This is a deliberate move to the design-system default, but it is user-perceptible across the whole app. If we'd rather keep the old feel, a single `<TooltipProvider delay={200}>` in `apps/web/app/provider.tsx` covers all sites.

**ReactNode content (15 sites).** The published `label` is a `string`, so nothing was stuffed into it:
- **Editor toolbars (5)** — `item.name` + `<kbd>{shortcut}</kbd>` maps cleanly onto the published `shortcut` prop.
- **`<br/>` / fragment copy (6)** — folded into one wrapping `stacked` string.
- **`stickies/action-bar.tsx`** — this was never a tooltip; it rendered a full `<StickyNote>` hover preview. Moved to `PreviewCard` / `PreviewCardTrigger` / `PreviewCardContent`. **This is the one site worth a design review.**
- **`exporter/export-form.tsx`** — dead code inside a commented-out JSX block; updated so no stale `@plane/propel` reference is left behind.

**Type narrowing.** `DropdownButtonProps.tooltipContent`, date-range's `customTooltipContent`, and the two `rich-filters` components narrow from `ReactNode` to `string`. `rich-filters/root.tsx` guards `filterConfig.tooltipContent`, which stays `ReactNode` in `@plane/types` (out of scope for a web-only change).

**Nullable labels.** Folded labels use `?? ""` wherever the source expression is nullable, so they can't render the literal string `"undefined"` in places the in-repo tooltip rendered nothing.

**Known follow-ups (not in this PR):**
- `comments/card/display.tsx:130` keeps `delay={200}` — the one Tooltip already on `@makeplane/propel` before this PR, so it's now the only site at 200ms.
- `renderToolTipByDefault` / `renderByDefault` is now dead plumbing across ~15 dropdown files. Only the bindings that lint flagged were removed; ripping out the public dropdown prop belongs in its own PR.

### Type of Change

- [x] Code refactoring
- [ ] Documentation update

Recipe verified by rendering the published `Tooltip` against this repo's own compiled token CSS in both themes — `single` and `stacked`, light and dark. Surface, border, and text all flip correctly via `bg-layer-2` / `border-subtle-1` / `text-primary`.

### Test Scenarios

- **Open delay:** hover any tooltip trigger (sidebar project item, issue list row) and confirm the ~600ms delay is acceptable, or flag it so we add a `TooltipProvider`.
- **Long labels wrap:** hover a work item with a long title in the list, kanban, spreadsheet, and gantt layouts — the tooltip should wrap at `max-w-60`, not stretch off-screen.
- **Folded headings:** hover a label chip (`Labels: bug, feature`), a priority dropdown (`Priority: Urgent`), and sub-work-item/attachment/link counts on an issue row — the heading should read as a prefix, not a dangling `"Heading: "`.
- **Keyboard shortcuts:** hover bold/italic in the comment editor, sticky editor, and page editor toolbars — the shortcut should render in the dimmed `shortcut` slot beside the name.
- **Sticky hover preview:** with a recent sticky present, hover the sticky action bar button — the `PreviewCard` should show the sticky with its gradient fade. Open the recent-sticky panel and confirm the preview no longer appears.
- **Mobile:** on a touch viewport, confirm tooltips do not open (`isMobile` now maps to `disabled`).
- **Light + dark:** toggle themes and confirm tooltip surfaces adapt on both.
- **Positioned tooltips:** hover the favorites sidebar items and the sign-out control — offsets previously done via `className` now come from `sideOffset` / `alignOffset`.

### References

- Plan: `docs/plane-propel-to-makeplane-plan.md` (PR 3 — Tooltip)
- Mapping: `docs/plane-propel-to-makeplane-mapping.md`
- Propel docs: https://propel.plane.so/components/tooltip

---

## Summary by CodeRabbit

* **UI Improvements**
  * Updated tooltips across the application for more consistent presentation, positioning, and alignment.
  * Tooltips now use clearer labels and improved stacked layouts.
  * Tooltips are automatically disabled on mobile devices when appropriate.
  * Added accessible labels to key controls, including layout and project actions.
  * Keyboard shortcuts are displayed more clearly in editor toolbar tooltips.
  * Recent-item previews now appear in a dedicated preview card.

### DIFF
--- a/apps/web/app/(all)/[workspaceSlug]/(projects)/extended-project-sidebar.tsx
+++ b/apps/web/app/(all)/[workspaceSlug]/(projects)/extended-project-sidebar.tsx
@@ -117,6 +117,7 @@ export const ExtendedProjectSidebar = observer(function ExtendedProjectSidebar()
               <Tooltip label={t("create_project")}>
                 <button
                   type="button"
+                  aria-label={t("create_project")}
                   className="flex-shrink-0 rounded-sm p-0.5 text-tertiary transition-colors hover:bg-layer-1 hover:text-secondary"
                   onClick={() => {
                     setIsProjectModalOpen(true);
--- a/apps/web/app/(all)/[workspaceSlug]/(projects)/extended-project-sidebar.tsx
+++ b/apps/web/app/(all)/[workspaceSlug]/(projects)/extended-project-sidebar.tsx
@@ -13,7 +13,7 @@ import { useTranslation } from "@plane/i18n";
 import { EmptyStateCompact } from "@plane/propel/empty-state";
 import { PlusIcon, SearchIcon } from "@plane/propel/icons";
 import { TOAST_TYPE, setToast } from "@plane/propel/toast";
-import { Tooltip } from "@plane/propel/tooltip";
+import { Tooltip } from "@makeplane/propel/components/tooltip";
 import { copyUrlToClipboard, orderJoinedProjects } from "@plane/utils";
 // components
 import { CreateProjectModal } from "@/components/project/create-project-modal";
@@ -114,7 +114,7 @@ export const ExtendedProjectSidebar = observer(function ExtendedProjectSidebar()
           <div className="flex items-center justify-between">
             <span className="py-1.5 text-13 font-semibold text-tertiary">Projects</span>
             {isAuthorizedUser && (
-              <Tooltip tooltipHeading={t("create_project")} tooltipContent="">
+              <Tooltip label={t("create_project")}>
                 <button
                   type="button"
                   className="flex-shrink-0 rounded-sm p-0.5 text-tertiary transition-colors hover:bg-layer-1 hover:text-secondary"
--- a/apps/web/core/components/base-layouts/layout-switcher.tsx
+++ b/apps/web/core/components/base-layouts/layout-switcher.tsx
@@ -30,12 +30,14 @@ export function LayoutSwitcher(props: Props) {
 
   return (
     <div className="flex items-center gap-1 rounded-md bg-layer-3 p-1">
-      {BASE_LAYOUTS.filter((l) => (layouts ? layouts.includes(l.key) : true)).map((layout) => {
+      {BASE_LAYOUTS.map((layout) => {
+        if (layouts && !layouts.includes(layout.key)) return null;
         const Icon = layout.icon;
         return (
           <Tooltip key={layout.key} label={t(layout.label)} disabled={isMobile}>
             <button
               type="button"
+              aria-label={t(layout.label)}
               className={cn(
                 "group grid h-5.5 w-7 place-items-center overflow-hidden rounded-sm transition-all hover:bg-layer-transparent-hover",
                 {
--- a/apps/web/core/components/base-layouts/layout-switcher.tsx
+++ b/apps/web/core/components/base-layouts/layout-switcher.tsx
@@ -5,7 +5,7 @@
  */
 
 import { useTranslation } from "@plane/i18n";
-import { Tooltip } from "@plane/propel/tooltip";
+import { Tooltip } from "@makeplane/propel/components/tooltip";
 import type { TBaseLayoutType } from "@plane/types";
 import { cn } from "@plane/utils";
 import { usePlatformOS } from "@/hooks/use-platform-os";
@@ -33,7 +33,7 @@ export function LayoutSwitcher(props: Props) {
       {BASE_LAYOUTS.filter((l) => (layouts ? layouts.includes(l.key) : true)).map((layout) => {
         const Icon = layout.icon;
         return (
-          <Tooltip key={layout.key} tooltipContent={t(layout.label)} isMobile={isMobile}>
+          <Tooltip key={layout.key} label={t(layout.label)} disabled={isMobile}>
             <button
               type="button"
               className={cn(
--- a/apps/web/core/components/issues/issue-layouts/filters/header/layout-selection.tsx
+++ b/apps/web/core/components/issues/issue-layouts/filters/header/layout-selection.tsx
@@ -7,7 +7,7 @@
 // plane imports
 import { ISSUE_LAYOUTS } from "@plane/constants";
 import { useTranslation } from "@plane/i18n";
-import { Tooltip } from "@plane/propel/tooltip";
+import { Tooltip } from "@makeplane/propel/components/tooltip";
 import type { EIssueLayoutTypes } from "@plane/types";
 import { cn } from "@plane/utils";
 // components
@@ -34,7 +34,7 @@ export function LayoutSelection(props: Props) {
   return (
     <div className="flex items-center gap-1 rounded-md bg-layer-3 p-1">
       {ISSUE_LAYOUTS.filter((l) => layouts.includes(l.key)).map((layout) => (
-        <Tooltip key={layout.key} tooltipContent={t(layout.i18n_title)} isMobile={isMobile}>
+        <Tooltip key={layout.key} label={t(layout.i18n_title)} disabled={isMobile}>
           <button
             type="button"
             className={cn(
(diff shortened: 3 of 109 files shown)


apps/web/core/components/base-layouts/layout-switcher.tsx
```diff
@@ -33,7 +33,7 @@ export function LayoutSwitcher(props: Props) {
       {BASE_LAYOUTS.filter((l) => (layouts ? layouts.includes(l.key) : true)).map((layout) => {
```
anmolsinghbhatia: Fixed in ddc3383062 — `BASE_LAYOUTS.map()` now skips non-selected layouts with an early `return null` instead of pre-filtering, so it is a single pass. (Note this chain was pre-existing; the migration only touched the `<Tooltip>` line in this file.)

apps/web/core/components/issues/issue-layouts/filters/header/layout-selection.tsx
```diff
@@ -34,7 +34,7 @@ export function LayoutSelection(props: Props) {
   return (
     <div className="flex items-center gap-1 rounded-md bg-layer-3 p-1">
       {ISSUE_LAYOUTS.filter((l) => layouts.includes(l.key)).map((layout) => (
```
anmolsinghbhatia: Fixed in ddc3383062 — same treatment as `layout-switcher.tsx`: `ISSUE_LAYOUTS.map()` with an early `return null` rather than `.filter().map()`. Pre-existing chain, not introduced by the migration.

apps/web/core/components/issues/issue-layouts/properties/labels.tsx
```diff
@@ -95,14 +89,14 @@ function LabelSummary({ isMobile, fullWidth, noLabelBorder, disabled, projectLab
...
-        isMobile={isMobile}
-        position="top"
-        tooltipHeading={t("common.labels")}
-        tooltipContent={projectLabels
-          ?.filter((l) => value.includes(l?.id))
-          .map((l) => l?.name)
-          .join(", ")}
-        renderByDefault={false}
+        label={`${t("common.labels")}: ${
+          projectLabels
```
anmolsinghbhatia: Fixed in ddc3383062 — this one was actually three passes (`filter` → `map` → `join`) over an unbounded project label list, so it is now a single `reduce` into `selectedLabelNames` above the return, joined once at the call site.

apps/web/core/components/pages/header/syncing-badge.tsx
```diff
@@ -37,22 +37,20 @@ export function PageSyncingBadge({ syncStatus }: Props) {
   const badgeContent = {
```
anmolsinghbhatia: Fixed in ddc3383062 — hoisted to module scope as `BADGE_CONTENT`. It holds only static strings, so nothing is lost by building it once.

apps/web/core/components/stickies/action-bar.tsx
```diff
@@ -61,46 +72,38 @@ export const StickyActionBar = observer(function StickyActionBar() {
...
+            recentStickyButton
+          ) : (
+            <PreviewCard>
+              <PreviewCardTrigger render={recentStickyButton} />
+              <PreviewCardContent side="left">
+                <div className="relative max-h-[150px] overflow-hidden rounded-lg">
+                  <StickyNote
+                    className="w-full"
+                    workspaceSlug={workspaceSlug.toString()}
+                    stickyId={newSticky ? activeStickyId : recentStickyId || ""}
```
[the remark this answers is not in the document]
anmolsinghbhatia: Fixed in ddc3383062. Confirmed the analysis: the outer `recentStickyId &&` guard proves it truthy, so `|| ""` is dead. Applied the same removal to the second `stickyId` at line ~140, where the guard is `(newSticky || (showRecentSticky && recentStickyId))` — on every path that actually reads `recentStickyId` it is truthy there too. Behaviour is identical either way since `StickyNote` types `stickyId` as `string | undefined` and branches on `stickyId ? stickies[stickyId] : …`, treating `""` and `undefined` the same.

apps/web/app/(all)/[workspaceSlug]/(projects)/extended-project-sidebar.tsx
```diff
@@ -114,7 +114,7 @@ export const ExtendedProjectSidebar = observer(function ExtendedProjectSidebar()
           <div className="flex items-center justify-between">
             <span className="py-1.5 text-13 font-semibold text-tertiary">Projects</span>
             {isAuthorizedUser && (
-              <Tooltip tooltipHeading={t("create_project")} tooltipContent="">
+              <Tooltip label={t("create_project")}>
```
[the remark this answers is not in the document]
Comment (anmolsinghbhatia) [reply apps/web/app/(all)/[workspaceSlug]/(projects)/extended-project-sidebar.tsx:117]: Fixed in ddc3383062 — `aria-label` added at all four sites.

Verified the underlying claim by rendering the published `Tooltip` and dumping the DOM rather than reading the source, and it is worse than "label does not name the trigger": Base UI wires **no** aria between trigger and popup at all.

Trigger:
```html
<button id="icononly" data-popup-open="" data-base-ui-tooltip-trigger="">
```
Popup:
```html
<div … role="tooltip">   <!-- no id -->
```

No `aria-describedby`, and the popup has no `id` to point at — so the tooltip copy is invisible to assistive tech, not merely non-naming. Icon-only triggers announced as bare "button". Re-ran the same check with `aria-label` present and it survives Base UI’s `mergeProps` intact.

For the record this is pre-existing rather than a regression from this PR: the in-repo tooltip sat on the same unwired Base UI trigger (`@base-ui-components/react@1.0.0-beta.3` has zero `describedby` wiring in `tooltip/` either). Fixing it here since the PR touches these sites.

On `access-field.tsx`: rather than defaulting an optional label, `accessSpecifiers` now requires `i18n_label` and the optional pre-i18n `label` is gone, so `const label = t(access.i18n_label)` is non-empty by construction and feeds both the tooltip copy and the accessible name. Its only caller (`page-form.tsx`, via `PAGE_ACCESS_SPECIFIERS`) already types `i18n_label` as required, so this also retires the `// TODO: Remove label once i18n is done`.

codingwolf-at (approved):

PR #9754, opened 2026-09-03
fix: scope un-themed editor color fallback so dark-mode ramp wins

### LINKED ISSUE #9753: Stickies unreadable in dark mode: un-scoped :root editor color fallback overrides the dark theme ramp

### Current behavior

In dark mode, sticky notes (Stickies page, also sticky comments) render with light pastel backgrounds — e.g. the orange card computes to `rgb(255, 227, 205)` — while the text stays light (`oklch(0.92 …)`), making the content unreadable.

Expected: in dark mode the cards should use the dark ramp from `[data-theme*="dark"]` (orange `#583e2a` etc.) so the light editor text stays readable.

### Screenshots

Before (dark mode, current `preview`): light card backgrounds with unreadable light text.

dark mode before — light cards, unreadable text

After (dark mode, with the fix from the linked PR): dark ramp backgrounds, readable text.

dark mode after — dark cards, readable text

### Steps to reproduce

1. Run the current `preview` branch (observed on self-hosted v1.4.2).
2. Switch to the dark theme.
3. Open the Stickies page and create or view a sticky note.
4. Observe the light background with near-white text.

### Root cause analysis

The editor color variables used by stickies (`--editor-colors-*-background`, consumed via `STICKY_COLORS_LIST` → inline `backgroundColor: var(...)`) are defined in two places:

- `packages/editor/src/styles/variables.css` — `:root` light defaults first, then `[data-theme*="light"]` and `[data-theme*="dark"]` ramps.
- `apps/web/styles/globals.css` and `apps/space/styles/globals.css` each keep an **un-scoped `:root` fallback block** holding the same light values (intended as a pre-theme fallback).

Within the editor package the `:root` defaults come before the themed blocks, so the dark ramp wins there. But in the bundled app CSS, the apps' fallback `:root` blocks are ordered **after** the editor package's `[data-theme*="dark"]` rules — and at equal specificity (`:root` vs attribute selector both matching `html`) the later rule wins, so the light fallback overrides the dark ramp regardless of theme. Light mode was unaffected only because the fallback values happen to be identical to the light ramp.

### Proposed fix

Scope the fallback blocks so they can never override a themed ramp:

```css
:root:not([data-theme*="light"]):not([data-theme*="dark"]) { … }
```

This keeps the fallback semantics for the pre-theme window and for `custom` themes (next-themes includes a `custom` theme that matches neither selector, so deleting the block outright would leave those variables undefined).

Verified on a self-hosted instance rebuilt with the fix: dark mode computes sticky backgrounds to the dark ramp (`rgb(88, 62, 42)` for orange) with light readable text; light mode is unchanged (`rgb(255, 227, 205)` + dark text), since the fallback values are value-identical to the `[data-theme*="light"]` ramp.

I have the fix ready and will open a PR referencing this issue.

### Environment

Production / self-hosted (Docker) and local dev on `preview` branch

### Browser

Google Chrome

### Version

v1.4.2 / current `preview` branch

Liewzheng:
### Description

Fixes stickies (and any other consumer of `--editor-colors-*-background`) rendering with light pastel backgrounds in dark mode, making their light text unreadable.

Both `apps/web/styles/globals.css` and `apps/space/styles/globals.css` keep an un-scoped `:root` fallback block of editor color variables (light values) for the pre-theme window. In the bundled CSS these blocks are ordered after `@plane/editor`'s `[data-theme*="dark"]` ramp; at equal specificity the later rule wins, so the light fallback overrode the dark ramp in every theme.

The fix scopes the fallback with `:not([data-theme*="light"]):not([data-theme*="dark"])`, so it only applies before a theme lands and for `custom` themes (next-themes ships a `custom` theme matching neither selector — deleting the block outright would leave those variables undefined there).

### Type of Change

### Test Scenarios

- Rebuilt the web and space images from this branch on a self-hosted instance and verified in dark mode: sticky card backgrounds now compute to the dark ramp (`rgb(88, 62, 42)` for orange, `rgb(31, 73, 92)` for light-blue) with light, readable text.
- Light mode unchanged: computed values identical to before (`rgb(255, 227, 205)` for orange + dark text) — the fallback block's values are value-identical to the editor package's `[data-theme*="light"]` ramp, and the `:not()` scoping never matches once a theme is applied.
- `npx oxfmt --check` passes on both modified files.

### References

Closes #9753

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed themed editor colors being overridden by unthemed fallback styles.
  * Light and dark editor themes now display the intended background color ramps consistently.

### DIFF
--- a/apps/space/styles/globals.css
+++ b/apps/space/styles/globals.css
@@ -8,8 +8,10 @@
  * @plane/editor/styles (imported above), which is their single source of truth.
  * Only the un-themed :root fallback below is kept: the editor package declares
  * background colors solely under [data-theme*="light"] / [data-theme*="dark"],
- * so without this they are undefined until a theme lands on the element. */
-:root {
+ * so without this they are undefined until a theme lands on the element.
+ * Scoped with :not() so it never overrides the themed ramps — as a plain :root
+ * rule later in the bundle it would win over them regardless of theme. */
+:root:not([data-theme*="light"]):not([data-theme*="dark"]) {
   --editor-colors-gray-background: #d6d6d8;
   --editor-colors-peach-background: #ffd5d7;
   --editor-colors-pink-background: #fdd4e3;
--- a/apps/web/styles/globals.css
+++ b/apps/web/styles/globals.css
@@ -10,8 +10,10 @@
  * @plane/editor/styles (imported above), which is their single source of truth.
  * Only the un-themed :root fallback below is kept: the editor package declares
  * background colors solely under [data-theme*="light"] / [data-theme*="dark"],
- * so without this they are undefined until a theme lands on the element. */
-:root {
+ * so without this they are undefined until a theme lands on the element.
+ * Scoped with :not() so it never overrides the themed ramps — as a plain :root
+ * rule later in the bundle it would win over them regardless of theme. */
+:root:not([data-theme*="light"]):not([data-theme*="dark"]) {
   --editor-colors-gray-background: #d6d6d8;
   --editor-colors-peach-background: #ffd5d7;
   --editor-colors-pink-background: #fdd4e3;


Liewzheng:
Hi @anmolsinghbhatia — friendly ping on this PR. CLA is signed and checks are green. Could you or the team take a look when you have a moment? Happy to adjust anything. Thanks!

Issue #9788, opened 2026-09-08
[feature]: LaTeX math in the rich text editor ($$ display blocks + $ inline, KaTeX-rendered)


Liewzheng:
- [x] I have searched the existing issues

### Summary

Add LaTeX math rendering to the rich text editor: `$$...$$` display math blocks (single- or multi-line) and `$...$` inline math, rendered with KaTeX — in work item descriptions, comments, and pages.

### Why should this be worked on?

Teams writing technical specs, research notes or anything quantitative currently paste screenshots of formulas. Native math keeps formulas as text: editable, searchable, copy-pasteable.

Implementation notes (working PR attached):

- Two new atom nodes, `mathBlock` (`$$`) and `mathInline` (`$`), with click-to-edit source UX and graceful error display (source + parser message, never crashes the editor).
- Currency-safe: `$100` / `$5.99` are never treated as math (content may not start with a digit/whitespace; `$$` always prefers the block rule).
- KaTeX (~277KB) is loaded via dynamic `import()` only when a math node is on screen; `strict: true, trust: false` — no raw HTML injection.
- Full round-trip: Markdown parse/serialize (`$$..$$` / `$..$`), yjs collaborative editing (schema-only, works like image/table nodes), read-only surfaces (peek/space) render automatically, HTML sanitization allowlist extended (`math-block`/`math-inline` + `data-latex`).
- i18n: node-view strings translated (en + zh-CN, other locales synced per the sync-check CI).

sriramveeraghanta:
Thanks for the detailed proposal, @Liewzheng.

LaTeX/math support in the editor is part of Plane's commercial edition, and we don't have plans to bring it into the community edition — so we're closing this feature request along with #9789.

We appreciate the effort you put into both the issue and the accompanying PR.

PR #9789, opened 2026-09-08
feat(editor): LaTeX math support — $$ display blocks and $ inline (KaTeX)

Liewzheng:
### Description

Adds LaTeX math to the editor (work item descriptions, comments, pages):

- **`mathBlock`** — `1753889` display math (single- or multi-line), slash command "Math", input rule (`1753889` + space/enter).
- **`mathInline`** — `$...$` inline math with currency-safe matching (`$100`, `$5.99` stay plain text; content may not start with a digit/whitespace; `1753889` prefers the block rule).

Both are atom nodes with KaTeX preview + click-to-edit source (textarea, blur/Esc back to preview); invalid LaTeX shows the source plus the parser error without crashing. KaTeX is dynamically imported (out of the main bundle, SSR-safe), `strict: true, trust: false`. KaTeX CSS is imported in each app's root (Vite) so the 61 woff2 font assets emit correctly.

Round-trip covered: markdown-it parse + serialize (`1753889..1753889` / `$..$`), `renderHTML`/`renderText`, yjs collaboration (plain schema nodes), read-only surfaces render automatically, and the server-side nh3 allowlist accepts `math-block`/`math-inline` + `data-latex` (with unit tests proving `onclick`/`<script>` still get stripped).

### Type of Change

### Test Scenarios

- [x] vitest 22/22 (block + inline), including currency boundaries (`00`, `.99`, unclosed `$`, escaped `\$`) and mixed documents.
- [x] markdown-it end-to-end: inline math, prices, mixed price+formula sentences, block precedence.
- [x] `check:types` 13/13 (utils/editor/web), i18n sync-check 100% across locales, lint clean on new files, web build emits KaTeX fonts correctly.
- [x] Backend nh3 smoke: math tags + `data-latex` preserved, event handlers/scripts stripped. New pytest cases included (`test_content_validator_math_block.py`).

### References

- Closes #9788


sriramveeraghanta:
Thanks for the thorough work here, @Liewzheng — the implementation, tests, i18n coverage and the nh3 allowlist handling are all clearly well thought through.

That said, we have to close this one. LaTeX/math support in the editor is part of Plane's commercial edition, and we don't have plans to bring it into the community edition. Keeping a parallel implementation in CE would put us in conflict with the paid feature, so we can't merge it here.

This isn't a reflection on the quality of the PR — it's purely an edition/roadmap decision. We really appreciate the effort you put in, and we'd be glad to see contributions from you on other areas of the codebase.

Closing this along with #9788.

Issue #5827, opened 2024-10-14
[bug]: Url for images/attachments wrong http instead of https


tgivslife:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When i try to add an image or attachment to an issue, the url returned by the following request

<img width="878" alt="Screenshot 2024-10-14 at 15 42 00" src="https://github.com/user-attachments/assets/536cdc91-717e-42f0-8bc8-fd91b7549b66">

returns 
```
{
    "upload_data": {
        "url": "http://dev.domain/uploads",
        "fields": {
            "Content-Type": "image/png",
            "key": "0600252d-b281-4198-994a-1cd0f1fcdd04/bc455979f1aa45f6b454a8a3608de5d8-ab_solid_shadow_holo.9 (1).png",
            "x-amz-algorithm": "AWS4-HMAC-SHA256",
            "x-amz-credential": "access-key/20241014/eu-central-1/s3/aws4_request",
            "x-amz-date": "20241014T123859Z",
[...]

PR #9842, opened 2026-09-16
fix(storage): honour MINIO_ENDPOINT_SSL when building export presigned URLs

### LINKED ISSUE #9550: Worklogs export download links use http:// even with MINIO_ENDPOINT_SSL=1 set
Related to #5827, same root issue but specific to the Worklogs/Timesheets export feature.

We're self-hosting Plane Commercial (Docker Compose via prime-cli), behind Cloudflare Tunnel so TLS ends at Cloudflare and the internal proxy traffic is plain HTTP. Tested this on both v2.5.3 and v3.0.1, same result on both.

Every download link from Worklogs export (Project Settings > Worklogs > Download, plus the Previous Downloads list) comes back as an http:// presigned URL, like:

http://<domain>/uploads/<uuid>.csv?X-Amz-Algorithm=AWS4-HMAC-SHA256&...

Since the site itself is https, Chrome silently blocks these as insecure downloads. No error, no toast, the click just does nothing. Confirmed the file itself is fine, curl -L on the same url works and pulls the CSV down no problem, so it's really just the scheme on the link that's wrong.

We set MINIO_ENDPOINT_SSL=1 and AWS_S3_ENDPOINT_URL=https://<our domain>, which per storage.py is supposed to force https on generated presigned urls when behind a proxy like ours. Verified both vars are actually set inside the api and worker containers, and the app confirms it can reach the bucket fine through the new url on startup (Checking bucket... Bucket 'uploads' exists.).

Everything else that goes through S3Storage seems to respect this fine based on the code, but Worklogs export specifically doesn't. Freshly generated exports (checked via server request logs, not cached) still come back http:// even with the fix applied and containers freshly restarted. Seems like Worklogs/Timesheets export must be running separate code from the normal S3Storage.generate_presigned_url path, maybe since it's an enterprise only feature.

Anyone know if Worklogs export url generation is a separate code path, and if so can it get the same SSL override support?

Happy to share more logs if useful.

Hayyan612:
### Description

Presigned download URLs generated by the export task come back on `http://` even when `MINIO_ENDPOINT_SSL=1` is set, so on an https instance the browser blocks the download silently.

There are two presigned-URL paths and only one of them honoured the flag.

`S3Storage` checks it directly, `apps/api/plane/settings/storage.py:41`:

```python
if os.environ.get("MINIO_ENDPOINT_SSL") == "1":
    endpoint_protocol = "https"
else:
    endpoint_protocol = request.scheme if request else "http"
```

The export task never touches `S3Storage`. It builds its own boto3 client, `apps/api/plane/bgtasks/export_task.py:65-73`:

```python
presign_s3 = boto3.client(
    "s3",
    endpoint_url=(
        f"{settings.AWS_S3_URL_PROTOCOL}//{str(settings.AWS_S3_CUSTOM_DOMAIN).replace('/uploads', '')}/"
    ),
    ...
)
```

and `AWS_S3_URL_PROTOCOL` was taken from the scheme of `WEB_URL`, not from `MINIO_ENDPOINT_SSL` and not from `AWS_S3_ENDPOINT_URL`. That is the only definition and the only use of the setting in the repo.

This is the default path for self-hosters rather than an edge case: `deployments/cli/community/variables.env:15` hardcodes `WEB_URL=http://${APP_DOMAIN}`, so anyone terminating TLS at a proxy gets `http:` here regardless of what they set.

This PR makes the setting mirror the check `storage.py` already makes, so the two paths agree.

### Type of Change

- [ ] Code refactoring
- [ ] Documentation update

### Test Scenarios

Exercised the settings logic across the realistic configurations to confirm nothing but the intended case moves:

| config | before | after |
|---|---|---|
| TLS at proxy, internal http, `MINIO_ENDPOINT_SSL=1` | `http://plane.example.com/` | `https://plane.example.com/` |
| compose default, flag unset | `http://plane.example.com/` | unchanged |
| `WEB_URL` already https | `https://plane.example.com/` | unchanged |
| plain http deploy, `MINIO_ENDPOINT_SSL=0` | `http://192.168.1.50/` | unchanged |

Only a deployment that explicitly opted into SSL changes behaviour. `ruff check` passes and the changed line is `ruff format` clean.

Two things reviewers should know:

- `ExporterHistory.url` is a stored `URLField` written once at `export_task.py:118`, so rows created before this change keep their old `http://` links. Only new exports come back corrected.
- This looks like the same root cause as #5827, but I have not been able to test the images and attachments surface, so I have not claimed it there.

There is already a `# TODO: Change the upload_to_s3 function to use the new storage method` at `export_task.py:41`. Routing export through `S3Storage` would remove this whole class of divergence rather than patching one symptom, but it is a much larger change, so I kept this to the single line that makes the two paths agree. Happy to take it that way instead if you would prefer.

### References

Fixes #9550
Related #5827

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected storage URL handling so secure MinIO connections consistently use HTTPS.
  - Preserved the configured website URL scheme for other storage connections.
  - Improved custom storage domain handling based on the configured website URL and bucket name.


Hayyan612:
Good catch, added in dc63f51.

`plane/tests/unit/settings/test_export_url_protocol.py` reloads `plane.settings.common` under a patched environment, since as you noted the value is computed at import time and the existing `S3Storage` tests never reach it. Five cases:

- `MINIO_ENDPOINT_SSL=1` with an http `WEB_URL` gives `https:` (the TLS-at-proxy case this PR fixes)
- `MINIO_ENDPOINT_SSL=0` falls back to the `WEB_URL` scheme
- unset flag falls back to the `WEB_URL` scheme
- an https `WEB_URL` with no flag is unchanged
- `AWS_S3_CUSTOM_DOMAIN` is still built from the `WEB_URL` host plus the bucket, so the override does not disturb the host half of the endpoint

I checked the suite actually catches the regression rather than just passing alongside the fix. Reverting the one-line change locally:

```
FAILED test_export_url_protocol.py::TestExportURLProtocol::test_ssl_flag_forces_https_when_web_url_is_http
1 failed, 4 passed
```

The three fallback cases pass in both states, which is the point of them, and only the new behaviour fails without the fix. With the fix, 5 passed.

An autouse fixture reloads `plane.settings.common` again on teardown so the module is left as the rest of the suite expects. `ruff check` and `ruff format --check` both pass on the new file.

PR #9851, opened 2026-09-17
fix(i18n): add missing Polish plural form for epic.label

Pawel-Kranzberg:
## Summary
- Polish (`pl`) requires three plural forms — `one`, `few`, and `other` — but `work-item.json`'s `epic.label` only declared `one`/`other`: `"{count, plural, one {Epic} other {Epics}}"` → `"{count, plural, one {Epik} other {Epiki}}"`.
- Because Polish's `few` category (2–4, excluding 12–14) wasn't declared, those counts fell through to `other`, which happened to read correctly by coincidence ("2 Epiki" is right). But counts of 5+ also fell through to that same `other` bucket, which should instead use the genitive plural — so "5 Epiki" rendered where "5 Epików" is correct.
- Adds the missing `few` form and fixes `other` to the genitive: `"{count, plural, one {Epik} few {Epiki} other {Epików}}"`.

This mirrors the same fix already made in our fork's translation-coverage PR ([OSM-Piatnica/pi-plane#19](https://github.com/OSM-Piatnica/pi-plane/pull/19)), submitted upstream since the underlying string is unchanged from the base Plane project. Every other Polish plural string in the repo already uses this one/few/other pattern — this was the one exception.

## Test plan
- [x] Valid JSON (parses cleanly)
- [x] `npx tsx packages/i18n/scripts/sync-check.ts` — `pl` still at 100% key coverage (3,842/3,842), no missing/stale keys, no collisions
- [x] `npx oxfmt --check` on the changed file reports no issues
- [ ] Manual UI verification of the count-based rendering (not done as part of this change; the ICU MessageFormat plural categories `one`/`few`/`other` are standard Polish CLDR rules)

## Summary by CodeRabbit

- **Localization**
  - Updated the Polish translation for the plural form of “epic” to provide the appropriate wording for small quantities.


vihar (approved):
Checked this against the other Polish plurals in the repo and the CLDR rules: `few` for 2–4 with the genitive in `other` for 5+ is the same shape every other `pl` plural uses, and sync-check and oxfmt are clean with the patch applied. Merging, thanks.

One note: this wasn't quite the only exception. The comment count, allowed state transitions, and reviewers strings in `pl` have the same one/other gap. If you feel like a follow-up, happy to take it.

PR #9862, opened 2026-09-19
fix(web): stop 401 redirect reload loop on session expiry

### LINKED ISSUE #9861: [bug]: Expired session causes endless reload loop / error boundary instead of the sign-in screen
### Current Behavior

When the session cookie expires while the web app is open, the user is not returned to the sign-in screen. Depending on timing they either land on an error boundary ("Looks like something went wrong") or the page keeps reloading.

### Steps to Reproduce

1. Sign in, keep the tab open until the session expires (or invalidate the server-side session).
2. Interact with the app (e.g. open a work items list): the request returns 401.
3. The axios response interceptor in `apps/web/core/services/api.service.ts` runs `window.location.replace("/?next_path=<currentPath>")`.
4. The entry page `"/"` mounts and fires its own current-user request; that request 401s too.
5. The interceptor runs again — now `currentPath` is `"/"`, so it replaces `"/"` with `"/?next_path=/"`, reloading the entry page in an endless loop. The `NON_AUTHENTICATED` branch of `AuthenticationWrapper` never gets a stable chance to render the sign-in form.

### Root Cause

The 401 interceptor redirects unconditionally, including for requests made by the entry page itself, producing a self-referential replace loop on `"/"`.

### Suggested Fix

Skip the redirect when `window.location.pathname` is already `"/"`. A PR follows with the one-line guard.

### Environment

- Self-hosted v1.4.2 built from source.

yuzhiyang1:
Fixes #9861

## What

When a session expires, the 401 response interceptor in `apps/web/core/services/api.service.ts` redirects to `/?next_path=<currentPath>` unconditionally. The entry page `/` performs its own current-user request on mount; on an expired session that request 401s as well, and the interceptor replaces `/` with `/?next_path=/` — reloading the entry page in an endless loop. The sign-in screen never renders and users see an error boundary or a constantly reloading page.

## How

Skip the redirect when `window.location.pathname` is already `/`, so the `NON_AUTHENTICATED` branch of `AuthenticationWrapper` can render the sign-in form for requests originating from the entry page itself. Private routes keep the existing `next_path` behaviour (bounce to `/`, sign in, return where you were).

## Verification

Self-hosted v1.4.2 built from source: with an expired session cookie, loading `/` after this change renders the sign-in form instead of looping; authenticated flows are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unnecessary redirect loops when authentication fails on the root entry page.
  * Unauthorized requests on the home page now remain on that page instead of triggering a self-reloading redirect.
  * Preserved redirects for unauthorized requests on other pages while safely encoding the return path.


prateekshourya29 (approved):

prateekshourya29:
@yuzhiyang1 Thanks for your contribution! Everything looks good.

